PR #2, opened 2022-05-10
CONTENT-33: initial api work

jlsherrill:


rverdile:
I've noticed you tend to be explicit with the int size (int64). Do you think this is something we should be doing everywhere?

jlsherrill:
I'm not 100% sure why i originally went that way, but i do want to describe some quirks around this.  Limit() and Offset() in gorm seem to take in ints, while its Count() seems to return an int64.    It looks like x86_64 systems use 64bits for 'int' so in practice it really doesn't matter, however its possible that converting from int64 to int could lead to a smaller number.  

That said, i think it makes sense to use int and only convert from int64 in the one case we need (the Count() method).  I'm gonna update to do that.

pkg/handler/repositories.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"github.com/content-services/content-sources-backend/pkg/db"
+	"github.com/content-services/content-sources-backend/pkg/models"
+	"github.com/labstack/echo/v4"
+)
+
+type RepositoryItem struct {
+	UUID                string `json:"uuid"`
+	Name                string `json:"name"`
```
rverdile, line 11: What's the difference between a RepositoryItem and a RepositoryConfiguration? And is this struct meant to only work with RepositoryConfigurations, or will there be others?

jlsherrill: generally from what i've seen, you separate your api return data from your database models.  So the RepositoryConfiguration is the database representation, the RepositoryItem is the api representation.  And you have methods that can convert from one to the other.  This lets you make changes to the data model without affecting the UI, as well as combining multiple tables worth of data into a single 'item' in the repsonse (which will be used when we include information about # of packages, etc..)

rverdile:
How do I install the `swag` command? I can't run `make openapi` without it. Might also be good to add this to the README.

Makefile
```diff
@@ -1,9 +1,4 @@
-TEST_DB_HOST:=
-TEST_DB_PORT:=
-TEST_DB_USER:=
-TEST_DB_PASSWORD:=
-TEST_DB_NAME:=
-
+.PHONY: test
```
rverdile, line 1: Would it make sense to put every target in .PHONY? I'm finding Makefiles with Go a little confusing because Go already has it's own command line tools.

rverdile (commented):
Another small comment. Everything else so far looks good to me. Is there more pushes on the way?

Makefile
jlsherrill: i think its only needed when a directory exists with the same name as a target, but i'm 100% comfortable with it either

db/migrations/20220512132042_create_repository_configurations_table.up.sql
```diff
@@ -9,8 +9,7 @@ CREATE TABLE IF NOT EXISTS repository_configurations(
     account_id VARCHAR (255) NOT NULL,
     org_id VARCHAR (255) NOT NULL,
     created_at timestamp NOT NULL,
-    updated_at timestamp NOT NULL,
-    deleted_at timestamp
```
jlsherrill, line 13: i removed deleted_at as we discussed doing real deletes, and this was getting in the way of my testing

jlsherrill:
I think thats my last major push.

pkg/handler/repository_test.go
```diff
@@ -0,0 +1,149 @@
...
+		assert.Equal(t, int64(1), response.Meta.Count)
+		assert.Equal(t, 100, response.Meta.Limit)
+		assert.Equal(t, 1, len(response.Data))
+		assert.Equal(t, repoConfig.Name, response.Data[0].Name)
+		assert.Equal(t, repoConfig.URL, response.Data[0].Url)
+		assert.Equal(t, repoConfig.AccountID, response.Data[0].AccountId)
+	}
+}
+
+func (suite *ReposSuite) TestListOdd() {
```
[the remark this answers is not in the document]
rverdile: I think I see the difference now. Odd tests when the number of repositories isn't divisible by the limit, so the last page has of results has the stragglers. 

Is that right? If so, what if these were renamed or there's a comment clarifying? At first I thought it had to do with even/odd numbers.

Makefile
rverdile: I think we can leave it as is since it's working

pkg/handler/repository_test.go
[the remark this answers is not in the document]
jlsherrill: yeah, makes sense, i'll reword

rverdile (approved):
pending the small changes, and if there's no big changes, this looks good to me.

pkg/handler/repository_test.go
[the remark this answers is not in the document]
rverdile: it's clearer now!

PR #3, opened 2022-05-10
updating Readme and adding arch doc

jlsherrill:


README.md
```diff
@@ -1,2 +1,13 @@
-# content-sources-backend
-Backend for content sources service
+# Content Sources
+
+##What is it?
+Content Sources is an application for storing content references (currently YUM repositories) in a central location.
```
rverdile, line 4: It might be good to define what a "content reference" is, maybe with an example.

docs/architecture.md
```diff
@@ -0,0 +1,17 @@
...
+
+Note: this document is targeting the desired architecture for the first milestone (November 2022).  It will be updated as needed. 
+
+This application consists of these parts:
+* Rest Api
+  * Used for our frontend or for other applications to interact with ours
+* Postgresql Database
+  * Used for storing our data
+* Kafka Platform
+  * Used to push messages for the event listener and other applications that want information around content sources
```
rverdile, line 13: If there's an example you about what "information" would get pushed, that might be useful. But maybe it's too early for that :)

rverdile (commented):
Just some suggestions. Feel free to disagree. Obviously I really like examples :).  I could see a small "example" section or doc, but maybe we don't have that information yet?

jlsherrill:
good ideas @rverdile !    i reworded the first part and added examples for the rest.  Let me know if its clear or not

rverdile (approved):
Much clearer now! This looks good to me.

PR #4, opened 2022-05-12
CONTENT-39: repository configuration database and model

rverdile:
If you want to try to use this, you have to create a psql database and set the environment variables for connecting (see Makefile for names).

I made the DB connection a global variable in main just as an example. I don't know how we actually want to handle that, but I'm assuming we don't want to open new connection every time we access the database.

To-do List:

- [x] Add force rollback for failed migrations
- [x] Add template for migrations, so creating a new automatically includes transaction syntax
- [x] Remove repository auth configuration ID
- [x] Remove "name" flag in favor of dbmigrate new NAME
- [x] Move TestMain to its own file
- [ ] [Skip] Integrate with clowder for database configuration
- [x] Handle local env variables for database configuration


db/migrations/20220512132042_create_repository_configurations_table.up.sql
```diff
@@ -0,0 +1,16 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS repository_configurations(
+    uuid VARCHAR (255) UNIQUE NOT NULL PRIMARY KEY,
+    name VARCHAR (255) NOT NULL,
+    url VARCHAR (255) NOT NULL,
+    version VARCHAR (255),
+    arch VARCHAR (255),
+    account_id VARCHAR (255),
+    repository_auth_configuration_id VARCHAR (255),
```
jlsherrill, line 10: i'd leave off this repository_auth_configuration_id now.

jlsherrill: until we add that model that is :)

pkg/config/config.go
```diff
@@ -0,0 +1,41 @@
...
+	if clowder.IsClowderEnabled() {
+		//TODO: add handling for clowder
+		return nil
+	} else {
+
+		options.SetDefault("DBHost", os.Getenv("DB_HOST"))
+		options.SetDefault("DBPort", os.Getenv("DB_PORT"))
+		options.SetDefault("DBUser", os.Getenv("DB_USER"))
+		options.SetDefault("DBPassword", os.Getenv("DB_PASSWORD"))
+		options.SetDefault("DBName", os.Getenv("DB_NAME"))
```
jlsherrill, line 30: these env variables don't seem to match whats in the Makefile?

jlsherrill: oh ignore me, you've updated it :)

cmd/content-sources/main.go
```diff
@@ -3,18 +3,34 @@ package main
 import (
 	"log"
 
+	"github.com/content-services/content-sources-backend/pkg/db"
 	"github.com/gin-gonic/gin"
+	"gorm.io/gorm"
 )
 
+var DB *gorm.DB
```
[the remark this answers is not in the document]
rverdile: You're right, there should only be one. I didn't mean to put one in db. Where that global lives depends on how we want to call it: `DB` or `db.DB`. I don't actually use this DB global in my PR; it's just here as an example.

pkg/db/db.go
```diff
@@ -0,0 +1,153 @@
...
+		conf.DBName,
+		conf.DBHost,
+		conf.DBPort,
+	)
+}
+
+// Connect returns pointer to a new gorm DB connection
+func Connect() (*gorm.DB, error) {
+	dbURL := GetUrl()
+	DB, err := gorm.Open(pg.Open(dbURL), &gorm.Config{})
```
jlsherrill, line 39: this actaully creates a local DB variable due to the :=  and doesn't set the global one (it stays nil)

cmd/content-sources/main.go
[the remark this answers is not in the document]
jlsherrill: i think we actually want the one in db, because the api handlers need a way to access the db connection.  We don't want the api layer importing 'main' (as the main layer is importing the api package which would create a circular dependency)

pkg/db/db.go
rverdile: right, the global DB is there by mistake

cmd/content-sources/main.go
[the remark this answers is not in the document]
rverdile: That makes sense. I'll move it to db.

rverdile:
There's gonna be some misc print statements or inaccurate error messages while I debug the CI, so don't comment on those :)

cmd/content-sources/main.go
```diff
@@ -3,18 +3,31 @@ package main
...
+
+	err = db.Connect()
+	if err != nil {
+		log.Fatalf("%v", err)
+	}
+
+	err = db.MigrateDB(db.GetUrl(), "up")
+	if err != nil {
+		log.Fatalf("%v", err)
+	}
```
jlsherrill, line 21: I'd get rid of the migration on startup.  I think we'll use an 'init container' to accomplish this.   This also causes the server to die if its fully migrated

pkg/models/repository_configuration.go
```diff
@@ -0,0 +1,10 @@
+package models
+
+type RepositoryConfiguration struct {
+	Base
+	Name      string `json:"name" gorm:"default:null"`
+	URL       string `json:"url" gorm:"default:null"`
+	Version   string `json:"version" gorm:"default:null"`
+	Arch      string `json:"arch" gorm:"default:null"`
+	AccountID string `json:"account_id" gorm:"default:null"`
```
jlsherrill, line 9: Can you add an OrgId here too, as a string?

rverdile: can Orgid be null?

jlsherrill: nope, both should be non-null

jlsherrill (approved):
ACK

dance

PR #9, opened 2022-05-18
CONTENT-50

Andrewgdewar:


pkg/handler/api.go
```diff
@@ -66,7 +67,7 @@ func RegisterRoutes(engine *echo.Echo) {
 
 	data, err := json.MarshalIndent(engine.Routes(), "", "  ")
 	if err == nil {
-		log.Debug(string(data))
+		log.Debug().Msg(string(data))
```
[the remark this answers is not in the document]
jlsherrill: In this case there isn't an error (notice err==nil).  Here's i'm just debugging all our routes on startup in debug mode.  This isn't really necessary, but i found it handy when debugging stuff early on.

jlsherrill:
can we somehow integrate in with https://github.com/rs/zerolog#pretty-logging  maybe via an Env variable?  

Running as is is really hard to read

pkg/handler/api.go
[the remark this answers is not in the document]
Andrewgdewar: KK

jlsherrill:
i guess its fine for now.  We can re-evaulate as we go along.  

I was getting a log for each request on your previous push, but now i'm not getting anything when curling /ping, any idea?

jlsherrill:
You might also checkout the provisioning team's zerolog integration: https://github.com/RHEnVision/provisioning-backend/blob/c6f31b463defcd55861434ab42b74fdd8d6ddd91/internal/logging/zerolog.go

jlsherrill:
and we should probably integrate with https://github.com/spf13/viper at some point, in a simialr way.  I'll add a card to do that to the board

Andrewgdewar:
> 

I'm investigating now.

cmd/content-sources/main.go
```diff
@@ -1,25 +1,53 @@
...
 )
 
 func main() {
+	e := echo.New()
+
+	logLevel := golog.ERROR
+
+	if false { //TODO: Dev debug based on env
+		logLevel = golog.DEBUG
+	}
```
jlsherrill, line 23: Lets go ahead and turn on console logging in this pr, and then make it configurable via viper integration.  I think just this:

log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})

and then passing  log.Logger instead of os.Stdout below in the lecho.New() call does it

Andrewgdewar: Nice.

Andrewgdewar: Addressed.

jlsherrill (approved):
ACK

PR #12, opened 2022-05-24
trim hash for image pushes

jlsherrill:


jlsherrill:
tested this first by having it run on a pr and it pushed 9a0ee23:  https://quay.io/repository/jlsherri/content-sources?tab=tags   so i think this is good

rverdile (approved):
ack

PR #13, opened 2022-05-24
Update arch doc to use puml

jlsherrill:


Andrewgdewar (approved):
Super neat, Tested output on markdown website. Looks good.
Fix conflicts then good to go.

PR #14, opened 2022-05-25
Adding postgresql instructions

jlsherrill:

PR #16, opened 2022-05-25
simpler loading of .env file in Makefile

jlsherrill:
This makes loading .env more generic and a bit simpler.  Curious your thoughts.


jlsherrill:
It also adds a sample .env file.  I'll update the README after https://github.com/content-services/content-sources-backend/pull/14 goes in

rverdile:
@jsherril i'm not familiar with the sed command. what's that doing?

jlsherrill:
so say you have an env file with: 
```
FOO=BAR
BAZ=FOZ
```

the make target 'make .env-export' takes your .env and creates a .env-export file:
```
export FOO=BAR
export BAZ=FOZ
```

Then in any command we need to use the configuration in, we can just use  '. .env-export && COMMAND' 
and it will basically export all the variables into the process.  I'm a bit surprised how obtuse this all is, but this seemed like the easiest way.

jlsherrill:
I think when we switch to viper, this may all be a bit easier though.

Makefile
```diff
@@ -11,8 +12,12 @@ dbmigrate:
 seed:
 	go run cmd/dbmigrate/main.go seed
 
-test:	
-	DATABASE_PASSWORD=$(DATABASE_PASSWORD) DATABASE_NAME=$(DATABASE_NAME) DATABASE_HOST=$(DATABASE_HOST) DATABASE_USER=$(DATABASE_USER) DATABASE_PORT=$(DATABASE_PORT) go test ./...
+
+.env-export: .env
```
rverdile, line 16: what does the .env dependency here do? it seems to run okay without it.

jlsherrill:
actually, i think i'm just gonna integrate with viper

PR #19, opened 2022-05-26
CONTENT-40: Create content source

rverdile:
Adds create endpoint for content source. Adds a 'dao' layer and 'api' package.

The 'api' package defines all the API request/response structs. For example, with create, when the handler gets a request, it binds it to the 'CreateRepository' struct in the 'api' package. Then, that struct is passed to the 'dao' package which interacts with the models and the database to the necessary items to the database.

To test you need to use the "x-hr-identity" header. You want a base64 encoded JSON that looks like: 

> "identity": {
        "account_number": "12345",
        "org_id": "54321"
}

Handling the account_number and org_id relies on this common package: https://github.com/RedHatInsights/platform-go-middlewares/tree/master/identity

TODO:

- [x]  Create endpoint with OpenAPI comments
- [x]  Unique constraint between org ID and URL
- [x]  dao layer
- [x]  tests
- [x]  address swag param type.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ test-ci:
 	go test ./...
 
 openapi:
-	swag init --generalInfo api.go  --dir pkg/handler/
+	swag init --generalInfo api.go --dir pkg/handler/ --pd pkg/api
 	#convert from swagger to openapi
 	go run ./cmd/swagger2openapi/main.go docs/swagger.json docs/openapi.json
 	rm docs/swagger.json docs/swagger.yaml
--- a/pkg/api/repository.go
+++ b/pkg/api/repository.go
@@ -1,0 +1,46 @@
+package api
+
+import "github.com/content-services/content-sources-backend/pkg/models"
+
+// Repository holds data received from, or returned by, a repositories API request or response
+type Repository struct {
+	UUID                string `json:"uuid" readonly:"true"`
+	Name                string `json:"name"`
+	URL                 string `json:"url"`                                //URL of the remote yum repository
+	DistributionVersion string `json:"distribution_version" example:"7"`   //Version to restrict client usage to
+	DistributionArch    string `json:"distribution_arch" example:"x86_64"` //Architecture to restrict client usage to
+	AccountID           string `json:"account_id" readonly:"true"`         //Account ID of the owner
+	OrgID               string `json:"org_id" readonly:"true"`             //Organization ID of the owner
+}
+
+// CreateRepository holds data received from request to create repository
+type CreateRepository struct {
+	UUID                string `json:"uuid" readonly:"true" swaggerignore:"true"`
+	Name                string `json:"name"`
+	URL                 string `json:"url"`                                             //URL of the remote yum repository
+	DistributionVersion string `json:"distribution_version" example:"7"`                //Version to restrict client usage to
+	DistributionArch    string `json:"distribution_arch" example:"x86_64"`              //Architecture to restrict client usage to
+	AccountID           string `json:"account_id" readonly:"true" swaggerignore:"true"` //Account ID of the owner
+	OrgID               string `json:"org_id" readonly:"true" swaggerignore:"true"`     //Organization ID of the owner
+}
+
+func (r *Repository) FromRepositoryConfiguration(repoConfig models.RepositoryConfiguration) {
+	r.UUID = repoConfig.UUID
+	r.Name = repoConfig.Name
+	r.URL = repoConfig.URL
+	r.DistributionVersion = repoConfig.Version
+	r.DistributionArch = repoConfig.Arch
+	r.AccountID = repoConfig.AccountID
+	r.OrgID = repoConfig.OrgID
+}
+
+type RepositoryCollectionResponse struct {
+	Data  []Repository     `json:"data"`  //Requested Data
+	Meta  ResponseMetadata `json:"meta"`  //Metadata about the request
+	Links Links            `json:"links"` //Links to other pages of results
+}
+
+func (r *RepositoryCollectionResponse) SetMetadata(meta ResponseMetadata, links Links) {
+	r.Meta = meta
+	r.Links = links
+}
--- a/pkg/dao/repository.go
+++ b/pkg/dao/repository.go
@@ -1,0 +1,43 @@
+package dao
+
+import (
+	"fmt"
+
+	"github.com/content-services/content-sources-backend/pkg/api"
+	"github.com/content-services/content-sources-backend/pkg/db"
+	"github.com/content-services/content-sources-backend/pkg/models"
+)
+
+type repositoryDaoImpl struct {
+}
+
+func GetRepositoryDao() repositoryDaoImpl {
+	return repositoryDaoImpl{}
+}
+
+const uniqueConstraintErrMsg = "ERROR: duplicate key value violates unique constraint \"url_and_org_id_unique\" (SQLSTATE 23505)"
+
+func (r repositoryDaoImpl) Create(newRepo api.CreateRepository) error {
+
+	newRepoConfig := models.RepositoryConfiguration{
+		Name:      newRepo.Name,
+		URL:       newRepo.URL,
+		Version:   newRepo.DistributionVersion,
+		Arch:      newRepo.DistributionArch,
+		AccountID: newRepo.AccountID,
+		OrgID:     newRepo.OrgID,
+	}
+
+	dest := models.RepositoryConfiguration{}
+	result := db.DB.Where("org_id = ? AND url = ?", newRepo.OrgID, newRepo.URL).Find(&dest)
+	if result.Error != nil {
+		return result.Error
+	}
+	if err := db.DB.Create(&newRepoConfig).Error; err != nil {
+		if err.Error() == uniqueConstraintErrMsg {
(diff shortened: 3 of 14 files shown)


pkg/dao/repository.go
```diff
@@ -0,0 +1,30 @@
...
+	return repositoryDaoImpl{}
+}
+
+func (r repositoryDaoImpl) Create(c echo.Context) error {
+
+	newRepository := models.RepositoryConfiguration{}
+
+	if err := c.Bind(&newRepository); err != nil {
+		return fmt.Errorf("bind error: %w", err)
+	}
```
jlsherrill, line 25: I'm glad you started on a dao layer!  I dont' think the dao layer should bind parameters at all.  (and the layer that does should bind to the an api structure, not a db model IMO).   I sorta see the layers like this:

Handler/Api Layer -> Transforms user input into Api Layer objects (RepositoryInfo for example).   Then transforms those into db model objects (RepositoryConfiguration).  Passes these to the DAO layer.
DAO Layer -> Takes the db layer objects and does creates/updates, returns db layer objects for queries

The dependency graph would look kinda like this:

image


I had originally thought that the DAO layer could convert between the Api structures and the db models, but then we'd end up with a dep graph like this:

image

which would be a dependency cycle.

jlsherrill: So i think the flow would go:   Params -> Handler binds to RepositoryInfo  -> passes to DAO layer, which converts to RepositoryConfiguration (and down the road AuthenticationConfiguration)  -> Saves it to the db

jlsherrill: and conversely within the list api, the flow would be:

Handler  calls into dao layer, dao layer loads RepositoryConfiguration models -> dao converts to RepositoryInfo structs, returns to handler -> handler renders to json. 

I'm kinda thinking out loud here, so please let me know if this isn't making sense or you have a differing opinion :)

rverdile:
@jlsherrill I think I did what you're describing, can you take a look?

pkg/api/repository.go
```diff
@@ -0,0 +1,11 @@
+package api
+
+type RepositoryRequest struct {
+	UUID                string `json:"uuid"`
+	Name                string `json:"name"`
+	URL                 string `json:"url"`                                //URL of the remote yum repository
+	DistributionVersion string `json:"distribution_version" example:"7"`   //Version to restrict client usage to
+	DistributionArch    string `json:"distribution_arch" example:"x86_64"` //Architecture to restrict client usage to
+	AccountID           string `json:"account_id"`                         //Account Id of the owner
+	OrgID               string `json:"org_id"`                             //Organization Id of the owner
```
jlsherrill, line 10: can we not use RepositoryItem for this?

jlsherrill: and maybe renaming it makes more sense.  The handler package was sorta meant to be an 'api' package, we could rename that too

rverdile: I put the 'api' package because handler and dao can't import each other. What about replacing 'RepositoryItem' and 'RepositoryRequest' with 'RepositoryAPI'?

jlsherrill: yeah, i guess moving all the api structs to 'api' could make sense, lets give that a try.  So i'd move RepositoryInfo & RepositoryCollectionResponse

jlsherrill: and we may want to move all the api structs out of handlers/api.go too.  We can leave that for another day though

rverdile: If I move RepositoryCollectionResponse, I think that means I have to move all of the api structs, since it's dependent on them. Would it make sense for api.go to live in 'api' instead of 'handler'?

jlsherrill: it may make sense for all the structs to live there.  I'm not sure it makes sense for the functions to?

rverdile:
Moved the other structs to 'api' package. OpenAPI not working yet for some reason.

Makefile
```diff
@@ -18,7 +18,7 @@ test-ci:
 	go test ./...
 
 openapi:
-	swag init --generalInfo api.go  --dir pkg/handler/
+	swag init --generalInfo api.go --dir pkg/api/
```
jlsherrill, line 21: i think you'd have to specify multiple directories?  (via comma separated list)

cmd/content-sources/main.go
```diff
@@ -50,7 +50,7 @@ func main() {
 	}
 
 	handler.RegisterRoutes(e)
-	err = e.Start(":8000")
+	err = e.Start(":8001")
```
jlsherrill, line 53: 8001?

pkg/dao/repository.go
```diff
@@ -0,0 +1,42 @@
...
+
+	dest := models.RepositoryConfiguration{}
+	result := db.DB.Where("org_id = ? AND url = ?", newRepo.OrgID, newRepo.URL).Find(&dest)
+	if result.Error != nil {
+		return result.Error
+	}
+	if result.RowsAffected > 0 {
+		return fmt.Errorf("repository with this URL already belongs to organization")
+	}
```
jlsherrill, line 36: would it make sense to just create a uniquness constraint and check for that error on create?

pkg/api/repository.go
```diff
@@ -0,0 +1,35 @@
...
+
+// Repository holds data received from, or returned by, a repositories API request or response
+type Repository struct {
+	UUID                string `json:"uuid"`
+	Name                string `json:"name"`
+	URL                 string `json:"url"`                                //URL of the remote yum repository
+	DistributionVersion string `json:"distribution_version" example:"7"`   //Version to restrict client usage to
+	DistributionArch    string `json:"distribution_arch" example:"x86_64"` //Architecture to restrict client usage to
+	AccountID           string `json:"account_id"`                         //Account ID of the owner
+	OrgID               string `json:"org_id"`                             //Organization ID of the owner
```
jlsherrill, line 13: we should probably mark UUID, AccountId, & OrgID as readonly, like this:  https://github.com/swaggo/swag/blob/93acd2ad51e848ef3c78b25cb3964745fecde054/testdata/format_dst/web/handler.go#L13

pkg/handler/repositories.go
```diff
@@ -59,12 +36,47 @@ func listRepositories(c echo.Context) error {
...
+	return c.JSON(200, collectionResponse(&api.RepositoryCollectionResponse{Data: repos}, c, total))
+}
+
+// CreateRepository godoc
+// @Summary      Create Repository
+// @ID           createRepository
+// @Description  create a repository
+// @Tags         repositories
+// @Accept       json
+// @Produce      json
```
jlsherrill, line 48: I guess you have to specify @Param  here for each parameter that can be passed?  like this:  https://github.com/swaggo/swag/blob/master/example/celler/controller/examples.go#L126

i don't see an easy way to just reuse the struct, but maybe i'm missing it?

pkg/handler/repositories.go
```diff
@@ -62,7 +37,48 @@ func listRepositories(c echo.Context) error {
...
+// @Param  name       query   string true "name of repository"
+// @Param  url        query   string true "url of repository"
+// @Param  version    query   string true "os version"
+// @Param  arch       query   string true "os architecture"
```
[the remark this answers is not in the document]
jlsherrill: ahh, so i think we can do something like this:  https://github.com/RedHatInsights/patchman-engine/blob/6b922b634df0da82760fe7dbade8524f563b8528/manager/controllers/baseline_create.go#L41

and specify the struct as what the body should be.  I missed that when looking through their docs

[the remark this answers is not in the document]
jlsherrill: and in that case they created a struct purely for the creation body, so maybe that makes sense?  (I'm not sure if it'll add account_id & org_id to creation options, which we wouldn't want).  Maybe the solution is to make a new struct

pkg/api/repository.go
```diff
@@ -0,0 +1,46 @@
...
+
+// CreateRepository holds data received from request to create repository
+type CreateRepository struct {
+	UUID                string `json:"uuid" readonly:"true" swaggerignore:"true"`
+	Name                string `json:"name"`
+	URL                 string `json:"url"`                                             //URL of the remote yum repository
+	DistributionVersion string `json:"distribution_version" example:"7"`                //Version to restrict client usage to
+	DistributionArch    string `json:"distribution_arch" example:"x86_64"`              //Architecture to restrict client usage to
+	AccountID           string `json:"account_id" readonly:"true" swaggerignore:"true"` //Account ID of the owner
+	OrgID               string `json:"org_id" readonly:"true" swaggerignore:"true"`     //Organization ID of the owner
```
jlsherrill, line 24: we don't want AccountID and OrgId on this struct, as its pulled from the headers and not passed by the user

jlsherrill: ohh i see you've got swaggerignore on it  👍🏻

cmd/content-sources/main.go
```diff
@@ -18,7 +18,7 @@ func main() {
 	}
 
 	handler.RegisterRoutes(echo)
-	err = echo.Start(":8000")
+	err = echo.Start(":8001")
```
jlsherrill, line 21: This is still 8001 ;)

rverdile: fixed

jlsherrill (approved):

PR #20, opened 2022-05-27
CONTENT-59: integrate viper for config loading

jlsherrill:


pkg/config/config.go
```diff
@@ -1,44 +1,124 @@
...
+		log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
+	}
+}
+
+func ConfigureEcho() *echo.Echo {
+	e := echo.New()
+	echoLogger := lecho.From(log.Logger,
+		//logger := lecho.New(
+		//	log.Logger,
+		lecho.WithLevel(1),
```
jlsherrill, line 112: oops gotta fix this

jlsherrill: figuring out why .From isn't setting the level properly :)

pkg/db/db.go
```diff
@@ -20,14 +20,13 @@ var DB *gorm.DB
...
 func GetUrl() string {
-	conf := config.Get()
 	return fmt.Sprintf(
 		"user=%s password=%s dbname=%s host=%s port=%d sslmode=disable",
-		conf.DBUser,
-		conf.DBPassword,
-		conf.DBName,
-		conf.DBHost,
-		conf.DBPort,
+		config.Config().Database.User,
```
Andrewgdewar, line 25: Should we be calling the config function every time we need this? I think the previous method of assigning a variable to conf (config.Config().Database) would be more ideal no?

pkg/config/config.go
```diff
@@ -1,44 +1,124 @@
...
+	if Config().Logging.Console {
+		log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
+	}
+}
+
+func ConfigureEcho() *echo.Echo {
+	e := echo.New()
+	echoLogger := lecho.From(log.Logger,
+		//logger := lecho.New(
+		//	log.Logger,
```
Andrewgdewar, line 111: Can likely delete this.

pkg/config/config.go
```diff
@@ -1,44 +1,124 @@
...
-	DBUser     string
-	DBPassword string
-	DBName     string
+type Configuration struct {
+	Database Database
+	Logging  Logging
+	Loaded   bool
 }
 
-func Get() *Config {
```
Andrewgdewar, line 18: config.Get() > config.Config()  
Facts... okay definitely #nit/preference. Thoughts?

pkg/db/db.go
Andrewgdewar: In my (javascript) mind, this is running the same function 5 times, when it only needs to be instantiated once.  Is there Golang magic that I am unaware of?

jlsherrill: Its largely for ease of use.  config.Config() lazily-loads the config and returns the struct.  If we refer to the struct directly, we have to make sure to call a Load() method prior to anytime that is needed, which means in every single test (or really every MainTest of every package).  It seemed 'nicer' to just call this to have it lazy-loaded, and then we never have to explicitly call Load()  anywhere (every command, every test, etc..)

jlsherrill: although i should have it return a pointer, and not the actual struct!   i'll make that change.  Let me know if you prefer us calling Load() directly everywhere and I can think about it a bit more

pkg/config/config.go
jlsherrill: fixed!

pkg/db/db.go
Andrewgdewar: I don't think we need Load() everywhere, if this is the drawback between the two ideas.

jlsherrill:
updated

pkg/db/db.go
jlsherrill: You have to have Load() called before any reference to config values, so you have to have Load() called early in the process everywhere (tests, commands, etc..)

jlsherrill: Although we have to do that with the database today.  So maybe having something like  `config.Bootstrap()` isn't a terrible idea?  It could handle both

pkg/db/db.go
```diff
@@ -20,14 +20,13 @@ var DB *gorm.DB
 
 // GetUrl Get database config and return url
 func GetUrl() string {
-	conf := config.Get()
```
Andrewgdewar, line 23: so this isn't possible to do: 
conf := config.Get().Database 

I think I don't get why we are doing this still.

Andrewgdewar: Everything else looks good, just asking this more for my own understanding.

Andrewgdewar (approved):
Ack!

pkg/db/db.go
Andrewgdewar: Does Load just have to be called once early in the chain? Like during an Init() function for example? or is it  everytime 
we access a value?

PR #25, opened 2022-06-02
CONTENT-51 - add put/patch for repositories

jlsherrill:
TODO: 

* [x] add tests
* [x] more docs

### DIFF
--- a/pkg/api/repository.go
+++ b/pkg/api/repository.go
@@ -2,8 +2,8 @@ package api
 
 import "github.com/content-services/content-sources-backend/pkg/models"
 
-// Repository holds data received from, or returned by, a repositories API request or response
-type Repository struct {
+// RepositoryResponse holds data received from, or returned by, a repositories API request or response
+type RepositoryResponse struct {
 	UUID                string `json:"uuid" readonly:"true"`
 	Name                string `json:"name"`
 	URL                 string `json:"url"`                                //URL of the remote yum repository
@@ -13,18 +13,38 @@ type Repository struct {
 	OrgID               string `json:"org_id" readonly:"true"`             //Organization ID of the owner
 }
 
-// CreateRepository holds data received from request to create repository
-type CreateRepository struct {
-	UUID                string `json:"uuid" readonly:"true" swaggerignore:"true"`
-	Name                string `json:"name"`
-	URL                 string `json:"url"`                                             //URL of the remote yum repository
-	DistributionVersion string `json:"distribution_version" example:"7"`                //Version to restrict client usage to
-	DistributionArch    string `json:"distribution_arch" example:"x86_64"`              //Architecture to restrict client usage to
-	AccountID           string `json:"account_id" readonly:"true" swaggerignore:"true"` //Account ID of the owner
-	OrgID               string `json:"org_id" readonly:"true" swaggerignore:"true"`     //Organization ID of the owner
+// RepositoryRequest holds data received from request to create repository
+type RepositoryRequest struct {
+	UUID                *string `json:"uuid" readonly:"true" swaggerignore:"true"`
+	Name                *string `json:"name"`
+	URL                 *string `json:"url"`                                             //URL of the remote yum repository
+	DistributionVersion *string `json:"distribution_version" example:"7"`                //Version to restrict client usage to
+	DistributionArch    *string `json:"distribution_arch" example:"x86_64"`              //Architecture to restrict client usage to
+	AccountID           *string `json:"account_id" readonly:"true" swaggerignore:"true"` //Account ID of the owner
+	OrgID               *string `json:"org_id" readonly:"true" swaggerignore:"true"`     //Organization ID of the owner
 }
 
-func (r *Repository) FromRepositoryConfiguration(repoConfig models.RepositoryConfiguration) {
+func (r *RepositoryRequest) FillDefaults() {
+	//Fill in default values in case of PUT request, doesn't have to be valid, let the db validate that
+	defaultName := ""
+	defaultUrl := ""
+	defaultVersion := ""
+	defaultArch := ""
+	if r.Name == nil {
+		r.Name = &defaultName
+	}
+	if r.URL == nil {
+		r.URL = &defaultUrl
+	}
+	if r.DistributionVersion == nil {
+		r.DistributionVersion = &defaultVersion
+	}
+	if r.DistributionArch == nil {
+		r.DistributionArch = &defaultArch
+	}
+}
+
+func (r *RepositoryResponse) FromRepositoryConfiguration(repoConfig models.RepositoryConfiguration) {
 	r.UUID = repoConfig.UUID
 	r.Name = repoConfig.Name
 	r.URL = repoConfig.URL
@@ -35,9 +55,9 @@ func (r *Repository) FromRepositoryConfiguration(repoConfig models.RepositoryCon
 }
 
 type RepositoryCollectionResponse struct {
-	Data  []Repository     `json:"data"`  //Requested Data
-	Meta  ResponseMetadata `json:"meta"`  //Metadata about the request
-	Links Links            `json:"links"` //Links to other pages of results
+	Data  []RepositoryResponse `json:"data"`  //Requested Data
+	Meta  ResponseMetadata     `json:"meta"`  //Metadata about the request
+	Links Links                `json:"links"` //Links to other pages of results
 }
 
 func (r *RepositoryCollectionResponse) SetMetadata(meta ResponseMetadata, links Links) {
--- a/pkg/dao/error.go
+++ b/pkg/dao/error.go
@@ -1,0 +1,11 @@
+package dao
+
+type Error struct {
+	Message       string
+	NotFound      bool
+	BadValidation bool
+}
+
+func (e *Error) Error() string {
+	return e.Message
+}
--- a/pkg/dao/repository.go
+++ b/pkg/dao/repository.go
@@ -1,43 +1,81 @@
 package dao
 
 import (
-	"fmt"
-
 	"github.com/content-services/content-sources-backend/pkg/api"
 	"github.com/content-services/content-sources-backend/pkg/db"
 	"github.com/content-services/content-sources-backend/pkg/models"
+	"github.com/jackc/pgconn"
 )
 
 type repositoryDaoImpl struct {
 }
 
-func GetRepositoryDao() repositoryDaoImpl {
+func GetRepositoryDao() RepositoryDao {
 	return repositoryDaoImpl{}
 }
 
-const uniqueConstraintErrMsg = "ERROR: duplicate key value violates unique constraint \"url_and_org_id_unique\" (SQLSTATE 23505)"
-
-func (r repositoryDaoImpl) Create(newRepo api.CreateRepository) error {
-
+func (r repositoryDaoImpl) Create(newRepo api.RepositoryRequest) error {
 	newRepoConfig := models.RepositoryConfiguration{
-		Name:      newRepo.Name,
-		URL:       newRepo.URL,
-		Version:   newRepo.DistributionVersion,
-		Arch:      newRepo.DistributionArch,
-		AccountID: newRepo.AccountID,
-		OrgID:     newRepo.OrgID,
+		AccountID: *newRepo.AccountID,
+		OrgID:     *newRepo.OrgID,
 	}
+	ApiFieldsToModel(&newRepo, &newRepoConfig)
 
-	dest := models.RepositoryConfiguration{}
-	result := db.DB.Where("org_id = ? AND url = ?", newRepo.OrgID, newRepo.URL).Find(&dest)
-	if result.Error != nil {
-		return result.Error
-	}
 	if err := db.DB.Create(&newRepoConfig).Error; err != nil {
-		if err.Error() == uniqueConstraintErrMsg {
-			return fmt.Errorf("repository with this URL already belongs to organization")
+		if isUniqueViolation(err) {
+			return &Error{BadValidation: true, Message: "RepositoryResponse with this URL already belongs to organization "}
 		}
 		return err
 	}
 	return nil
+}
+
+func (r repositoryDaoImpl) Fetch(orgId string, uuid string) api.RepositoryResponse {
+	repo := api.RepositoryResponse{}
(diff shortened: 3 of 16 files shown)


pkg/dao/repository.go
```diff
@@ -1,22 +1,19 @@
...
 	"github.com/content-services/content-sources-backend/pkg/db"
 	"github.com/content-services/content-sources-backend/pkg/models"
+	"github.com/jackc/pgconn"
 )
 
 type repositoryDaoImpl struct {
 }
 
-func GetRepositoryDao() repositoryDaoImpl {
+func GetRepositoryDao() RepositoryDao {
```
jlsherrill, line 13: I changed this return value to be the interface, that way other things could implement the interface and the handler code wouldn't care.

pkg/dao/error.go
```diff
@@ -0,0 +1,11 @@
...
+
+type Error struct {
+	Message       string
+	NotFound      bool
+	BadValidation bool
+}
+
+func (e *Error) Error() string {
+	return e.Message
+}
```
jlsherrill, line 11: this is a custom error struct that should allow us to pass more information up into the handler layer.  For example if we hit a uniquness constraint, we could tell the api that through this error struct.  that way it could return BadRequest rather than ISE 500

pkg/dao/repository.go
```diff
@@ -34,10 +31,52 @@ func (r repositoryDaoImpl) Create(newRepo api.CreateRepository) error {
...
+	repo := api.Repository{}
+	repo.FromRepositoryConfiguration(r.fetch(orgId, uuid))
+	return repo
+}
+
+func (r repositoryDaoImpl) fetch(orgId string, uuid string) models.RepositoryConfiguration {
+	found := models.RepositoryConfiguration{}
+	db.DB.First(&found, "UUID = ? AND ORG_ID = ?", uuid, orgId)
+	return found
+}
```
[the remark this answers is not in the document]
rverdile: I think fetchRepoConfig is a better name for it at the moment. But in the future, would it fetch things other than RepositoryConfiguration?

pkg/dao/repository.go
```diff
@@ -34,10 +31,52 @@ func (r repositoryDaoImpl) Create(newRepo api.CreateRepository) error {
...
+		return &Error{Message: result.Error.Error(), BadValidation: isUniqueViolation(result.Error)}
+	}
+	return nil
+}
+
+func ApiToModel(apiRepo *api.CreateRepository, repoConfig *models.RepositoryConfiguration) {
+	repoConfig.Name = apiRepo.Name
+	repoConfig.Arch = apiRepo.DistributionArch
+	repoConfig.Version = apiRepo.DistributionVersion
+	repoConfig.URL = apiRepo.URL
```
rverdile, line 81: is leaving out AccountID and OrgID intentional?

pkg/dao/repository.go
```diff
@@ -34,10 +31,52 @@ func (r repositoryDaoImpl) Create(newRepo api.CreateRepository) error {
...
+	pgError, ok := err.(*pgconn.PgError)
+	if ok {
+		if pgError.Code == "23505" {
+			return true
+		}
+	}
+	return false
+}
+
+func (r repositoryDaoImpl) FullUpdate(orgId string, uuid string, repoParams api.CreateRepository) error {
```
rverdile, line 64: What does "full" update mean? Is there going to be a "partial" update?

pkg/dao/repository.go
```diff
@@ -34,10 +31,52 @@ func (r repositoryDaoImpl) Create(newRepo api.CreateRepository) error {
...
-			return fmt.Errorf("repository with this URL already belongs to organization")
+		if isUniqueViolation(err) {
+			return &Error{BadValidation: true, Message: "Repository with this URL already belongs to organization "}
 		}
 		return err
 	}
 	return nil
 }
+
+func (r repositoryDaoImpl) Fetch(orgId string, uuid string) api.Repository {
```
rverdile, line 42: Is this PR also going to add a GET for /repositories/:uuid?

pkg/handler/repositories.go
```diff
@@ -68,24 +71,68 @@ func listRepositories(c echo.Context) error {
...
+	}
+
+	err = dao.FullUpdate(OrgID, uuid, repoParams)
+	if err != nil {
+		return echo.NewHTTPError(codeForError(err), err.Error())
+	}
+	return c.String(http.StatusOK, "Repository Updated.\n")
+}
+
+func codeForError(err error) int {
```
rverdile, line 121: what about naming this httpCodeForError, httpErrorCode, or something with "http" in it?

pkg/dao/repository.go
jlsherrill: Yeah, PUT does a full update, patch does a partial update (only fields passed in).

pkg/dao/repository.go
jlsherrill: This is kinda on purpose.  You don't want the user to be able to pass in org id or account id in the request

pkg/handler/repositories.go
```diff
@@ -68,24 +71,68 @@ func listRepositories(c echo.Context) error {
...
+// @Accept       json
+// @Produce      json
+// @Param  body       body    api.CreateRepository true  "request body"
+// @Param  org_id     header  string         	   true  "organization id"
+// @Param  account_id header  string               true  "account number"
+// @Success      201
+// @Router       /repositories/{uuid} [put]
+func fullUpdate(c echo.Context) error {
+	uuid := c.Param("uuid")
+	repoParams := api.CreateRepository{}
```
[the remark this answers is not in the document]
jlsherrill: yeah, i was trying to think of a better name. I'll continue :)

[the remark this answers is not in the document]
jlsherrill: How about  RepositoryRequest  as it is the Request coming in for a repository, either for create, or update

pkg/dao/repository.go
jlsherrill: added!

pkg/handler/repositories.go
```diff
@@ -60,53 +64,128 @@ func listRepositories(c echo.Context) error {
 // @Tags         repositories
 // @Accept       json
 // @Produce      json
-// @Param  body       body    api.CreateRepository true  "request body"
-// @Param  org_id     header  string         	   true  "organization id"
-// @Param  account_id header  string               true  "account number"
```
jlsherrill, line 65: FYI i removed these org_id & account_id header declarations.  They aren't strictly correct, since it has to be within the identity header.  I don't think they need to be here since the user would actually be passing basic auth with their request and the 3scale infrastructure would add in the identity header, but we can see what other apps are doing

pkg/handler/repository_test.go
```diff
@@ -9,17 +9,77 @@ import (
...
+		return rr
+	} else {
+		return api.RepositoryResponse{}
+	}
+}
+
+func (r *MockRepositoryDao) Update(orgId string, uuid string, repoParams api.RepositoryRequest) error {
+	args := r.Called(orgId, uuid, repoParams)
+	return args.Error(0)
+}
```
jlsherrill, line 49: This is a 'mock' version of the dao layer, it lets us test the api without actually invoking the database.  So the goal is for the dao to have a very well defined interface, and we can test each layer separately.

pkg/handler/repositories.go
```diff
@@ -9,14 +9,24 @@ import (
...
 	"github.com/content-services/content-sources-backend/pkg/models"
+	"github.com/davecgh/go-spew/spew"
 	"github.com/labstack/echo/v4"
 	"github.com/redhatinsights/platform-go-middlewares/identity"
 )
 
-func RegisterRepositoryRoutes(engine *echo.Group) {
+type RepositoryHandler struct {
+	RepositoryDao dao.RepositoryDao
+}
```
jlsherrill, line 19: so now the handler is an object, which lets you instantiate the handler with our mocked Dao and run requests against it.

pkg/dao/repository.go
```diff
@@ -1,43 +1,81 @@
...
+	return false
+}
+
+func (r repositoryDaoImpl) Update(orgId string, uuid string, repoParams api.RepositoryRequest) error {
+	repoConfig := r.fetchRepoConfig(orgId, uuid)
+	if repoConfig.UUID == "" {
+		return &Error{NotFound: true, Message: "Could not find RepositoryResponse with uuid " + uuid}
+	}
+	ApiFieldsToModel(&repoParams, &repoConfig)
+	result := db.DB.Model(&repoConfig).Updates(repoConfig)
```
rverdile, line 61: Updating with a struct won't update zero'd fields. Maybe we need a "convert to map" utility?

rverdile, line 61: Updating with a struct won't update zero'd fields. Maybe we need a "convert to map" utility?

pkg/handler/repositories.go
```diff
@@ -60,53 +70,127 @@ func listRepositories(c echo.Context) error {
...
+		return c.JSON(200, response)
 	}
-	newRepository.AccountID = AccountID
-	newRepository.OrgID = OrgID
+}
 
-	repositoryDao := dao.GetRepositoryDao()
-	if err := repositoryDao.Create(newRepository); err != nil {
-		return echo.NewHTTPError(http.StatusInternalServerError, "error creating repository: "+err.Error())
+// FullUpdateRepository godoc
```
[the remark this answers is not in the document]
rverdile: "this" being Full Update

pkg/handler/repositories.go
```diff
@@ -9,14 +9,24 @@ import (
...
+func RegisterRepositoryRoutes(engine *echo.Group, rDao *dao.RepositoryDao) RepositoryHandler {
+	rh := RepositoryHandler{RepositoryDao: *rDao}
 	engine.GET("/repositories/", listRepositories)
+	engine.GET("/repositories/:uuid", rh.fetch)
+	engine.PUT("/repositories/:uuid", rh.fullUpdate)
+	engine.PATCH("/repositories/:uuid", rh.partialUpdate)
 	engine.DELETE("/repositories/:uuid", deleteRepository)
-	engine.POST("/repositories/", createRepository)
+	engine.POST("/repositories/", rh.createRepository)
+	return rh
```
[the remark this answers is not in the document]
jlsherrill: you're right, I think its not explicitly needed to be passed in or returned for the current tests.  I think it made sense for me to pass in the dao from a central place rather than have each handler instantiate the dao itself.  

If we move to these style tests:  https://github.com/content-services/content-sources-backend/blob/main/pkg/handler/api_test.go#L15-L36  it probably would need to be passed in at least, but not returned.   I'll remove the return but would probably prefer to keep it being passed in unless you object

pkg/dao/repository.go
jlsherrill: yeah, i solved this basically with what you suggested

pkg/handler/repositories.go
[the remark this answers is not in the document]
jlsherrill: yeah i think the 'put' instead of patch was messing this up

pkg/handler/repositories.go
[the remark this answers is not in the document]
rverdile: We can leave it as being passed in. I would also just change the name here from dao to rDao, so it doesn't conflict with the import: https://github.com/content-services/content-sources-backend/blob/5390ddc35a58951c711ca0bb346ea29cb9bf35c1/pkg/handler/api.go#L51-L52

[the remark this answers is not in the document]
jlsherrill: I imagine in the future, we could have a 'master dao' object, that you could pass around too instead of a dao object for each type, but another thought for another day

pkg/models/repository_configuration.go
```diff
@@ -9,3 +9,15 @@ type RepositoryConfiguration struct {
...
+//When updating a model with gorm, we want to explicitly update any field that is set to
+// empty string.  We always fetch the object and then update it before saving
+// so every update is the full model of user changable fields.
+// So OrgId and account Id or excluded
+func (rc *RepositoryConfiguration) MapForUpdate() map[string]interface{} {
+	forUpdate := make(map[string]interface{})
+	forUpdate["Name"] = rc.Name
+	forUpdate["URL"] = rc.URL
+	forUpdate["Arch"] = rc.Arch
+	return forUpdate
```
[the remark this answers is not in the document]
jlsherrill: doh!

pkg/models/repository_configuration.go
```diff
@@ -9,3 +9,15 @@ type RepositoryConfiguration struct {
 	AccountID string `json:"account_id" gorm:"default:null"`
 	OrgID     string `json:"org_id" gorm:"default:null"`
 }
+
+//When updating a model with gorm, we want to explicitly update any field that is set to
+// empty string.  We always fetch the object and then update it before saving
+// so every update is the full model of user changable fields.
+// So OrgId and account Id or excluded
```
jlsherrill, line 16: haha! Go my thwart my terrible spelling but it will never overcome my horrendous grammar!

pkg/api/repository.go
```diff
@@ -13,18 +13,38 @@ type Repository struct {
...
+	AccountID           *string `json:"account_id" readonly:"true" swaggerignore:"true"` //Account ID of the owner
+	OrgID               *string `json:"org_id" readonly:"true" swaggerignore:"true"`     //Organization ID of the owner
+}
+
+func (r *RepositoryRequest) FillDefaults() {
+	//Fill in default values in case of PUT request, doesn't have to be valid, let the db validate that
+	defaultName := ""
+	defaultUrl := ""
+	defaultVersion := ""
+	defaultArch := ""
```
rverdile, line 32: Right now the code isn't blocking "" URLs or "" Names. This might be the place to fix it?

jlsherrill: I thought the model might be a better place? but i'm not sure.  I'll file a bug around it and we can take a look

jlsherrill: filed!

jlsherrill:
opened a 2nd pr for the docs update i was intending to do: https://github.com/content-services/content-sources-backend/pull/27

jlsherrill:
https://issues.redhat.com/browse/14674583

jlsherrill:

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-71

PR #28, opened 2022-06-10
CONTENT-68 - support multiple versions for repo config

jlsherrill:


Andrewgdewar (approved):
LGTM, testing with UI now.

db/migrations/20220512132042_create_repository_configurations_table.up.sql
```diff
@@ -1,10 +1,10 @@
 BEGIN;
 
 CREATE TABLE IF NOT EXISTS repository_configurations(
-    uuid VARCHAR (255) UNIQUE NOT NULL PRIMARY KEY,
+    uuid UUID UNIQUE NOT NULL PRIMARY KEY,
```
jlsherrill, line 4: note i chnaged this to use the UUID type in postgresql, it may have some slight performance advantages

pkg/models/repository_configuration.go
```diff
@@ -1,13 +1,15 @@
...
 	Base
-	Name      string `json:"name" gorm:"default:null"`
-	URL       string `json:"url" gorm:"default:null"`
-	Version   string `json:"version" gorm:"default:null"`
-	Arch      string `json:"arch" gorm:"default:null"`
-	AccountID string `json:"account_id" gorm:"default:null"`
-	OrgID     string `json:"org_id" gorm:"default:null"`
+	Name      string         `json:"name" gorm:"default:null"`
+	URL       string         `json:"url" gorm:"default:null"`
+	Versions  pq.StringArray `json:"version" gorm:"type:text[],default:null"`
```
rverdile, line 9: what's the difference between pq.StringArray and normal go array?

jlsherrill: for Gorm, you can't use a normal go string array, it doesn't know how to map it to postgres's varchar array.  This special type implements some methods that gorm needs to translate it properly (https://gorm.io/docs/data_types.html) .  That said, its just a string array under the hood:  https://github.com/lib/pq/blob/master/array.go#L626   and as you'll see in the other code, you can assign a string array to it.

rverdile: thanks!

Andrewgdewar (approved):
Still looks good, was able to get this up and running locally.

PR #31, opened 2022-06-10
Fix differences between podman and docker

jlsherrill:
docker doesn't support 'container exists', so use inspect instead.
Docker and podman use different health text in inspect, so check both

### DIFF
--- a/mk/db.mk
+++ b/mk/db.mk
@@ -8,8 +8,8 @@
 .PHONY: db-up
 db-up: DOCKER_IMAGE=docker.io/postgres:14
 db-up: $(GO_OUTPUT)/dbmigrate  ## Start postgres database
-	$(DOCKER) volume exists postgres || $(DOCKER) volume create postgres
-	$(DOCKER) container exists postgres || $(DOCKER) run \
+	$(DOCKER) volume inspect postgres &> /dev/null || $(DOCKER) volume create postgres
+	$(DOCKER) container inspect postgres &> /dev/null || $(DOCKER) run \
 	  -d \
 	  --rm \
 	  --name postgres \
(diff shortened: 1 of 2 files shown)


jlsherrill:
@Andrewgdewar if you can check this works for you too, that'd be helpful!

jlsherrill:
@avisiedo   @Andrewgdewar is using docker on an M1 mac, and ran into some issues, so i think i fixed them here.

mk/db.mk
```diff
@@ -9,7 +9,7 @@
 db-up: DOCKER_IMAGE=docker.io/postgres:14
 db-up: $(GO_OUTPUT)/dbmigrate  ## Start postgres database
 	$(DOCKER) volume exists postgres || $(DOCKER) volume create postgres
```
avisiedo, line 11:
```suggestion
	$(DOCKER) volume inspect postgres &> /dev/null || $(DOCKER) volume create postgres
```

mk/db.mk
```diff
@@ -30,13 +30,13 @@ db-up: $(GO_OUTPUT)/dbmigrate  ## Start postgres database
...
-	@$(DOCKER) container exists postgres
+	@$(DOCKER) container inspect postgres &> /dev/null
 	@$(DOCKER) exec postgres pg_isready
 
 .PHONY: .db-health-wait
 .db-health-wait:
-	@$(DOCKER) container exists postgres
-	@while [ "$$($(DOCKER) inspect -f '{{.State.Healthcheck.Status}}' postgres)" != "healthy" ]; do echo -n "."; sleep 1; done
+	@$(DOCKER) container inspect postgres &> /dev/null
+	@while [ "$$($(DOCKER) inspect -f '{{.State.Health.Status}}' postgres 2> /dev/null | $(DOCKER) inspect -f '{{.State.Healthcheck.Status}}' postgres 2> /dev/null)" != "healthy" ]; do echo -n "."; sleep 1; done
```
avisiedo, line 39:
```suggestion
ifeq (podman,$(DOCKER))
DOCKER_HEALTH_PATH := .State.Healthcheck.Status
else
DOCKER_HEALTH_PATH := .State.Health.Status
endif
.PHONY: .db-health-wait
.db-health-wait:
	@$(DOCKER) container inspect postgres &> /dev/null
	@while [ "$$($(DOCKER) inspect -f '{{$(DOCKER_HEALTH_PATH)}}' postgres 2> /dev/null)" != "healthy" ]; do printf "."; sleep 1; done
```

Comments about the suggested change:
- Decide the `DOCKER_HEALTH_PATH` value depending on the `DOCKER` value.
- Use that in the command to check the healthy state.
- Update `echo -n "."` by `printf "."`. I got issues when running that command in macos (x86_64) `sh -c 'echo -n "."'` (ignoring `-n` option, and printing `-n .` with a new line at the end, instead of just `.` without new line at the end.

avisiedo:
@jlsherrill @Andrewgdewar I have added some suggestions, I could check in macos, but not with a m1 core; maybe more things could change.

jlsherrill:
@Andrewgdewar can you give this a spin and db-down  db-clean  and db-up again to make sure it still works for you?

Andrewgdewar (approved):
Confirmed this is working for me. Feel free to get other approvals and/or ping me if further changes are made. ACK.

PR #32, opened 2022-06-13
 CONTENT-78: Add Filters to list Api

Andrewgdewar:
- [x] Tests

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -12,6 +12,14 @@ type PaginationData struct {
 	Offset int `query:"offset" json:"offset"` //Offset into the total results
 }
 
+type FilterData struct {
+	Search              string `query:"search" json:"search" `                              //Search string based query to optionally filter-on
+	Arch                string `query:"arch" json:"arch" `                                  //Architecture to optionally filter-on (IE 'x86_64' would return Repositories with the 'x86_64' only)
+	Version             string `query:"version" json:"version"`                             //Version to optionally filter-on  (IE 7 would return Repositories with the version 7 only)
+	AvailableForArch    string `query:"available_for_arch" json:"available_for_arch"`       //Filter by compatible arch (IE 'x86_64' would return Repositories with the 'x86_64' arch and Repositories where arch is not set)
+	AvailableForVersion string `query:"available_for_version" json:"available_for_version"` //Filter by compatible version (IE 7 would return Repositories with the version 7 or where version is not set)
+}
+
 type ResponseMetadata struct {
 	Limit  int   `query:"limit" json:"limit"`   //Limit of results used for the request
 	Offset int   `query:"offset" json:"offset"` //Offset into results used for the request
--- a/pkg/dao/interfaces.go
+++ b/pkg/dao/interfaces.go
@@ -8,6 +8,6 @@ type RepositoryDao interface {
 	Create(newRepo api.RepositoryRequest) error
 	Update(orgID string, uuid string, repoParams api.RepositoryRequest) error
 	Fetch(orgID string, uuid string) (api.RepositoryResponse, error)
-	List(orgID string, limit int, offset int) (api.RepositoryCollectionResponse, int64, error)
+	List(orgID string, paginationData api.PaginationData, filterData api.FilterData) (api.RepositoryCollectionResponse, int64, error)
 	Delete(orgID string, uuid string) error
 }
--- a/pkg/handler/api.go
+++ b/pkg/handler/api.go
@@ -20,6 +20,11 @@ import (
 
 const DefaultOffset = 0
 const DefaultLimit = 100
+const DefaultSearch = ""
+const DefaultArch = ""
+const DefaultVersion = ""
+const DefaultAvailableForArch = ""
+const DefaultAvailableForVersion = ""
 const DefaultAppName = "content_sources"
 const MaxLimit = 200
 const ApiVersion = "1.0"
@@ -96,9 +101,27 @@ func createLink(c echo.Context, offset int) string {
 	req := c.Request()
 	q := req.URL.Query()
 	page := ParsePagination(c)
+	filters := ParseFilters(c)
 
 	q.Set("limit", strconv.Itoa(page.Limit))
 	q.Set("offset", strconv.Itoa(offset))
+
+	if filters.Search != "" {
+		q.Set("search", filters.Search)
+	}
+	if filters.Arch != "" {
+		q.Set("arch", filters.Arch)
+	}
+	if filters.Version != "" {
+		q.Set("version", filters.Version)
+	}
+	if filters.AvailableForArch != "" {
+		q.Set("available_for_arch", filters.AvailableForArch)
+	}
+
+	if filters.AvailableForVersion != "" {
+		q.Set("available_for_version", filters.AvailableForVersion)
+	}
 
 	params, _ := url.PathUnescape(q.Encode())
 	return fmt.Sprintf("%v?%v", req.URL.Path, params)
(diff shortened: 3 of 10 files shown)


pkg/seeds/seeds.go
```diff
@@ -16,19 +17,40 @@ type SeedOptions struct {
...
+			}
+		}
+
+		randomAmount := rand.Intn(4) + 1
+		allowMultipleVersion := rand.Intn(10) >= 9
+		versionArray := []string{}
+		newValue := rand.Intn(9)
+		for k := 0; k < randomAmount && (len(versionArray) == 0 || allowMultipleVersion); k++ {
+			versionArray = append(versionArray, fmt.Sprintf("%d", newValue+k+1))
+		}
```
jlsherrill, line 38: maybe extract this to a new function?

pkg/handler/api.go
```diff
@@ -96,10 +99,13 @@ func createLink(c echo.Context, offset int) string {
...
 	q := req.URL.Query()
 	page := ParsePagination(c)
+	filters := ParseFilters(c)
 
 	q.Set("limit", strconv.Itoa(page.Limit))
 	q.Set("offset", strconv.Itoa(offset))
-
+	q.Set("search", filters.Search)
+	q.Set("arch", filters.Arch)
+	q.Set("version", filters.Version)
```
jlsherrill, line 108: should these just be omitted if they are blank?

Andrewgdewar: Yes. Will do some if statement magic.

pkg/dao/interfaces.go
```diff
@@ -8,6 +8,13 @@ type RepositoryDao interface {
...
 	Update(orgID string, uuid string, repoParams api.RepositoryRequest) error
 	Fetch(orgID string, uuid string) (api.RepositoryResponse, error)
-	List(orgID string, limit int, offset int) (api.RepositoryCollectionResponse, int64, error)
+	List(
+		orgID string,
+		limit int,
+		offset int,
+		search string,
+		arch string,
+		version string,
```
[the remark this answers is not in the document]
Andrewgdewar: Yes, though we would then want to do the same with paginationData no?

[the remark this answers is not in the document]
jlsherrill: great idea!

jlsherrill:
wanna switch this to ready for review?  
Mind updating to a better commit message?

Andrewgdewar:
> 

Will do, have to update a few more tests with alternate data, then it'll be good.

pkg/handler/api.go
```diff
@@ -169,3 +183,21 @@ func ParsePagination(c echo.Context) api.PaginationData {
...
+		Arch:    DefaultArch,
+		Version: DefaultVersion}
+	err := echo.QueryParamsBinder(c).
+		String("search", &filterData.Search).
+		String("arch", &filterData.Arch).
+		String("version", &filterData.Version).
+		BindError()
+
+	if err != nil {
+		log.Fatal().Err(err)
```
jlsherrill, line 199: log.Fatal() actually causes the program to exit after the message is logged which is bad! I see that ParsePagination is doing that too, mind updating both to log.Error()?

Andrewgdewar: Roger. Still dealing with seed changes on tests.

pkg/api/api.go
```diff
@@ -12,6 +12,14 @@ type PaginationData struct {
 	Offset int `query:"offset" json:"offset"` //Offset into the total results
 }
 
+type FilterData struct {
+	Search              string `query:"search" json:"search" `                              //Search string based query to optionally filter-on
+	Arch                string `query:"arch" json:"arch" `                                  //architecture to optionally filter-on
+	Version             string `query:"version" json:"version"`                             //Version to optionally filter-on
+	AvailableForArch    string `query:"available_for_arch" json:"available_for_arch"`       //Filter by compatible arch (IE x86_64 and x86_64 or x86_64 and "Any Arch")
+	AvailableForVersion string `query:"available_for_version" json:"available_for_version"` //Filter by compatible version (IE 7 and 7 or 7 and "Any Version")
```
rverdile, line 20: what are these for? how are these different from the other arch/version queries?

pkg/seeds/seeds.go
```diff
@@ -16,19 +17,16 @@ type SeedOptions struct {
...
 			Name:      fmt.Sprintf("%s - %s - %s", RandStringBytes(2), "TestRepo", RandStringBytes(10)),
 			URL:       fmt.Sprintf("https://%s.com/%s", RandStringBytes(20), RandStringBytes(5)),
-			Versions:  []string{"9"},
-			Arch:      "x86_64",
-			AccountID: fmt.Sprintf("%d", rand.Intn(9999)),
-			OrgID:     options.OrgID,
+			Versions:  createVersionArray(i),
+			Arch:      createArch(i),
+			AccountID: strconv.Itoa(rand.Intn(9999)),
+			OrgID:     createOrgId(options.OrgID, i),
```
rverdile, line 27: If I'm understanding this right, you're using the index of the for loop to help create random values. Would it be easier to create random values without depending on the for loop and keep the randomness entirely within the function?

Andrewgdewar: This was intentional, I want predictable randomness for tests. Without the addition of the index, the outcomes of repo counts on particular filters were variable. Now they are not.

pkg/api/api.go
Andrewgdewar: arch/version filter on those values, the "availableFor" variants basically do the same thing with the caveat that they will also match to "Empty" states, which represent "Any version"/"Any Arch" within the repos.

pkg/seeds/seeds.go
rverdile: Would that predictable randomness be, for example, if you wanted your seeded data to have 5 repos with version 9? Something like that might be better as some kind of seedOptions variable, but I'm not sure what that would look like.

Also, I believe the randomness here depends on how many repos you seed at a time, which feels odd. Maybe I just want to seed a few repos for local testing.

But we don't have to block the PR on this. I'm assuming it works okay if the tests are passing!

pkg/api/api.go
Andrewgdewar: They could, this is a design question though, as we likely want the ability to match to both. It depends on what the API user/UI requires I suppose.

jlsherrill: Version & Arch are meant to be 'filters' to filter the repo list to a certain state, so you may want to filter 'any arch' vs 'x86_64'.  "AvailableForArch" is meant to be used by other apps like Image Builder that want "All repos applicable for building an image for x86_64", which would include 'x86_64' and 'any arch' repos.  Suggestions welcome for better naming :)

pkg/seeds/seeds.go
Andrewgdewar: It does work as today, most tests use a sample size of 20 in the DB, and this seed allows for at least 1 of each type filter within that small of sample.

pkg/api/api.go
Andrewgdewar: I had originally suggested a boolean value for each item to match to blank (any) variants of the filter. This outcome was determined by @jlsherrill. I'm open to a change if it is something that makes sense.

rverdile: I get it now. I don't have a better name for the variable, but maybe the description for it should have an example more like justin's? 'e.g. "x86_64" returns items with "x86_64" or "Any Arch" '

rverdile: It might also make sense to have an example for arch like ' "x86_64" returns repos with "x86_64" `, so there's an obvious distinction between the two.

pkg/api/api.go
```diff
@@ -12,6 +12,14 @@ type PaginationData struct {
 	Offset int `query:"offset" json:"offset"` //Offset into the total results
 }
 
+type FilterData struct {
+	Search              string `query:"search" json:"search" `                              //Search string based query to optionally filter-on
+	Arch                string `query:"arch" json:"arch" `                                  //architecture to optionally filter-on
+	Version             string `query:"version" json:"version"`                             //Version to optionally filter-on
+	AvailableForArch    string `query:"available_for_arch" json:"available_for_arch"`       //Filter by compatible arch (IE x86_64 and x86_64 or x86_64 and "Any Arch")
```
jlsherrill, line 19: Filter by compatible arch for client consumption (IE 'x86_64' would return Repositories with the 'x86_64' arch and Repositories where arch is not set)

jlsherrill: does that sound better?

rverdile: yeah I think that's more clear

Andrewgdewar:
Changes pushed around updating the naming.

pkg/api/api.go
```diff
@@ -12,6 +12,14 @@ type PaginationData struct {
 	Offset int `query:"offset" json:"offset"` //Offset into the total results
 }
 
+type FilterData struct {
+	Search              string `query:"search" json:"search" `                              //Search string based query to optionally filter-on
+	Arch                string `query:"arch" json:"arch" `                                  //Architecture to optionally filter-on (IE 'x86_64' would return Repositories with the 'x86_64' only)
+	Version             string `query:"version" json:"version"`                             //Version to optionally filter-on  (IE 7 would return Repositories with the version 7 only)
+	AvailableForArch    string `query:"available_for_arch" json:"available_for_arch"`       //Filter by compatible arch (IE 'x86_64' would return Repositories with the 'x86_64' arch and Repositories where arch is not set)
+	AvailableForVersion string `query:"available_for_version" json:"available_for_version"` //Filter by compatible version (IE 7 would return Repositories with the version 7 or where version is not set)
```
rverdile, line 20: nice!

PR #36, opened 2022-06-15
Update Dependencies

jlsherrill:


Andrewgdewar:
Did you delete your go.sum before updating? I find it cleans out old values best that way. Not sure if it's best practice.

Andrewgdewar (commented):
After pulling this into my branch and "go get ./..." and "go mod tidy" I see a bit of a different output. 
Did you run tidy on this?

jlsherrill:
good call, updated!

Andrewgdewar (approved):
No changes when I updated! ACK

PR #40, opened 2022-06-21
proposal:configuration which allow to run and debug unit tests from VSCode

avisiedo:
- Set configuration for .gitignore
- For tests, set CONFIG_PATH to point out the current 'configs` direcotory
  using the absolute workspace path.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "go.testEnvVars": {
+        "CONFIG_PATH": "${workspaceFolder}/configs"
+    }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ release/
 .idea
 .env
 configs/config.yaml
+
+# Visual Studio Code settings
+.vscode/**
+!.vscode/settings.json


jlsherrill:
@Andrewgdewar mind looking at this since you use vscode?

Andrewgdewar:
> @Andrewgdewar mind looking at this since you use vscode?

Will do.

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "go.testEnvVars": {
+        "CONFIG_PATH": "${workspaceFolder}/configs"
```
Andrewgdewar, line 3: Would "./configs" not work here?

For the .vscode folder to work I thought you had to have the project folder open anyway.

avisiedo: Sorry for the delay mate! I was getting an error when launching the tests in visual studio code; by the situation that I got and the observations, from the ide the test binaries are generated in each package directory; when the tests are run from the ide, the current working directory is changed to the package directory, making the relative reference wrong.

It happens when we try to debug a test from the test lab view. I attach a screenshot

avisiedo: image

Andrewgdewar (approved):
Tested, LGTM

avisiedo:
@Andrewgdewar thanks for the review!

PR #41, opened 2022-06-22
Properly exit during db migration errors

jlsherrill:


jlsherrill:
From the zerolog docs:  
`
It is very important to note that when using the zerolog chaining API, as shown above (log.Info().Msg("hello world"), the chain must have either the Msg or Msgf method call. If you forget to add either of these, the log will not occur and there is no compile time error to alert you of this.`

rverdile (approved):
Looks good and didn't find any other places missing "Msg"

PR #38, opened 2022-06-17
Fixes 90: commands to download and load external repos

jlsherrill:
This adds a couple commands:

```
Manage external repo information
  repos-download   Download external repo urls from Image Builder
  repos-import     Import External repo urls from Image Builders into the DB
```

so you can run:

```
rm ./pkg/external_repos/external_repos.json    #clear out the existing cache
make repos-download  
```
the file should be recreated exactly

```
make repos-import
```

check your db and see the repos import:
```
make db-cli-connect


 select * from repositories ;
```

This also adds a "public" field to repositories, to mark repos that are available to all accounts, regardless of presence of repository_configuration

PR #42, opened 2022-06-22
Fixes 47: add api for listing rpms in a repository

avisiedo:
As a developer I want to retrieve the packages that belong to a repository that meet with:
- An api endpoint "/repositories/:uuid/rpms" where uuid is the id for the repository.
- A new table repositoriesand repository_configurations table relate to it.
- A new table repository_rpms where the rpm information will be stored with:
  - a uuid that identify the register.
  - the rpm information we want to store into the database.
  - a foreign key that relate to the repositories table that the rpm belong to.
- Add gorm models.
- The migration scripts for the above changes.
- Add the new table to 'dbmigrate seed' so that we can populate with random data.
- Tests.


db/migrations/20220610144605_hmscontent_47.up.sql
```diff
@@ -0,0 +1,44 @@
...
+CREATE TABLE IF NOT EXISTS repositories (
+    uuid UUID UNIQUE NOT NULL PRIMARY KEY,
+    created_at TIMESTAMP WITH TIME ZONE,
+    updated_at TIMESTAMP WITH TIME ZONE,
+    url VARCHAR(255) NOT NULL,
+    last_read_time TIMESTAMP WITH TIME ZONE DEFAULT NULL,
+    last_read_error VARCHAR(255) DEFAULT NULL,
+    refer_repo_config UUID DEFAULT NULL
+);
+
```
jlsherrill, line 15: we want a uniquness constraint on url i belive, since all repo_configs with the same url would share one repository entry

db/migrations/20220610144605_hmscontent_47.up.sql
```diff
@@ -0,0 +1,44 @@
...
+    name VARCHAR(255) NOT NULL,
+    arch VARCHAR(255) NOT NULL,
+    version VARCHAR(255) NOT NULL,
+    release VARCHAR(255),
+    epoch INTEGER DEFAULT 0 NOT NULL,
+    summary VARCHAR(255) NOT NULL,
+    description TEXT NOT NULL,
+    refer_repo UUID NOT NULL
+);
+
```
jlsherrill, line 38: we probably want a uniquness constraint on  (refer_repo, name, arch, version, release, epoch)

db/migrations/20220610144605_hmscontent_47.up.sql
```diff
@@ -0,0 +1,44 @@
...
+    created_at TIMESTAMP WITH TIME ZONE,
+    updated_at TIMESTAMP WITH TIME ZONE,
+    name VARCHAR(255) NOT NULL,
+    arch VARCHAR(255) NOT NULL,
+    version VARCHAR(255) NOT NULL,
+    release VARCHAR(255),
+    epoch INTEGER DEFAULT 0 NOT NULL,
+    summary VARCHAR(255) NOT NULL,
+    description TEXT NOT NULL,
+    refer_repo UUID NOT NULL
```
jlsherrill, line 36: normally (in my past projects), we'd use something like  'repository_uuid' to represent the reference to the repositories table, i'm curious about your selection of 'refer_repo'

pkg/api/repositoryRpm.go
```diff
@@ -0,0 +1,60 @@
...
+	if m != nil {
+		r.UUID = m.UUID
+		r.Name = m.Name
+		r.Arch = m.Arch
+		r.Version = m.Version
+		r.Release = m.Release
+		r.Epoch = pointy.Int32(*m.Epoch)
+		r.Summary = m.Summary
+		r.Description = m.Description
+	}
```
jlsherrill, line 51: we try to keep the model layer out of the api, and let the dao layer convert between model to api struct

pkg/dao/repository_test.go
```diff
@@ -38,31 +16,31 @@ func (suite *ReposSuite) TestCreate() {
 	t := suite.T()
 
 	found := models.RepositoryConfiguration{}
-	db.DB.First(&found)
+	suite.tx.First(&found)
 
-	err := GetRepositoryDao().Create(api.RepositoryRequest{
+	err := GetRepositoryDao(suite.tx).Create(api.RepositoryRequest{
```
jlsherrill, line 21: i like this solution of passing the transaction into the dao, this is a much better way!

pkg/api/repositoryRpm.go
```diff
@@ -0,0 +1,60 @@
...
+	Arch string `json:"arch"`
+	// The version for this package
+	Version string `json:"version"`
+	// The release for this package
+	Release string `json:"release"`
+	// Epoch is a way to define weighted dependencies based
+	// on version numbers. It's default value is 0 and this
+	// is assumed if an Epoch directive is not listed in the RPM SPEC file.
+	// https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/packaging_and_distributing_software/advanced-topics#packaging-epoch_epoch-scriplets-and-triggers
+	Epoch       *int32 `json:"epoch"`
```
jlsherrill, line 23: any reason to not just use 'int'?   also, why is this a pointer?

db/migrations/20220610144605_hmscontent_47.up.sql
```diff
@@ -0,0 +1,44 @@
...
+);
```
jlsherrill, line 37: I played a bit with this and started building ontop of it over here:  https://github.com/content-services/content-sources-backend/pull/38

And i realized we probably should split out rpms into a many to many association. so i envision something like:

repositories:   uuid, url
rpms:   uuid, name, version, release, arch, epoch
repository_rpms:   repository_uuid, rpm_uuid

This would allow some normalization and deduplication of rpm data

db/migrations/20220610144605_hmscontent_47.up.sql
jlsherrill: actually! lets not do this, instead lets add a 'checksum' column and make a uniqueness constraint on that

pkg/models/repository.go
```diff
@@ -0,0 +1,45 @@
...
+	// Repository URL
+	URL string `json:"url" gorm:"not null"`
+	// Last time the repo meta data was read
+	LastReadTime *time.Time `gorm:"default:null"`
+	// Last time the repo meta data failed to be read
+	LastReadError *string `gorm:"default:null"`
+	// ReferRepoConfig to Repository UUID
+	ReferRepoConfig *string `gorm:"default:null"`
+	// RepoConfig is the repository configuration
+	RepoConfig *RepositoryConfiguration `gorm:"foreignKey:UUID;references:ReferRepoConfig"`
```
rverdile, line 21: which relationship is this creating with RepositoryConfiguration, "has one" or "belongs to"?

pkg/models/repository_configuration.go
```diff
@@ -1,19 +1,19 @@
...
-	Arch      string         `json:"arch" gorm:"default:null"`
-	AccountID string         `json:"account_id" gorm:"default:null"`
-	OrgID     string         `json:"org_id" gorm:"default:null"`
+	Name         string         `json:"name" gorm:"default:null"`
+	URL          string         `json:"url" gorm:"default:null"`
+	Versions     pq.StringArray `json:"version" gorm:"type:text[];default:null"`
+	Arch         string         `json:"arch" gorm:"default:null"`
+	AccountID    string         `json:"account_id" gorm:"default:null"`
+	OrgID        string         `json:"org_id" gorm:"default:null"`
+	Repositories []Repository   `json:"packages" gorm:"foreignKey:UUID"`
```
rverdile, line 16: should "json" for Repositories be "repositories"? 

Also, I think  "foreignKey" would be "ReferRepoConfig" instead of "UUID". Following the "UserRefer" example: https://gorm.io/docs/has_many.html

avisiedo: thanks! it should be "repositories"

avisiedo: I have just realized about a wrong understanding about GORM and relationships; thanks for this @rverdile ; I have wrong in the other relationships too. I will review and update it.

db/migrations/20220610144605_hmscontent_47.up.sql
```diff
@@ -0,0 +1,44 @@
...
+-- repositories
+--
+CREATE TABLE IF NOT EXISTS repositories (
+    uuid UUID UNIQUE NOT NULL PRIMARY KEY,
+    created_at TIMESTAMP WITH TIME ZONE,
+    updated_at TIMESTAMP WITH TIME ZONE,
+    url VARCHAR(255) NOT NULL,
+    last_read_time TIMESTAMP WITH TIME ZONE DEFAULT NULL,
+    last_read_error VARCHAR(255) DEFAULT NULL,
+    refer_repo_config UUID DEFAULT NULL
```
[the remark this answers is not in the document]
avisiedo: thanks! updating this and reviewing all the other database stuffs

db/migrations/20220610144605_hmscontent_47.up.sql
avisiedo: I misread an example in gorm and it tooks me to use "refer_*" naming; indeed, by default gorm is using the "TableNamePrimaryKeyFieldName" such as "RepositoriesUUID" that is translated to the database as "repositories_uuid"; I will correct this! thanks

pkg/api/repositoryRpm.go
avisiedo: Thanks! I miss this! I can see that into librpm an int32 is used here: https://ftp.osuosl.org/pub/rpm/api/4.4.2.2/parsePreamble_8c-source.html ; in the other hand, the pointer is because initially I set the field as default to null into the database (but I have checked it out and it is not in that way into the migration, so I have to remove the pointer).  I will change `*int32` by `int32`.

pkg/models/suite_test.go
```diff
@@ -0,0 +1,127 @@
...
+const orgIdTest = "acme"
+const accountIdTest = "817342"
+
+var repoConfigTest1 = RepositoryConfiguration{
+	Base: Base{
+		// UUID generated by `uuid -v 4`
+		UUID:      "67eb30d9-9264-4726-9d90-8959e0945a55",
+		CreatedAt: time.Now(),
+		UpdatedAt: time.Now(),
+	},
```
rverdile, line 69: Base will be created when Create is called, so I don't think specifying it here is necessary. In all the places I've found that you've used these demo models, you're calling create, and these values are being overwritten anyway.

pkg/models/suite_test.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+type ModelsSuite struct {
+	suite.Suite
+	db *gorm.DB
+	tx *gorm.DB
+}
+
+func getDSNWithOptions(user string, password string, dbname string, host string, port int) string {
+	return fmt.Sprintf(
```
rverdile, line 23: It seems like these database connection functions (here and in seeds) accomplish almost the same result as the functions in the db package; the only database being connected to is the one returned by config.Get(). Is that right?

Instead of creating new functions here, perhaps the db package can be used and/or modified slightly? Maybe you could add `GetDbConnection(skipDefaultTransaction bool)`? I think that would cover the use cases here in the tests.

README.md
```diff
@@ -78,7 +78,16 @@ $ make openapi
 
 ### Configuration
 
-The default configuration file in ./configs/config.yaml.example shows all available config options.  Any of these can be overridden with an environment variable.  For example  "database.name" can be passed in via an environment variable named "DATABASE_NAME".
+The default configuration file in `./configs/config.yaml.example` shows all available config options. Any of these can be overridden with an environment variable. For example `database.name` can be passed in via an environment variable named `DATABASE_NAME`.
+
+Actually some settings are included for the workspace into the
```
swadeley, line 83:
```suggestion
Actually some settings are included for the workspace into the
```
This sentence makes no sense. Maybe you mean something like:
Some settings are included for the workspace in the same configuration file
used from the CLI (./configs/config.yaml); this lets

api/docs.go
```diff
@@ -97,6 +97,30 @@ const docTemplate = `{
                 }
             }
         },
+        "/repositories/:uuid/rpms": {
+            "get": {
+                "description": "get repositories RPMS",
```
swadeley, line 102:
```suggestion
                "description": "get repositories RPMS",
```
can you make all those RPMS into RPMs ? easier to read.

README.md
```diff
@@ -78,7 +78,16 @@ $ make openapi
...
+repository that let you use the same configuration that you
+are using from the cli (`./configs/config.yaml` file); this let
+you run and debug the tests directly from VSCode if it is your
+prefered IDE.
```
swadeley, line 87:
```suggestion
preferred IDE.
```

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-47

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT--

pkg/dao/interfaces.go
```diff
@@ -11,3 +12,13 @@ type RepositoryDao interface {
...
 }
+
+type RpmDao interface {
+	Create(orgID string, accountID string, repo *models.Repository, newRpm *models.Rpm) error
+	// TODO Implement
+	// Update(a *api.RepositoryRpm) error
+	Fetch(OrgID string, AccountID string, rpmUUID string) (*api.RepositoryRpm, error)
+	List(orgID string, accountID string, uuidRepo string, limit int, offset int) (api.RepositoryRpmCollectionResponse, int64, error)
+	// TODO Implement
+	// Delete(orgId string, accountNumber string, uuid string) error
```
jlsherrill, line 23: no need for update/delete here, if its not used in the code, don't include it

pkg/dao/repository.go
```diff
@@ -30,13 +34,32 @@ func DBErrorToApi(e error) error {
...
+	// Check arguments
+	if newRepo.URL == nil || *newRepo.URL == "" {
+		return fmt.Errorf("URL cannot be nil nor empty string")
+	}
+	if newRepo.OrgID == nil || *newRepo.OrgID == "" {
+		return fmt.Errorf("OrgID cannot be nil nor empty string")
+	}
+	if newRepo.AccountID == nil || *newRepo.AccountID == "" {
+		return fmt.Errorf("AccountID cannot be nil nor empty string")
+	}
```
jlsherrill, line 46: you can do validation in the model like this:  https://github.com/content-services/content-sources-backend/blob/main/pkg/models/repository_configuration.go#L32-L47

pkg/dao/repository.go
```diff
@@ -84,29 +107,47 @@ func (r repositoryDaoImpl) fetchRepoConfig(orgID string, uuid string) (models.Re
...
 	repoConfig, err := r.fetchRepoConfig(orgID, uuid)
 	if err != nil {
 		return err
 	}
 	ApiFieldsToModel(&repoParams, &repoConfig)
-	result := db.DB.Model(&repoConfig).Updates(repoConfig.MapForUpdate())
-	if result.Error != nil {
-		return DBErrorToApi(result.Error)
+	if repoParams.URL != nil {
```
jlsherrill, line 132: i don't /think/ this check should be necessary because if URL is nil, the repository_configuration should fail to save?

db/migrations/20220512132042_create_repository_configurations_table.up.sql
```diff
@@ -1,18 +1,97 @@
...
+-- repository_configurations
+--
 CREATE TABLE IF NOT EXISTS repository_configurations(
     uuid UUID UNIQUE NOT NULL PRIMARY KEY,
+    created_at TIMESTAMP WITH TIME ZONE NOT NULL,
+    updated_at TIMESTAMP WITH TIME ZONE NOT NULL,
+
     name VARCHAR (255) NOT NULL,
-    url VARCHAR (255) NOT NULL,
+    -- url VARCHAR (255) NOT NULL,
```
jlsherrill, line 33: I thought we would keep the url in both places and lookup the repository from the repo_config's url when needed, but looking through your pr, it looks like you moved the URL there.   I think i like this, but let me think about it some more :)

pkg/dao/repository.go
```diff
@@ -84,29 +107,47 @@ func (r repositoryDaoImpl) fetchRepoConfig(orgID string, uuid string) (models.Re
...
+	return found, nil
+}
+
 func (r repositoryDaoImpl) Update(orgID string, uuid string, repoParams api.RepositoryRequest) error {
+	var repo models.Repository
+	if repoParams.URL != nil {
+		var err error
+		repo.URL = *repoParams.URL
+		if err = r.db.Where("URL = ?", repoParams.URL).FirstOrCreate(&repo).Error; err != nil {
+			return err
```
jlsherrill, line 124: Should probably wrap this in a call to DBErrorToApi

pkg/dao/repository.go
jlsherrill: oh right you moved url to Repository, so it won't fail!  I think repo.UUID would be nil in that case, so its probably still not necessary (i could be wrong though)

pkg/dao/rpm.go
```diff
@@ -0,0 +1,171 @@
...
+func GetRpmDao(db *gorm.DB) RpmDao {
+	return rpmDaoImpl{
+		db: db,
+	}
+}
+
+func (r rpmDaoImpl) isOwnedRepository(orgID string, accountID string, repoUUID string) error {
+	var repoConfigs []models.RepositoryConfiguration
+	if err := r.db.
+		Where("org_id = ? and account_id = ? and repository_uuid = ?", orgID, accountID, repoUUID).
```
jlsherrill, line 25: i *Think* we should worry about org_id and not account id, so just query off of that

pkg/dao/rpm.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	if accountID == "" {
+		return fmt.Errorf("accountID can not be an empty string")
+	}
+	if repo == nil {
+		return fmt.Errorf("repo can not be nil")
+	}
+	if newRpm == nil {
+		return fmt.Errorf("newRpm can not be nil")
+	}
```
jlsherrill, line 46: Moving the validations to the model here would be preferred too

pkg/dao/rpm.go
```diff
@@ -0,0 +1,171 @@
...
+		},
+	}
+	if err := r.db.Table(models.TableNameRpmsRepositories).Create(&repositories_rpms).Error; err != nil {
+		return DBErrorToApi(err)
+	}
+
+	return nil
+}
+
+func (r rpmDaoImpl) List(orgID string, accountID string, uuidRepo string, limit int, offset int) (api.RepositoryRpmCollectionResponse, int64, error) {
```
jlsherrill, line 71: only use orgID for authentication.  I kinda question if the dao layer here should care.  If the api layer has lookedup the uuid of the repo, maybe thats enough?

jlsherrill: i guess i'm saying maybe the authentication should be handled at the handler layer and not this layer

pkg/models/rpm_test.go
```diff
@@ -0,0 +1,187 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func (s *ModelsSuite) TestRpmCreate() {
+	t := s.T()
+	tx := s.tx
+
+	repoConfig := repoConfigTest1.DeepCopy()
+	repo := repoTest1.DeepCopy()
+	rpm := rpmTest1.DeepCopy()
```
jlsherrill, line 15: why were these deep copy's needed?

pkg/models/suite_test.go
avisiedo: Thanks! duplication removed! I realized I didn't need the options finally just keeping the connector that represent the transaction.

pkg/models/suite_test.go
avisiedo: Thanks! Updated!

pkg/models/repository.go
avisiedo: Thanks mate! Updated! Initially I understood the relationships of the entities differently; Actually, one RepositoryConfiguration "has one" Repository; one Repository "has many" Repository; I have updated the name of the fields that reference the foreign keys to match the gorm conventions.

db/migrations/20220610144605_hmscontent_47.up.sql
avisiedo: Thanks! Updated! now it has a many-to-many relationship for repositories and rpms.

pkg/api/repositoryRpm.go
avisiedo: Updated

db/migrations/20220610144605_hmscontent_47.up.sql
avisiedo: updated!

db/migrations/20220610144605_hmscontent_47.up.sql
avisiedo: thanks! updated!

pkg/dao/repository_test.go
avisiedo: thanks!

README.md
avisiedo: Thanks! Restoring the previous state! Wrong commit for this pr. I started to update it for https://github.com/content-services/content-sources-backend/pull/40

api/docs.go
avisiedo: thanks! updated

README.md
avisiedo: thanks! restored

pkg/dao/interfaces.go
avisiedo: thanks! updated

pkg/dao/repository.go
avisiedo: thanks! I have realized I was not considering empty string here; I have added that; I have added validation at BeforeCreate to validate that no blank is tried for the URL.

About RepositoryConfiguration.RepositoryUUID, it cannot be null in the database neither into the field, and it is a foreign key to a Repository record; I have realized a mistake I did when mapping the Repository fields for updating as I was not including the URL field :s   now it is updated.

db/migrations/20220512132042_create_repository_configurations_table.up.sql
avisiedo: ups! I understood it in a different way! sorry! anyway, happy to change it; thanks

pkg/dao/repository.go
avisiedo: Yes I have :)  thanks! updated!

db/migrations/20220610144605_hmscontent_47.up.sql
avisiedo: thanks! updated

pkg/models/repository.go
avisiedo: several updates here; I have related all the models and from the related one, can be loaded the slice or related record by using `Preload` method from gorm.
- `RepositoryConfig` belongs to `Repository`.
- `Repository` has many `RepositoryConfig`.
- `Repository` has many `Rpms`.
- `Rpms` has many `Repository`.

pkg/dao/repository.go
avisiedo: thanks! updated

pkg/dao/repository.go
```diff
@@ -116,21 +157,21 @@ func ApiFieldsToModel(apiRepo *api.RepositoryRequest, repoConfig *models.Reposit
...
 	if apiRepo.DistributionArch != nil {
 		repoConfig.Arch = *apiRepo.DistributionArch
 	}
 	if apiRepo.DistributionVersions != nil {
 		repoConfig.Versions = *apiRepo.DistributionVersions
 	}
+	if apiRepo.AccountID != nil {
+		repoConfig.AccountID = *apiRepo.AccountID
+		repoConfig.OrgID = *apiRepo.OrgID
+	}
```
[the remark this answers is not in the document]
avisiedo: got it! the structure confused me; I will update the interface to pass the org_id and account_id as arguments, so the values can be injected from the headers value retrieved at the handler.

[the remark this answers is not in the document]
avisiedo: stepping back; it evokes some additional changes that I think are out of the scope of this pr; happy to do the changes in a different ticket. Only removing the lines above from the ApiFieldsToModel, and moving that to the Create method into the Repository dao implementation.

pkg/dao/rpm.go
avisiedo: sorry but I cannot move this validations to the model; they are validating nil references from the arguments but not the content of the fields.

pkg/api/repositoryRpm.go
avisiedo: updated

jlsherrill:

pkg/dao/rpm.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// Add to repositories_rpms the entry to relate
+	// the rpm with the repository it belongs to
+	var repositories_rpms []map[string]interface{} = []map[string]interface{}{
+		{
+			"repository_uuid": repo.UUID,
+			"rpm_uuid":        newRpm.UUID,
+		},
+	}
+	if err := r.db.Table(models.TableNameRpmsRepositories).Create(&repositories_rpms).Error; err != nil {
```
rverdile, line 78: why not have a separate model for the `repositories_rpms` table?

pkg/dao/rpm.go
jlsherrill: oh right!  i misread

pkg/dao/rpm.go
jlsherrill: +1 to this

db/migrations/20220512132042_create_repository_configurations_table.up.sql
```diff
@@ -1,21 +1,91 @@
...
+--
+-- rpm
+--
+
+CREATE TABLE IF NOT EXISTS rpms (
+    uuid UUID UNIQUE NOT NULL PRIMARY KEY,
+    created_at TIMESTAMP WITH TIME ZONE,
+    updated_at TIMESTAMP WITH TIME ZONE,
+
+    name VARCHAR(255) NOT NULL,
```
jlsherrill, line 55: There isn't actually a performance benefit for limiting this to 255:  https://dba.stackexchange.com/questions/20974/should-i-add-an-arbitrary-length-limit-to-varchar-columns

The default limit is 65K, we should probably leave it alone or set it to some really really high limit, like 2,000.  Since we're pulling this data in from other sources, the only reason to limit it would be to reduce malicious users that may just be trying to fill up space  (granted they could do that with just a large number of rpms i guess).  So maybe we shouldn't limit any fields

db/migrations/20220512132042_create_repository_configurations_table.up.sql
```diff
@@ -1,21 +1,91 @@
...
+    arch VARCHAR(255) NOT NULL,
+    version VARCHAR(255) NOT NULL,
+    release VARCHAR(255),
+    epoch INTEGER DEFAULT 0 NOT NULL,
+    summary VARCHAR(255) NOT NULL,
+    description TEXT NOT NULL,
```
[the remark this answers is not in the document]
avisiedo: thanks! updated

db/migrations/20220512132042_create_repository_configurations_table.up.sql
avisiedo: Thanks! My postgres skills are rusted :cry:   ; I have seen gorm map strings to `TEXT` column type; is it ok to update column types to `TEXT`? should all the database fields be changed or only Rpm table?

pkg/dao/rpm.go
avisiedo: thanks! Indeed I defined at some point the type struct, but I removed it because I didn't use it.

Checking again the code, I don't like how I have made this; I am hard-coding the name of the fields into the map[string] which make it more risky to future changes; using the model here future changes are detected in compilation time instead of runtime with the tests and it improves the maintenance.

Updated

pkg/dao/rpm.go
```diff
@@ -0,0 +1,200 @@
...
+// Before create the record, it checks that the provided
+// Repository register is related with a RepositoryConfiguration
+// that belong to the indicated organization.
+// orgID It is used to check the repository which we are trying
+// to add the Rpm record belongs to the indicated organization.
+// repo The repository record that the new rpm record will
+// belong to.
+// newRpm The Rpm record to be created into the database.
+// Return error if something goes wrong, else nil.
+func (r rpmDaoImpl) Create(orgID string, repo *models.Repository, newRpm *models.Rpm) error {
```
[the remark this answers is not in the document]
jlsherrill: also normally we'd not have Create() accept a model, it would accept an api struct, but since this doesn't make sense for Rpm (you wouldn't create them from the api), i think making it private  keeps the spirit of the separation.

pkg/dao/rpm.go
```diff
@@ -0,0 +1,200 @@
...
+			Offset: offset,
+			Limit:  limit,
+		},
+	}, totalRpms, nil
+}
+
+func (r rpmDaoImpl) RepositoryRpmListFromModelToResponse(repoRpm []models.Rpm) []api.RepositoryRpm {
+	repos := make([]api.RepositoryRpm, len(repoRpm))
+	for i := 0; i < len(repoRpm); i++ {
+		// repos[i].CopyFromModel(&repoRpm[i])
```
jlsherrill, line 130: commented code  :)

avisiedo: thanks! updated

db/migrations/20220512132042_create_repository_configurations_table.up.sql
jlsherrill: resolving this as i like the direct connection

pkg/models/repositories_rpm.go
```diff
@@ -0,0 +1,20 @@
+package models
+
+import "gorm.io/gorm"
+
+const TableNameRpmsRepositories = "repositories_rpms"
+
+type RepositoriesRpms struct {
+	RepositoryUUID string `json:"repository_uuid" gorm:"not null"`
+	RpmUUID        string `json:"rpm_uuid" gorm:"not null"`
+}
```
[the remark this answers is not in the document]
avisiedo: I have renamed the files; for the RepositoryRpm struct I had to implement the `Tabler` interface (new behavior learned). Thanks mate! :+1:

pkg/dao/rpm.go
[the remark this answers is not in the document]
avisiedo: thanks! I have deleted the no required methods from the interface and updated the tests

db/migrations/20220512132042_create_repository_configurations_table.up.sql
avisiedo: thanks! updated

jlsherrill:
3.  i changed isOwnedRepository to just return true, and i'm not getting any rpms back from the api:
{"data":[],"meta":{"limit":100,"offset":0,"count":0},"links":{"first":"/api/content_sources/v1.0/repositories/2a831a10-b667-41cb-9dab-12a1f4e6db8c/rpms?limit=100\u0026offset=0","last":"/api/content_sources/v1.0/repositories/2a831a10-b667-41cb-9dab-12a1f4e6db8c/rpms?limit=100\u0026offset=0"}}

I think the join here: https://github.com/content-services/content-sources-backend/pull/42/files#diff-6585c6f15d60726520ed4e0be761b0fe6bb1b269792bc972f7dc551aa268009cR61
is joining repository_configurations with repository_rpms, when it needs to join:

repository_configurations -> repositories -> repository_rpms -> rpms

avisiedo:
thanks! I have fixed the `uuid` that is used to recover the rpm list.

I am having a look to the code that retrieve the Repository.url

avisiedo:
thanks! I have fixed the empty `url`

pkg/dao/repository.go
```diff
@@ -32,13 +34,31 @@ func DBErrorToApi(e error) error {
 }
 
 func (r repositoryDaoImpl) Create(newRepo api.RepositoryRequest) error {
-	newRepoConfig := models.RepositoryConfiguration{
-		AccountID: *newRepo.AccountID,
-		OrgID:     *newRepo.OrgID,
+	url := ""
+	if newRepo.URL != nil {
+		url = *newRepo.URL
+	}
```
rverdile, line 40: `api.createRepository` calls `FillDefaults`, which already checks for `nil` like this and replaces it with `""`

rverdile: was the redundant error check intentional?

rverdile: actually I think it makes sense to error check in both places since these are separate layers

jlsherrill:
All 3 items are looking much better.  I do notice that url is still empty when fetching the repository (but not listing):

```
$ curl  -H "`./scripts/header.sh 1 1`"  -X GET http://localhost:8000/api/content_sources/v1.0/repositories/2a831a10-b667-41cb-9dab-12a1f4e6db8c   | python -m json.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   150  100   150    0     0  55207      0 --:--:-- --:--:-- --:--:-- 75000
{
    "uuid": "2a831a10-b667-41cb-9dab-12a1f4e6db8c",
    "name": "Foo",
    "url": "",
    "distribution_versions": [],
    "distribution_arch": "",
    "account_id": "1",
    "org_id": "1"
}

```

jlsherrill:
I think thats the last thing and then this is good to go

jlsherrill (approved):
ACK! 🔥

jlsherrill:
https://issues.redhat.com/browse/14649800

PR #44, opened 2022-06-23
CONTENT-107: Allow arch/version to accept comma separated values

Andrewgdewar:
Screen Shot 2022-06-23 at 2 19 36 PM

### DIFF
--- a/pkg/dao/repository.go
+++ b/pkg/dao/repository.go
@@ -1,6 +1,8 @@
 package dao
 
 import (
+	"strings"
+
 	"github.com/content-services/content-sources-backend/pkg/api"
 	"github.com/content-services/content-sources-backend/pkg/db"
 	"github.com/content-services/content-sources-backend/pkg/models"
@@ -68,11 +70,16 @@ func (r repositoryDaoImpl) List(
 	}
 
 	if filterData.Arch != "" {
-		filteredDB = filteredDB.Where("arch = ?", filterData.Arch)
+		arches := strings.Split(filterData.Arch, ",")
+		filteredDB = filteredDB.Where("arch IN ?", arches)
 	}
 
 	if filterData.Version != "" {
-		filteredDB = filteredDB.Where("? = any (versions)", filterData.Version)
+		versions := strings.Split(filterData.Version, ",")
+		filteredDB = filteredDB.Where("? = any (versions)", versions[0])
+		for i := 1; i < len(versions); i++ {
+			filteredDB.Or("? = any (versions)", versions[i])
+		}
 	}
 
 	filteredDB.Find(&repoConfigs).Count(&totalRepos)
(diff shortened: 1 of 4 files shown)


pkg/dao/repository.go
```diff
@@ -68,11 +71,17 @@ func (r repositoryDaoImpl) List(
...
 	}
 
 	if filterData.Version != "" {
-		filteredDB = filteredDB.Where("? = any (versions)", filterData.Version)
+		versions := strings.Split(filterData.Version, ",")
+		versionQuery := fmt.Sprintf("'%s' = any (versions)", versions[0])
+		for i := 1; i < len(versions); i++ {
+			versionQuery = versionQuery + fmt.Sprintf(" OR '%s' = any (versions)", versions[i])
+		}
+		filteredDB = filteredDB.Where(versionQuery)
```
Andrewgdewar, line 84: If you can think of a better way of comparing across two arrays, please let me know...

jlsherrill: This query works:

 select count(uuid) from repository_configurations where ARRAY['9', '8']::varchar[] @>  versions;

and might support something like:
```
filteredb.Where("ARRAY[?]::varchar[] @>  versions;", versions)
```
haven't tried i though :)

jlsherrill:
don't forget to update api docs, and add tests :)

pkg/dao/repository_test.go
```diff
@@ -380,6 +380,54 @@ func (suite *ReposSuite) TestListFilterArch() {
...
+	}
+
+	err := seeds.SeedRepositoryConfigurations(db.DB, 20, seeds.SeedOptions{OrgID: orgID})
+	assert.Nil(t, err)
+
+	response, count, err := GetRepositoryDao().List(orgID, pageData, filterData)
+
+	assert.Nil(t, err)
+	assert.Equal(t, 10, len(response.Data))
+	assert.Equal(t, int64(14), count)
```
jlsherrill, line 404: I thought there was randomness to the seed data?    I think i'd prefer if you just create 2-3 repository configs explicitly and run the query on those.  That way if seed data changes it doesn't break the test

Andrewgdewar: I removed the randomness from the data.

pkg/seeds/seeds.go
```diff
@@ -36,32 +39,35 @@ func SeedRepositoryConfigurations(db *gorm.DB, size int, options SeedOptions) er
...
 }
 
-func createVersionArray(index int) []string {
-	versionArray := []string{}
+func createVersionArray(versionArray []string, emptyArray bool) []string {
+	if len(versionArray) > 0 || emptyArray {
+		return versionArray
+	}
+	randomNum := rand.Intn(20)
 	length := 0
```
rverdile, line 60:
```suggestion
	length := rand.Intn(4)
```
Could this replace all the if statements? I think currently it is most likely to produce versionArray=["7"], but is the predictability still a factor here?

rverdile: And I could see similar changes to arch and OrgID generation to remove some of the conditional logic

jlsherrill (approved):

PR #45, opened 2022-06-24
Fixes 95: preset arches and versions

jlsherrill:
* App api /repository_parameters to return a list of versions and arches
* Validate repository_configurations are saved with a valid version and arch

### DIFF
--- a/pkg/config/value_constraints.go
+++ b/pkg/config/value_constraints.go
@@ -1,0 +1,78 @@
+package config
+
+type DistributionVersion struct {
+	Name  string `json:"name" `  //Human-readable version of the version
+	Label string `json:"label" ` //Static label of the version
+}
+type DistributionArch struct {
+	Name  string `json:"name" `  //Human-readable version of the arch
+	Label string `json:"label" ` //Static label of the arch
+}
+
+const El7 = "el7"
+const El8 = "el8"
+const El9 = "el9"
+
+var DistributionVersions = [...]DistributionVersion{
+	{
+		Name:  "el7",
+		Label: El7,
+	}, {
+		Name:  "el8",
+		Label: El8,
+	}, {
+		Name:  "el9",
+		Label: El9,
+	},
+}
+
+const X8664 = "x86_64"
+const S390x = "s390x"
+const PPC64LE = "ppc64le"
+const AARCH64 = "aarch64"
+
+var DistributionArches = [...]DistributionArch{
+	{
+		Name:  "x86_64",
+		Label: X8664,
+	}, {
+		Name:  "s390x",
+		Label: S390x,
+	}, {
+		Name:  "ppc64le",
+		Label: PPC64LE,
+	}, {
+		Name:  "aarch64",
+		Label: AARCH64,
+	},
+}
+
+// ValidDistributionVersionLabels Given a list of labels, return true
+// if every item of the list is a valid distribution version.  If at least one
+//  is not valid, returns false and the first invalid version
+func ValidDistributionVersionLabels(labels []string) (bool, string) {
+	for j := 0; j < len(labels); j++ {
+		found := false
+		for i := 0; i < len(DistributionVersions); i++ {
+			if DistributionVersions[i].Label == labels[j] {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return false, labels[j]
+		}
+	}
+	return true, ""
+}
+
+// ValidArchLabel Given a label, verifies that the label is a valid distribution
+// architecture label
+func ValidArchLabel(label string) bool {
+	for i := 0; i < len(DistributionArches); i++ {
+		if DistributionArches[i].Label == label {
+			return true
+		}
+	}
+	return false
+}
--- a/pkg/dao/suite_test.go
+++ b/pkg/dao/suite_test.go
@@ -4,9 +4,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/content-services/content-sources-backend/pkg/config"
 	"github.com/content-services/content-sources-backend/pkg/db"
 	"github.com/content-services/content-sources-backend/pkg/models"
 	"github.com/lib/pq"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
 	"gorm.io/gorm"
 )
@@ -47,7 +49,7 @@ var repoConfigTest1 = models.RepositoryConfiguration{
 	},
 	Name:           "Demo Repository Config",
 	Arch:           "x86_64",
-	Versions:       pq.StringArray{"6", "7", "8", "9"},
+	Versions:       pq.StringArray{config.El7, config.El8},
 	AccountID:      accountIdTest,
 	OrgID:          orgIdTest,
 	RepositoryUUID: repoTest1.Base.UUID,
@@ -97,9 +99,8 @@ func (s *RepositorySuite) SetupTest() {
 	s.tx = s.db.Begin()
 
 	// Remove the content for the 3 involved tables
-	s.tx.Where("1=1").Delete(models.Rpm{})
-	s.tx.Where("1=1").Delete(models.Repository{})
-	s.tx.Where("1=1").Delete(models.RepositoryConfiguration{})
+	err := models.DropAll(s.tx)
+	assert.Nil(s.T(), err)
 }
 
 func (s *RepositorySuite) TearDownTest() {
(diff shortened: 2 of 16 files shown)


jlsherrill:
TODO:
* [ ] add validation
* [ ] tests

pkg/config/value_constraints.go
```diff
@@ -0,0 +1,39 @@
+package config
+
+type DistributionVersion struct {
+	Name  string
+	Label string
+}
+type DistributionArch struct {
+	Name  string
+	Label string
+}
```
jlsherrill, line 10: You might ask, why are these structs defined in both places (api and config).  The reason is that the values defined need to be read by both the api and the model layer (for validation), so it makes sense for them to live in 'config' (you wouldn't want the 'api' package being read from the model layer.  I think it also doesn't make sense for api structs to live in the 'config' package.  
So for importing & organization purposes, i think it makes sense to duplicate these and have them be converted.

An alternative would be to basically do the same thing but in the DAO layer, but since these aren't really database related, i'm not sure that makes sense.

jlsherrill: And maybe a new packages makes more sense, open to ideas :)

jlsherrill:
Example:  
```
$ curl localhost:8000/api/content_sources/v1.0/repository_parameters/ | python -m json.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   281  100   281    0     0   115k      0 --:--:-- --:--:-- --:--:--  137k
{
    "distribution_versions": [
        {
            "name": "el7",
            "label": "el7"
        },
        {
            "name": "el8",
            "label": "el8"
        },
        {
            "name": "el9",
            "label": "el9"
        }
    ],
    "distribution_arches": [
        {
            "name": "x86_64",
            "label": "x86_64"
        },
        {
            "name": "s390x",
            "label": "s390x"
        },
        {
            "name": "ppc64le",
            "label": "ppc64le"
        },
        {
            "name": "aarch64",
            "label": "aarch64"
        }
    ]
}
```

pkg/config/value_constraints.go
rverdile: Just a thought, but what if they lived in config and were imported to api? 


```
// RepositoryParameterResponse holds data returned by a repositories API response
type RepositoryParameterResponse struct {
	config.DistributionVersions []DistributionVersion `json:"distribution_versions" ` //Versions available for repository creation
	config.DistributionArches   []DistributionArch    `json:"distribution_arches"`    //Architectures available for repository creation
}
```

jlsherrill: Not a bad idea, it would mean that we have json descriptors in the config package, but maybe that's better than having to copy the data from one to the other

pkg/config/value_constraints.go
```diff
@@ -0,0 +1,78 @@
...
+	}, {
+		Name:  "ppc64le",
+		Label: PPC64LE,
+	}, {
+		Name:  "aarch64",
+		Label: AARCH64,
+	},
+}
+
+// ValidDistributionVersionLabels Given a list of labels, return true or false
```
rverdile, line 50:
```suggestion
// ValidDistributionVersionLabels Given a list of labels, return true
```

jlsherrill: its actually part of the next statement  "return true if every item of the list is a valid distribution version."

rverdile: yeah but right now it says "return true or false if every item of the list is a valid distribution version."

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-95

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-95

pkg/config/value_constraints.go
jlsherrill: oh derp derp, will fix :0

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT--

jlsherrill:

jlsherrill:
rebased, but still have to address comment

pkg/config/value_constraints.go
jlsherrill: this is now done!

pkg/models/repository_configuration.go
```diff
@@ -44,5 +48,16 @@ func (rc *RepositoryConfiguration) BeforeCreate(tx *gorm.DB) (err error) {
 	if rc.OrgID == "" {
 		err = Error{Message: "Org ID cannot be blank.", Validation: true}
 	}
+
+	if rc.Arch != "" && !config.ValidArchLabel(rc.Arch) {
+		err = Error{Message: fmt.Sprintf("Specified Distribution Architecture %s is invalid.", rc.Arch),
```
rverdile, line 53: Small thing, but I think there should be quotes around %s, e.g. "Specified Distribution Architecture "bad arch" is invalid."

pkg/models/repository_configuration.go
```diff
@@ -44,5 +48,16 @@ func (rc *RepositoryConfiguration) BeforeCreate(tx *gorm.DB) (err error) {
...
+			Validation: true}
+	}
+	valid, invalidVer := config.ValidDistributionVersionLabels(rc.Versions)
+	if len(rc.Versions) > 0 && !valid {
+		err = Error{Message: fmt.Sprintf("Specified Distribution Version %s is invalid.", invalidVer),
```
rverdile, line 58: same as arch

pkg/models/repository_configuration.go
swadeley: yes, that is not a title, just two words in a sentence, so lower case for Distribution Architecture". This also need another rebase.

pkg/dao/suite_test.go
```diff
@@ -122,10 +123,10 @@ func (s *RpmSuite) SetupTest() {
...
 	s.tx = s.db.Begin()
 
-	// Remove the content for the 3 involved tables
-	s.tx.Where("1=1").Delete(models.Rpm{})
-	s.tx.Where("1=1").Delete(models.Repository{})
-	s.tx.Where("1=1").Delete(models.RepositoryConfiguration{})
+	err := models.DropAll(s.tx)
+	if err != nil {
+		s.FailNow(err.Error())
+	}
```
jlsherrill, line 129: I moved all of these delete calls to a single method for a couple reasons:
1) repeated code
2) dropping repository before repo config was actually a mistake and would error
3) errors werent' being caught and so 2) caused really weird behavior as the transaction would be aborted yet the code would still run

pkg/models/base.go
```diff
@@ -41,3 +41,19 @@ func (in *Base) DeepCopyInto(out *Base) {
...
+	result = db.Where("1=1").Delete(RepositoryConfiguration{})
+	if result.Error != nil {
+		return result.Error
+	}
+	result = db.Where("1=1").Delete(Repository{})
+	if result.Error != nil {
+		return result.Error
+	}
+	return nil
+}
```
jlsherrill, line 59: Wasn't sure exactly where to put this method, as you can't make functions public in _test files.   We could put it in seeds, but i think maybe we should move seeds into the model package in the long term.

pkg/seeds/seeds.go
```diff
@@ -56,19 +57,11 @@ func SeedRepositoryConfigurations(db *gorm.DB, size int, options SeedOptions) er
...
-	if options.Versions != nil {
-		versions = *options.Versions
-	}
 	for i := 0; i < size; i++ {
 		repoConfig := models.RepositoryConfiguration{
 			Name:           fmt.Sprintf("%s - %s - %s", RandStringBytes(2), "TestRepo", RandStringBytes(10)),
-			Versions:       versions,
-			Arch:           arch,
+			Versions:       createVersionArray(options.Versions),
+			Arch:           createArch(options.Arch),
```
jlsherrill, line 64: i didn't realize a previous pr had removed these methods of creating random data, so i added them back

jlsherrill:
rebased and updated

pkg/seeds/seeds.go
rverdile: random orgID is gone too, not sure if that's desired

jlsherrill:
> When I seeded data, I got this for versions: `{el7,el7,el8,el7}`. Should it be possible to have el7 3 times? Or just once?

I've fixed this during seeding, but we maybe should add a check for this.  i'll open a bug to add a validation

pkg/seeds/seeds.go
```diff
@@ -173,8 +167,51 @@ func SeedRpms(db *gorm.DB, repo *models.Repository, size int) error {
...
+func createVersionArray(existingVersionArray *[]string) []string {
+	if existingVersionArray != nil {
+		return *existingVersionArray
+	}
+	var versionArray []string
+	for k := rand.Intn(4); k < len(config.DistributionVersions); k++ {
+		ver := config.DistributionVersions[k].Label
+		versionArray = append(versionArray, ver)
+	}
+	return versionArray
```
rverdile, line 198: It isn't exactly an error, but if `k=3` then `versionArray` is `null`. I feel like we generally expect versions to be `[]` if it's empty because `FillDefaults` makes it `[]` on `createRepository` in the handler. Seeding with null here would cause inconsistencies with the seeded data vs the API data. Do we care about that?

jlsherrill: good catch, fixed!

jlsherrill:
Built and pushed image for cc5b7fe126b507ba9bb9cf0ec27c2da31b6f0c08

swadeley (approved):
ACK

jlsherrill:
https://issues.redhat.com/browse/14702683

PR #49, opened 2022-07-06
Fixes 116: Add create response object and openAPI spec fixes

rverdile:
Adds:
1. Create returns api.RepositoryResponse object
2. With 201 status, create will return Location header with the url to the resource. I see many sources online saying this is part of the spec, so I thought we might want it.

Fixes:
1. Fetch spec success now says it returns api.RepositoryResponse object
2. Delete spec now says it returns 204 instead of 200
3. Spec examples for distributions_versions were being formatted strangely

old
```
"example": [ 
  "['7'",
  "'8']"
 ]
```

new (I think this is preferred?)
```
"example": [ 
  "7",
  "8"
 ]
```

### DIFF
--- a/api/docs.go
+++ b/api/docs.go
@@ -69,30 +69,16 @@ const docTemplate = `{
                 ],
                 "responses": {
                     "201": {
-                        "description": ""
-                    }
-                }
-            }
-        },
-        "/repositories/:uuid": {
-            "delete": {
-                "tags": [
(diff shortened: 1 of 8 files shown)


jlsherrill:
reminder, hold off on merging until qe reviews :)

jlsherrill:
https://issues.redhat.com/browse/14733471

PR #51, opened 2022-07-13
Fixes 103: add search for packages

avisiedo:
- Add `Public` column with `false` as default value.
- Add handler for the `/rpms/names` route.
- Implement the logic to list the suggested packages given a query and a list of urls.
- Add unit tests.

### DIFF
--- a/pkg/api/repository_rpm.go
+++ b/pkg/api/repository_rpm.go
@@ -29,6 +29,16 @@ type RepositoryRpmCollectionResponse struct {
 	Links Links `json:"links"`
 }
 
+type SearchRpmRequest struct {
+	URLs  []string `json:"urls"`
+	Query string   `json:"query"`
+}
+
+type SearchRpmResponse struct {
+	// List of suggested package names
+	PackageNames []string `json:"package_names"`
+}
+
 // SetMetadata Map metadata to the collection.
 // meta Metadata about the request.
 // links Links to other pages of results.
--- a/pkg/api/repository_rpm.go
+++ b/pkg/api/repository_rpm.go
@@ -35,8 +35,8 @@ type SearchRpmRequest struct {
 }
 
 type SearchRpmResponse struct {
-	PackageName string `json:"package_name" gorm:"column:package_name"`
-	Summary     string `json:"summary" gorm:"column:summary"`
+	PackageName string `json:"package_name"`
+	Summary     string `json:"summary"`
 }
 
 // SetMetadata Map metadata to the collection.
--- a/pkg/api/repository_rpm.go
+++ b/pkg/api/repository_rpm.go
@@ -35,8 +35,8 @@ type SearchRpmRequest struct {
 }
 
 type SearchRpmResponse struct {
-	// List of suggested package names
-	PackageNames []string `json:"package_names"`
+	PackageName string `json:"package_name" gorm:"column:package_name"`
+	Summary     string `json:"summary" gorm:"column:summary"`
 }
 
 // SetMetadata Map metadata to the collection.
--- a/pkg/api/repository_rpm.go
+++ b/pkg/api/repository_rpm.go
@@ -30,8 +30,8 @@ type RepositoryRpmCollectionResponse struct {
 }
 
 type SearchRpmRequest struct {
-	URLs  []string `json:"urls"`
-	Query string   `json:"query"`
+	URLs   []string `json:"urls"`
+	Search string   `json:"search"`
 }
 
 type SearchRpmResponse struct {
--- a/pkg/dao/rpm.go
+++ b/pkg/dao/rpm.go
@@ -116,3 +116,36 @@ func (r rpmDaoImpl) modelToApiFields(in *models.Rpm, out *api.RepositoryRpm) {
 	out.Summary = in.Summary
 	out.Checksum = in.Checksum
 }
+
+func (r rpmDaoImpl) Search(orgID string, request api.SearchRpmRequest, limit int) (*api.SearchRpmResponse, error) {
+	// Retrieve the repository id list
+	if orgID == "" {
+		return nil, fmt.Errorf("orgID can not be an empty string")
+	}
+
+	// Prepare the filter
+	db := r.db.
+		Table("repository_configurations").
+		Select("rpms.name").
+		Distinct("rpms.name").
+		Joins("inner join repositories on repository_configurations.repository_uuid = repositories.uuid").
+		Joins("inner join repositories_rpms on repositories.uuid = repositories_rpms.repository_uuid").
+		Joins("inner join rpms on repositories_rpms.rpm_uuid = rpms.uuid")
+	dataResponse := &api.SearchRpmResponse{}
+	for i, url := range request.URLs {
+		if i == 0 {
+			r.db.Where("repositories.url = ?", url)
+		} else {
+			r.db.Or("repositories.url = ?", url)
+		}
+	}
+	db = db.
+		Where("rpms.name LIKE ?", fmt.Sprintf("%s%%", request.Query)).
+		Scan(&dataResponse.PackageNames)
+
+	if db.Error != nil {
+		return nil, db.Error
+	}
+
+	return dataResponse, nil
+}
--- a/pkg/dao/rpm.go
+++ b/pkg/dao/rpm.go
@@ -150,7 +150,7 @@ func (r rpmDaoImpl) Search(orgID string, request api.SearchRpmRequest, limit int
 		Joins("inner join repository_configurations on repository_configurations.repository_uuid = repositories.uuid").
 		Where("repository_configurations.org_id = ?", orgID).
 		Where("repositories.public").
-		Where("rpms.name LIKE ?", fmt.Sprintf("%s%%", request.Query)).
+		Where("rpms.name LIKE ?", fmt.Sprintf("%s%%", request.Search)).
 		Where("repositories.url in ?", request.URLs).
 		Order("rpms.name ASC").
 		Order("rpms.epoch DESC").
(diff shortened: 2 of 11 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-103

avisiedo:
Rebase once https://github.com/content-services/content-sources-backend/pull/38 has been merged as it uses fields created at there

jlsherrill:
https://issues.redhat.com/browse/14709253

pkg/dao/rpm.go
```diff
@@ -116,3 +116,50 @@ func (r rpmDaoImpl) modelToApiFields(in *models.Rpm, out *api.RepositoryRpm) {
...
+
+	// https://github.com/go-gorm/gorm/issues/5318
+	dataResponse := []api.SearchRpmResponse{}
+	db := r.db.
+		Select("DISTINCT ON(rpms.name) rpms.name as package_name", "rpms.summary").
+		Table(models.TableNameRpm).
+		Joins("inner join repositories_rpms on repositories_rpms.rpm_uuid = rpms.uuid").
+		Joins("inner join repositories on repositories.uuid = repositories_rpms.repository_uuid").
+		Joins("inner join repository_configurations on repository_configurations.repository_uuid = repositories.uuid").
+		Where("repository_configurations.org_id = ?", orgID).
```
jlsherrill, line 151: This requires there to be a repository configuration for each url, but what we actually want is: 
there's a repository configuration for the URL OR the repository is marked as public.

So  I think you'd want a left outer join repository configurations, and then say  "where repository_configurations.org_id = ID OR repositories.public is True".

pkg/dao/rpm.go
```diff
@@ -116,3 +116,50 @@ func (r rpmDaoImpl) modelToApiFields(in *models.Rpm, out *api.RepositoryRpm) {
...
+		Where("repositories.public").
+		Where("rpms.name LIKE ?", fmt.Sprintf("%s%%", request.Search)).
+		Where("repositories.url in ?", request.URLs).
+		Order("rpms.name ASC").
+		Order("rpms.epoch DESC").
```
jlsherrill, line 156: do we need to order by epoch?

pkg/handler/repository_rpm.go
```diff
@@ -3,17 +3,58 @@ package handler
...
+// @Description  Search RPMs for a given list of repository URLs
+// @Tags         repositories,rpms
+// @Accept       json
+// @Produce      json
+// @Success      200 {object} api.SearchRpmRequest
+// @Router       /rpms/names [post]
+func (rh *RepositoryRpmHandler) searchRpmByName(c echo.Context) error {
+	_, orgId, err := getAccountIdOrgId(c)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, err.Error())
```
jlsherrill, line 44: shouldn't this be bad request instead of ISE?

pkg/dao/rpm.go
```diff
@@ -116,3 +116,50 @@ func (r rpmDaoImpl) modelToApiFields(in *models.Rpm, out *api.RepositoryRpm) {
...
 }
+
+func (r rpmDaoImpl) Search(orgID string, request api.SearchRpmRequest, limit int) ([]api.SearchRpmResponse, error) {
+	// Retrieve the repository id list
+	if orgID == "" {
+		return nil, fmt.Errorf("orgID can not be an empty string")
+	}
+	if len(request.URLs) == 0 {
+		return nil, fmt.Errorf("request.URLs must contain at least 1 URL")
+	}
```
jlsherrill, line 127: I could see there being some confusion down the road around trialing slashes.  I.e. if we have a repository:
http://example.com/foo/

but someone searches for  http://example.com/foo

and gets no results.  Or the other way around.  Maybe we should check for both?  


We should probably also streamline the repository creation to only create with trailing slashes?  So that a user can't create http://example.com/foo/ & http://example.com/foo by mistake.     I'll file an issue for the creation/update story, but we could go ahead and handle searches for both here, thoughts?

jlsherrill:
This is looking really good!  Mostly minor comments.  I commented out the repository_configurations join and got results!  Awesome work!

```
curl -X POST  -H 'Content-Type: application/json'  http://localhost:8000/api/content_sources/v1.0/rpms/names   -d '{"urls":["https://packages.cloud.google.com/yum/repos/cloud-sdk-el8-x86_64"], "search":"goog"}'    -H "`./scripts/header.sh  1 1 `"
[{"package_name":"google-cloud-cli","summary":"Google Cloud CLI"},{"package_name":"google-cloud-cli-anthos-auth","summary":"Google Cloud CLI Anthos Auth"},{"package_name":"google-cloud-cli-app-engine-go","summary":"Google Cloud CLI App Engine Go"},{"package_name":"google-cloud-cli-app-engine-grpc","summary":"Google Cloud CLI App Engine gRPC"},{"package_name":"google-cloud-cli-app-engine-java","summary":"Google Cloud CLI App Engine Java"},{"package_name":"google-cloud-cli-app-engine-python","summary":"Google Cloud CLI App Engine Python"},{"package_name":"google-cloud-cli-app-engine-python-extras","summary":"Google Cloud CLI App Engine Python extra libraries"},{"package_name":"google-cloud-cli-bigtable-emulator","summary":"Google Cloud CLI Bigtable Emulator"},{"package_name":"google-cloud-cli-cbt","summary":"Google Cloud CLI Cloud Bigtable"},{"package_name":"google-cloud-cli-cloud-build-local","summary":"Google Cloud CLI Cloud Build Local Builder"},{"package_name":"google-cloud-cli-cloud-run-proxy","summary":"Google Cloud CLI Cloud Run Proxy"},{"package_name":"google-cloud-cli-config-connector","summary":"Google Cloud CLI Config Connector"},{"package_name":"google-cloud-cli-datalab","summary":"Google Cloud CLI Datalab"},{"package_name":"google-cloud-cli-datastore-emulator","summary":"Google Cloud CLI Datastore Emulator"},{"package_name":"google-cloud-cli-firestore-emulator","summary":"Google Cloud CLI Firestore Emulator"},{"package_name":"google-cloud-cli-gke-gcloud-auth-plugin","summary":"Google Cloud CLI GKE Auth Plugin"},{"package_name":"google-cloud-cli-harbourbridge","summary":"Google Cloud CLI Harbourbridge"},{"package_name":"google-cloud-cli-kpt","summary":"Google Cloud CLI kpt"},{"package_name":"google-cloud-cli-kubectl-oidc","summary":"Google Cloud CLI Kubectl OIDC"},{"package_name":"google-cloud-cli-local-extract","summary":"Google Cloud CLI Local Extract"}]

```

pkg/dao/rpm.go
jlsherrill: i recently did an OR query here and learned a new thing :) :  https://github.com/content-services/content-sources-backend/pull/52/files#diff-4613b72907d5407f2d396f8946b7744d5e95d1d8c57545a7fbcf4fecace55a82R106-R108

pkg/handler/repository_rpm.go
rverdile: we've been using badIdentity for an error here: https://github.com/content-services/content-sources-backend/blob/main/pkg/handler/repositories.go#L80-L83

pkg/handler/repository_rpm.go
```diff
@@ -25,8 +66,7 @@ func RegisterRepositoryRpmRoutes(engine *echo.Group /*, rDao *dao.RepositoryRpmD
 // @Produce      json
 // @Success      200 {object} api.RepositoryRpmCollectionResponse
 // @Router       /repositories/:uuid/rpms [get]
-//
-func listRepositoryRpms(c echo.Context) error {
+func (rh *RepositoryRpmHandler) listRepositoriesRpm(c echo.Context) error {
```
rverdile, line 69: I'm in favor of the original name here, `listRepositoryRpms`

pkg/api/repository_rpm.go
```diff
@@ -29,6 +29,16 @@ type RepositoryRpmCollectionResponse struct {
...
 }
 
+type SearchRpmRequest struct {
+	URLs   []string `json:"urls"`
+	Search string   `json:"search"`
+}
+
+type SearchRpmResponse struct {
+	PackageName string `json:"package_name" gorm:"column:package_name"`
+	Summary     string `json:"summary" gorm:"column:summary"`
```
rverdile, line 39: just curious, does specifying the column names here had functionality somewhere?

rverdile (commented):
It looks like repository_rpm is missing handler tests, even on main branch. Might be good to add that here.

pkg/dao/rpm.go
avisiedo: I would say it is not important here!! removing as I can see the field is not indexed and it could be less efficient the query. thanks!

pkg/dao/rpm.go
avisiedo: :+1: for normalizing the url into the database! I'd rather a cleaner query, so I would add a mark `// FIXME 103` and a description so we can locate it and update it in the future

pkg/api/repository_rpm.go
avisiedo: Honestly I don't remember the reason because I included it; my current understanding is that `gorm` will map the to the `snake_case` by convention, so it is not necessary to specify because the convention match the specified column name. Thanks for the heads up ;)  I will remove it.

swadeley:
@avisiedo Hello, rebase please.

pkg/dao/rpm.go
avisiedo: I have added a quick pre-process to the handler, so `http://example.com/foo/` will be converted to `http://example.com/foo`; I have added the above mark for future change.

pkg/api/repository_rpm.go
avisiedo: wait!!! It is not into the `pkg/models` side; it does not make sense at all the annotation; I have removed it. thanks

swadeley (approved):
ACK

PR #55, opened 2022-07-20
Fixes 49: add introspection command

jlsherrill:
This still needs a bit of work, 

* around having the ssl certificates be able to be passed in, and likely set via the config
* possibly look at introspecting all repos, not just public ones
* moar tests


jlsherrill:
Built and pushed image for 479c12510d8e2000e4ec21be27c5206cf673327f

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-49

jlsherrill:
Built and pushed image for 345225dc5963fa17b30df11682ae83b3000df242

pkg/dao/rpm.go
```diff
@@ -170,3 +173,166 @@ func (r rpmDaoImpl) Search(orgID string, request api.SearchRpmRequest, limit int
...
+func (r rpmDaoImpl) deleteUnneeded(repo models.Repository, rpm_uuids []string) error {
+	//First get uuids that are there:
+	var existing_rpm_uuids []string
+
+	r.db.Model(&models.RepositoryRpm{}).Where("repository_uuid = ?", repo.UUID).Pluck("rpm_uuid", &existing_rpm_uuids)
+
+	rpmsToDelete := difference(existing_rpm_uuids, rpm_uuids)
+
+	//result := db.DB.Where("repositories_rpms.repository_uuid = ? and repositories_rpms.rpm_uuid not in ?", repo.UUID, *rpm_uuids).Delete(&models.RepositoryRpm{})
+	result := r.db.Where("repositories_rpms.repository_uuid = ? and repositories_rpms.rpm_uuid in ?", repo.UUID, rpmsToDelete).Delete(&models.RepositoryRpm{})
```
jlsherrill, line 306: the reason it was "not in UUID_ARRAY"   is that it was passing the list of UUIDs that we want associated to the repo into the function.  So this is meant to delete the ones that aren't in that array.

For example, say a repo already has  [a,b,c]  and then we introspect again, and the list is now [b,c,d].  We'd want to delete anything that isn't in [b,c,d].

configs/config.yaml.example
```diff
@@ -7,3 +7,6 @@ database:
 logging:
   level: debug
   console: True
+certs:
+  capath: "/cdncert/ca.pem"
```
jlsherrill, line 11: i think it might make sense to just embed the ca cert into our project.  Since its not a secret, i think this makes more sense.  Also, we would need to pull in the latest ca:  https://github.com/Katello/katello/blob/master/ca/redhat-uep.pem

pkg/dao/rpm.go
```diff
@@ -170,3 +173,168 @@ func (r rpmDaoImpl) Search(orgID string, request api.SearchRpmRequest, limit int
 
 	return dataResponse, nil
 }
+
+const pagedRpmInsertsLimit = 5000
```
jlsherrill, line 177: mentioned in chat, but adding here too.  Maybe it makes sense for this to be a setting with some sensible default value.  That way we could adjust and change it more easily

jlsherrill:
A lot of these methods i think would benefit from a comment, i'm going to do that myself since i wrote a lot of them and push that up.

pkg/dao/rpm.go
jlsherrill: ignore me, i need to get my eyes checked

pkg/dao/rpm_test.go
```diff
@@ -241,14 +253,95 @@ func (s *RpmSuite) TestRpmSearch() {
...
+			for i := 0; i < limit+1; i++ {
+				randomYumPackage(&pkgs[i])
+			}
+			return pkgs
+		}
+	default:
+		{
+			return []yum.Package{}
+		}
+	}
```
jlsherrill, line 332: It might make sense to extract the contents of each case to another method, like this:

```
func (s *RpmSuite) prepareScenarioRpms(scenario int, limit int) []yum.Package {
	switch scenario {
	case scenario0:
		{
		        return makeYumPackage(0)
		}
	case scenario3:
		// The reason of this scenario is to make debugging easier
		{
			return makeYumPackage(3)
		}
	case scenarioUnderThreshold:
		{
			return makeYumPackage(limit-1)
		}
	case scenarioThreshold:
		{
			return makeYumPackage(limit)
		}
	case scenarioOverThreshold:
		{
			return makeYumPackage(limit+1)
		}
	default:
		{
			return makeYumPackage(0)
		}
	}
```

and just define a make yumYumPackage function to do the actual making.

pkg/dao/rpm_test.go
```diff
@@ -263,3 +356,109 @@ func (s *RpmSuite) TestRpmSearchError() {
...
+
+	dao := GetRpmDao(tx, map[string]interface{}{
+		OptionPagedRpmInsertsLimit: pagedRpmInsertsLimit,
+	})
+	p := s.preparePagedRpmInsert(scenario3, pagedRpmInsertsLimit)
+	records, err := dao.InsertForRepository(uuid.NewString(), p)
+
+	assert.Error(t, err)
+	assert.Equal(t, records, int64(0))
+}
```
jlsherrill, line 464: It would also be good to add a test that:

1.  inserts some set of rpms to a repo:  [a,b,c]
2.  does the same with a new set of semi-overlapping rpms:   [b,c,d]

and verify that `a` is no longer in the repo, but `d` is in the repo

pkg/external_repos/introspect_test.go
```diff
@@ -0,0 +1,132 @@
...
+	</data>
+	<data type="other">
+	<checksum type="sha256">e13969fb0f677c92d7c3268ca5ab8967e3049d9cae1ce252390807de3b8b6e2d</checksum>
+	<open-checksum type="sha256">375dabaacebf2707d626fb6c66d8a17214a1532e3d66f6d94e0a94295e075fd6</open-checksum>
+	<location href="repodata/other.xml.gz"></location>
+	<timestamp>1658448098524979</timestamp>
+	<size>1066</size>
+	<open-size>2986</open-size>
+	</data>
+</repomd>`
```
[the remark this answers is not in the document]
avisiedo: yes! embedded! I have learned that it can be loaded directly into a variable

```golang
//go:embed "myfile"
var myfile []byte
```

pkg/config/config.go
```diff
@@ -32,6 +34,19 @@ type Logging struct {
...
+	CaPath   string
+	CertPath string
+}
+
+type Options struct {
+	PagedRpmInsertsLimit int
+}
+
+const (
+	DefaultPagedRpmInsertsLimit = 200
```
jlsherrill, line 47: i feel like 200 may be a tad low, but i did a few tries at 200 & 1000 and the time difference was minimal (3m54s vs 3m49s), so i'm okay with 200

configs/config.yaml.example
```diff
@@ -7,3 +7,7 @@ database:
 logging:
   level: debug
   console: True
+certs:
+  certpath: "/cdncert/cert.pem"
+options:
+  pagedrpminsertslimit: 100
```
jlsherrill, line 13: I'd probably prefer camel case:  cert_path, paged_rpm_Inserts_limit

pkg/dao/rpm_test.go
avisiedo: good test! and I think the current behavior does not match this; so when `InsertForRepository` is called on this scenario, `[b,c]` is preserved and not inserted again, `a` is removed because it is not in `[b,c,d]` set and `d` is added as it is not part of the rpms yet; am I right?

pkg/dao/rpm_test.go
```diff
@@ -241,14 +253,94 @@ func (s *RpmSuite) TestRpmSearch() {
...
+
+func makeYumPackage(size int) []yum.Package {
+	var pkgs []yum.Package = []yum.Package{}
+
+	if size <= 0 {
+		return pkgs
+	}
+
+	pkgs = make([]yum.Package, size)
```
jlsherrill, line 296:
```suggestion
	pkgs := make([]yum.Package, size)
	if size <= 0 {
		return pkgs
	}
```

jlsherrill: i guess this wouldn't cover a negative number being passed in, but its test code

pkg/dao/rpm_test.go
```diff
@@ -263,3 +355,109 @@ func (s *RpmSuite) TestRpmSearchError() {
...
+	p := s.prepareScenarioRpms(testCase.given, pagedRpmInsertsLimit)
+	records, err := dao.InsertForRepository(s.repo.Base.UUID, p)
+
+	if testCase.expected != "" {
+		assert.Error(t, err)
+		assert.Contains(t, err.Error(), testCase.expected)
+	} else {
+		assert.NoError(t, err)
+		assert.Equal(t, int64(len(p)), records)
+	}
```
jlsherrill, line 405: It might be good to also actually query to make sure that the packages are associated to the repository, or at least the count is what you'd expect.

pkg/external_repos/introspect.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/content-services/content-sources-backend/pkg/db"
+	"github.com/content-services/content-sources-backend/pkg/models"
+	"github.com/content-services/yummy/pkg/yum"
+	"github.com/rs/zerolog/log"
+)
+
+const (
+	RhCdnHost   = "cdn.redhat.com"
+	EnvCertPath = "CERT_PATH"
+	EnvCaPath   = "CA_PATH"
```
jlsherrill, line 22: i think these two ENV variables can be removed

pkg/config/config.go
```diff
@@ -32,6 +34,19 @@ type Logging struct {
 	Console bool
 }
 
+type Certs struct {
+	CaPath   string
```
jlsherrill, line 38: CaPath isn't needed anymore right?

pkg/external_repos/introspect.go
```diff
@@ -0,0 +1,125 @@
...
+			caCert []byte
+			err    error
+		)
+		configuration := config.Get()
+
+		if configuration.Certs.CertPath == "" {
+			return http.Client{}, fmt.Errorf("Configuration for cert path not found")
+		}
+
+		cert, err = tls.LoadX509KeyPair(configuration.Certs.CertPath, configuration.Certs.CertPath)
```
jlsherrill, line 104: If the keypair is coming in via an env variable (likely on our containerized deployment via a secret).    But lets push this off to a 2nd task:  https://issues.redhat.com/browse/HMSCONTENT-70

pkg/dao/rpm_test.go
jlsherrill: yes exactly!    i imagine testing that b & c are not inserted again, might be difficult.  You could check the uuid of the rpm i guess.  I don't know that you could easily check that the repositories_rpms entry wasn't deleted and re-inserted though?

pkg/dao/rpm.go
```diff
@@ -170,3 +190,173 @@ func (r rpmDaoImpl) Search(orgID string, request api.SearchRpmRequest, limit int
...
+	repo, err := r.fetchRepo(repoUuid)
+	if err != nil {
+		return rowsAffected, err
+	}
+	checksums := make([]string, len(pkgs))
+	for i := 0; i < len(pkgs); i++ {
+		checksums[i] = pkgs[i].Checksum.Value
+	}
+
+	result, existingChecksums := r.existingChecksums(checksums)
```
rverdile, line 258: I don't think `existingChecksums` needs to a be function since it's only two lines and this is the only place calling it

pkg/dao/rpm.go
```diff
@@ -170,3 +190,173 @@ func (r rpmDaoImpl) Search(orgID string, request api.SearchRpmRequest, limit int
...
+func FilteredConvert(yumPkgs []yum.Package, excludeChecksums []string) []models.Rpm {
+	var dbPkgs []models.Rpm
+	for i := 0; i < len(yumPkgs); i++ {
+		yumPkg := yumPkgs[i]
+		if !stringInSlice(yumPkg.Checksum.Value, excludeChecksums) {
+			epoch := yumPkg.Version.Epoch
+			dbPkgs = append(dbPkgs, models.Rpm{
+				Base: models.Base{
+					UUID: uuid.NewString(),
+				},
```
rverdile, line 350: UUID is generated on creation, so assigning it here is unnecessary. Or is this UUID used somewhere before creation that I'm missing?

pkg/dao/rpm.go
```diff
@@ -5,17 +5,37 @@ import (
...
+const OptionPagedRpmInsertsLimit = "pagedRpmInsertsLimit"
+
 type rpmDaoImpl struct {
-	db *gorm.DB
+	db                   *gorm.DB
+	pagedRpmInsertsLimit int
 }
 
-func GetRpmDao(db *gorm.DB) RpmDao {
+func GetRpmDao(db *gorm.DB, options map[string]interface{}) RpmDao {
```
rverdile, line 23: I think it'd be better to make the options here a struct, like `RpmDaoOptions`

pkg/dao/public_repositories.go
```diff
@@ -0,0 +1,54 @@
...
+
+type PublicRepository struct {
+	UUID string
+	URL  string
+}
+
+type PublicRepositoryDao interface {
+	FetchForUrl(url string) (error, PublicRepository)
+	List() (error, []PublicRepository)
+}
```
rverdile, line 16: I believe this should be in `interfaces.go`

pkg/dao/rpm_test.go
avisiedo: thanks! I have updated; now it launch a panic when a negative number is provided, and for a size equals to 0 the empty slice.

pkg/config/config.go
avisiedo: right! cleaned-up! thanks

pkg/dao/rpm.go
avisiedo: thanks! I will remove it

avisiedo:
I am adding more tests and checking better some scenarios

pkg/dao/rpm.go
```diff
@@ -170,3 +195,173 @@ func (r rpmDaoImpl) Search(orgID string, request api.SearchRpmRequest, limit int
...
+	}
+
+	result, existingChecksums := r.existingChecksums(checksums)
+	if result.Error != nil {
+		return rowsAffected, result.Error
+	}
+
+	dbPkgs := FilteredConvert(pkgs, existingChecksums)
+
+	rowsAffected, error := r.PagedRpmInsert(&dbPkgs)
```
rverdile, line 270: "error" is a keyword, should rename to err

rverdile (commented):
~~Not sure if I'm using this incorrectly, but the command `./release/external-repos introspect https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/` fails with `{"level":"panic","error":"record not found","time":"2022-07-27T10:26:06-04:00","message":"Failed to introspect repositories"}`~~ Figured out how is works now.

Also, I think the error messages in within these introspect functions should be returning an error message along with the error, so it's easier to find where the error is coming from. I'll comment some examples.

pkg/dao/rpm.go
```diff
@@ -170,3 +195,173 @@ func (r rpmDaoImpl) Search(orgID string, request api.SearchRpmRequest, limit int
...
+// InsertForRepository inserts a set of yum packages for a given repository
+//   and removes any that are not in the list.  This will involve inserting the RPMs
+//   if not present, and adding or removing any associations to the Repository
+//   Returns a count of new RPMs added to the system (not the repo), as well as any error
+func (r rpmDaoImpl) InsertForRepository(repoUuid string, pkgs []yum.Package) (int64, error) {
+	var rowsAffected int64
+
+	repo, err := r.fetchRepo(repoUuid)
+	if err != nil {
+		return rowsAffected, err
```
rverdile, line 256:
```suggestion
		return rowsAffected, fmt.Errorf("failed to fetch repository %w: ", err)
```

If we try to do this in more places, it's easier to debug, because then the error message is something like "failed to introspect: failed to fetch repo: record not found"

mshriver:
test

avisiedo:
I have to import the repositories before run the command for a specific url. After I select a no redhat repository and I use the command with that url such as:

```sh
make repos-download repos-import
./release/external-repos introspect https://packages.cloud.google.com/yum/repos/google-compute-engine-el9-x86_64-stable
```

pkg/dao/rpm_test.go
avisiedo: no, I am checking the record counts; for the given scenario I expect some records inserted that could differ from the final rpm associated to the repository; but the numbers should match (4 new records inserted, but 5 records associated from 5 yum package records).

pkg/dao/rpm.go
```diff
@@ -170,3 +199,220 @@ func (r rpmDaoImpl) Search(orgID string, request api.SearchRpmRequest, limit int
...
+	// Remove dangling rpms
+	if err := r.db.
+		Unscoped().
+		Where("rpms.uuid in (?)", dangling_rpm_uuids).
+		Delete(&models.Rpm{}).
+		Error; err != nil {
+		return err
+	}
+
+	return nil
```
jlsherrill, line 387: I think it might make more sense to just leave dangling rpms within the db at introspection time and remove them either at the end of the process, or some other time.   Lets leave this for now and i filed:  https://issues.redhat.com/browse/HMSCONTENT-144

pkg/external_repos/introspect.go
```diff
@@ -0,0 +1,123 @@
...
+//  Returns the number of new RPMs inserted system-wide and all errors encountered
+func IntrospectAll() (int64, []error) {
+	var repos []models.Repository
+	var errors []error
+	var total int64
+	var count int64
+	var err error
+	thisdb := db.DB
+	rpmDao := dao.GetRpmDao(thisdb, nil)
+	result := thisdb.Find(&repos)
```
jlsherrill, line 70: Can we move this Find() to the repository dao layer?

jlsherrill: lets do this as a another pr, and rename some of the dao layer.  I'll open an issue

pkg/dao/public_repositories.go
```diff
@@ -0,0 +1,49 @@
...
+	repo := models.Repository{}
+	result := p.db.Where("public = true and URL = ?", url).First(&repo)
+	if result.Error != nil {
+		return result.Error, PublicRepository{}
+	}
+	return nil, PublicRepository{
+		UUID: repo.UUID,
+		URL:  repo.URL,
+	}
+}
```
jlsherrill, line 33: Ryan's issue with the introspect command was with regard to this method only working with Public repos, while introspect-all works with public and user-defined.  We should probably make this consistent.    For now lets just change this to not care about public = true.  

I'm gonna file an issue to rename these dao structs to be more clear

jlsherrill: issue:  https://issues.redhat.com/browse/HMSCONTENT-145

avisiedo: thanks! updated!

jlsherrill:
I'm good with this i think, will leave the final ack to you @rverdile !

pkg/seeds/seeds.go
```diff
@@ -221,10 +221,14 @@ func createArch(existingArch *string) string {
 
 const letterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
 
-func RandStringBytes(n int) string {
+func RandStringWithTable(n int, lookup string) string {
```
rverdile, line 224: what does "WithTable" mean here? It looks like this returns a string using letters randomly grabbed from lookup. Is "RandStringWithChars" a better fit?

pkg/dao/rpm.go
```diff
@@ -5,17 +5,46 @@ import (
...
 
 	"github.com/content-services/content-sources-backend/pkg/api"
+	"github.com/content-services/content-sources-backend/pkg/config"
 	"github.com/content-services/content-sources-backend/pkg/models"
+	"github.com/content-services/yummy/pkg/yum"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 )
 
+const OptionPagedRpmInsertsLimit = "pagedRpmInsertsLimit"
```
rverdile, line 15: I don't think this is being used anywhere. Is this still meant to be here?

rverdile (commented):
Left two small comments, but this is looking good. Commands are working for me and I see all the additional error messages were added.

pkg/seeds/seeds.go
avisiedo: I mean `RandStringWithLookupTable`, but I like more `RandStringWithChars` that is shorter and explanatory. Thanks!

rverdile (approved):
ack

PR #66, opened 2022-08-01
Fixes 122: Add validation to Repository update

rverdile:
Adds validation to updating repositories.

This could also be done using a `BeforeSave` hook that would run before both Update and Create, but keeping the hooks separate seems more future-proof to me.


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-122

swadeley (approved):
ACK

PR #69, opened 2022-08-02
Fixes 124: Search filter error on URL column

rverdile:


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-124

jlsherrill:
I noticed that on 'main' if you search with a URL you get no actual error, it just returns zero items.  I think this type of error should throw a 500. Would you mind fixing it as part of this?

rverdile:
good find. updated.

jlsherrill (approved):

PR #75, opened 2022-08-09
Fixes 125: uniqify versions before save

jlsherrill:


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-125

pkg/models/repository_configuration.go
```diff
@@ -48,12 +51,28 @@ func (rc *RepositoryConfiguration) BeforeCreate(tx *gorm.DB) (err error) {
...
+	if err := rc.UniqueVersions(tx); err != nil {
+		return err
+	}
 	if err := rc.validate(); err != nil {
 		return err
 	}
 	return nil
 }
 
+func (rc *RepositoryConfiguration) UniqueVersions(tx *gorm.DB) (err error) {
```
rverdile, line 63: nitpick, but I like the name `DedupeVersions` (or maybe `UniqifyVersions`) because deduplication is the action being performed on the versions. I feel it communicates the intention of the function more clearly.

pkg/models/repository_configuration_test.go
```diff
@@ -59,6 +60,22 @@ func (suite *ModelsSuite) TestCreateInvalidVersion() {
 	assert.True(suite.T(), strings.Contains(res.Error.Error(), "version"))
 }
 
+func (suite *ModelsSuite) TestCreateDuplicateVersion() {
```
rverdile, line 63: Should there be a test for Updating too? I notice the whole file only tests create, so it's not the precedent, but the beforeupdate hook could theoretically differ from the beforecreate hook.

pkg/models/repository_configuration.go
jlsherrill: oh yeah i like dedup.  Uniqify was the best i came up with and i'm not sure thats actually a word hah.

Although, should it be Dedup or Dedupe  ?

rverdile: I'm not sure :). I don't think either of them are an actual word. "Deduplicate" would be proper. I went with dedupe because it's spelled how it sounds.

jlsherrill: went with Dedupe

pkg/models/repository_configuration_test.go
jlsherrill: added as part of this test

rverdile (approved):
lgtm

swadeley (approved):
ACK

PR #79, opened 2022-08-15
Fixes 153: add openapi parameters for listRepositories()

jlsherrill:


api/docs.go
```diff
@@ -34,6 +34,50 @@ const docTemplate = `{
...
+                    },
+                    {
+                        "type": "integer",
+                        "description": "Limit the number of items returned",
+                        "name": "limit",
+                        "in": "query"
+                    },
+                    {
+                        "type": "string",
+                        "description": "Comma separated list of architecture to optionally filter-on (IE 'x86_64,s390x' would return Repositories with x86_64 or s390x only)",
```
mshriver, line 52: Not sure if this is used elsewhere, but `IE 'some,thing` isn't really appropriate here.

1. It needs to be an abbreviation, and shouldn't be capitalized (i.e.)
2. `e.g.` is more appropriate because its an example, and the latin translation is `for example`.

jlsherrill: updated

mshriver (approved):

rverdile (approved):
ack

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-153

swadeley (approved):
ACK

jlsherrill:
https://issues.redhat.com/browse/HMS-328

PR #81, opened 2022-08-16
Fixes 149: Check revision number before introspection

rverdile:
Adds
- A check before introspection to see if the repository revision number has changed. If not, don't introspect because it is assumed the repository hasn't changed since the last introspection. 
  - If it has not changed, it won't error, just "successfully update 0 packages". 
- A "revision" column to the "repositories" table that saves the previous revision number.

I tested this by creating a yum repo locally and then serving it with apache. Then I could change the revision number to whatever I wanted.

Depends on this yummy PR: https://github.com/content-services/yummy/pull/6

~~To use this yummy PR, check it out locally and then add `replace github.com/content-services/yummy => path/to/local/yummy`, to `go.mod`, above `require`~~

Still needs:

- [ ] better tests I think

### DIFF
--- a/pkg/dao/public_repositories.go
+++ b/pkg/dao/public_repositories.go
@@ -6,8 +6,9 @@ import (
 )
 
 type PublicRepository struct {
-	UUID string
-	URL  string
+	UUID     string
+	URL      string
+	Revision string
 }
 
 func GetPublicRepositoryDao(db *gorm.DB) PublicRepositoryDao {
@@ -27,8 +28,9 @@ func (p publicRepositoryDaoImpl) FetchForUrl(url string) (error, PublicRepositor
 		return result.Error, PublicRepository{}
 	}
 	return nil, PublicRepository{
-		UUID: repo.UUID,
-		URL:  repo.URL,
+		UUID:     repo.UUID,
+		URL:      repo.URL,
+		Revision: repo.Revision,
 	}
 }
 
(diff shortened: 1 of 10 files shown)


pkg/dao/public_repositories.go
```diff
@@ -47,3 +49,22 @@ func (p publicRepositoryDaoImpl) List() (error, []PublicRepository) {
 	}
 	return nil, publicRepos
 }
+
+func (p publicRepositoryDaoImpl) UpdateRepository(pubRepo PublicRepository) error {
```
rverdile, line 53: I'm not sure about this approach, but I think public repositories part of the code is gonna get refactored away anyway.

pkg/external_repos/introspect.go
```diff
@@ -51,6 +52,18 @@ func Introspect(repo dao.PublicRepository, rpm dao.RpmDao) (int64, error) {
...
+	revision, _ := yum.GetRevision(client, repo.URL)
+	if repo.Revision != "" && revision != "" && revision == repo.Revision {
+		// If repository hasn't changed, no need to update
+		return 0, nil
+	}
+
+	repo.Revision = revision
+	if err = repoDao.UpdateRepository(repo); err != nil {
+		return 0, err
+	}
```
jlsherrill, line 65: we should save the revision only if the introspection completes successfully

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-149

pkg/external_repos/introspect.go
rverdile: Good point

pkg/dao/public_repositories.go
jlsherrill: +1 makes sense.

jlsherrill:
Code looks good, worked great, will just wait on better tests

rverdile (commented):
Added to the current introspect to check that introspecting with the given revision number results in 0 package updates.

I don't know how comprehensive this test is, but it feels a little tricky to test more since the introspect tests mock the dao layer.

jlsherrill (approved):

PR #85, opened 2022-08-24
Fixes 164: properly set versions when de-dupeing

jlsherrill:
on hooks you have to call SetColumn, you cannot just
update the attribute directly, even if it works for creation

### DIFF
--- a/pkg/dao/repository_test.go
+++ b/pkg/dao/repository_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/content-services/content-sources-backend/pkg/config"
 	"github.com/content-services/content-sources-backend/pkg/models"
 	"github.com/content-services/content-sources-backend/pkg/seeds"
+	"github.com/lib/pq"
 	"github.com/openlyinc/pointy"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


jlsherrill:
Create a repo:
```
curl -X POST http://localhost:8000/api/content_sources/v1/repositories/   -H "`scripts/header.sh 1 1`"  -d '{"url": "http://foo.com/bar", "versions": ["7"], "name": "foo"}'  -H "Content-Type: application/json"   -v
```

Then update it with duplicated versions (setting the right UUID):
```
 1137  curl -X PUT http://localhost:8000/api/content_sources/v1/repositories/29126567-ff05-4608-9bf4-4630148be298   -H "`scripts/header.sh 1 1`"  -d '{"url": "http://foo.com/bazr", "distribution_versions": ["7", "7"], "name": "foo2"}'  -H "Content-Type: application/json"   -v
```

Then fetch the repo:
```
curl -X GET http://localhost:8000/api/content_sources/v1/repositories/29126567-ff05-4608-9bf4-4630148be298  -H "`scripts/header.sh 1 1`" 
```

You shouldn't see duplicated versions, only "7"

jlsherrill:
fixes this iqe test:

tests.test_repositories.test_create_update_distro_version_is_unique (from pytest) 

```
AssertionError: assert ['8', '8'] == ['8']
  Left contains one more item: '8'
  Full diff:
  - ['8']
  + ['8', '8']
```

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-164

mshriver:
Test failures aren't related to the intent of this PR, and are already being addressed.

The intended test failure has been addressed as demonstrated by the results in ci.int.devshift.
```14:15:33 tests/test_repositories.py::test_create_update_distro_version_is_unique PASSED [ 96%]```

mshriver (approved):

PR #86, opened 2022-08-24
Fixes 143: Remove null fields from api responses

rverdile:
There were three fields I could find that were returning null:
1. The bulk create error field, when there's no error.
~2. The bulk create repository field, when there is an error.~ leaving out because it's not causing a CI failure and it's not a simple fix
3. The distribution versions field after a successful creation, if no versions were specified.

I changed these to:
1. empty string
~2. empty object~
3. empty array

### DIFF
--- a/pkg/api/repository.go
+++ b/pkg/api/repository.go
@@ -23,7 +23,7 @@ type RepositoryRequest struct {
 }
 
 type RepositoryBulkCreateResponse struct {
-	ErrorMsg   *string             `json:"error"`      // Error during creation
+	ErrorMsg   string              `json:"error"`      // Error during creation
 	Repository *RepositoryResponse `json:"repository"` // Repository object information
 }
 
--- a/pkg/models/repository_configuration.go
+++ b/pkg/models/repository_configuration.go
@@ -62,7 +62,7 @@ func (rc *RepositoryConfiguration) BeforeUpdate(tx *gorm.DB) error {
 
 func (rc *RepositoryConfiguration) DedupeVersions(tx *gorm.DB) error {
 	var versionMap = make(map[string]bool)
-	var unique pq.StringArray
+	var unique = make(pq.StringArray, 0)
 	for i := 0; i < len(rc.Versions); i++ {
 		if _, found := versionMap[rc.Versions[i]]; !found {
 			versionMap[rc.Versions[i]] = true
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -5,3 +5,6 @@ linters:
     - gofmt
     - whitespace
     - govet
+issues:
+  exclude:
+    - composite
--- a/docs/openapi/Models/api.RepositoryResponseJSON.md
+++ b/docs/openapi/Models/api.RepositoryResponseJSON.md
@@ -1,0 +1,15 @@
+# api.RepositoryResponseJSON
+## Properties
+
+| Name | Type | Description | Notes |
+|------------ | ------------- | ------------- | -------------|
+| **account\_id** | **String** | Account ID of the owner | [optional] [default to null] |
+| **distribution\_arch** | **String** | Architecture to restrict client usage to | [optional] [default to null] |
+| **distribution\_versions** | **List** | Versions to restrict client usage to | [optional] [default to null] |
+| **name** | **String** | Name of the remote yum repository | [optional] [default to null] |
+| **org\_id** | **String** | Organization ID of the owner | [optional] [default to null] |
+| **url** | **String** | URL of the remote yum repository | [optional] [default to null] |
+| **uuid** | **String** | UUID of the object | [optional] [default to null] |
+
+[[Back to Model list]](../README.md#documentation-for-models) [[Back to API list]](../README.md#documentation-for-api-endpoints) [[Back to README]](../README.md)
+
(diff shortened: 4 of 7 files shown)


pkg/api/repository.go
```diff
@@ -11,6 +11,11 @@ type RepositoryResponse struct {
 	OrgID                string   `json:"org_id" readonly:"true"`              // Organization ID of the owner
 }
 
+// RepositoryResponseJSON wraps RepositoryResponse so the response can return as an empty object instead of nil
+type RepositoryResponseJSON struct {
+	Response *RepositoryResponse
+}
```
rverdile, line 17: If we don't want to use this, I think the alternative would be a RepositoryResponse object that would return all the fields with their zero values. (mostly empty string)

pkg/models/repository_configuration.go
```diff
@@ -62,7 +62,7 @@ func (rc *RepositoryConfiguration) BeforeUpdate(tx *gorm.DB) (err error) {
 
 func (rc *RepositoryConfiguration) DedupeVersions(tx *gorm.DB) (err error) {
 	var versionMap = make(map[string]bool)
-	var unique []string
+	var unique = make(pq.StringArray, 0)
```
rverdile, line 65: this prevents distribution versions from returning null

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-143

pkg/api/repository.go
jlsherrill: Does this actually work with a python binding?  The generated json doesn't seem to indicate that the value can be null

jlsherrill: (and i tried something similar in my validation pr, as i wanted null responses if 'name' or 'url' wasn't provided, but couldn't actually get it to work with the generated python api bindings :(

rverdile: The value wouldn't be null it would be `{}`

rverdile: Not sure how to check the generated python api bindings

jlsherrill: run this to generate it in the output/ directory
```
 podman run --rm -v .:/project:z docker.io/openapitools/openapi-generator-cli generate -i /project/api/openapi.json -g python -o  /project/output/
```

then:
```
cd ./output
```

and then run python.  You can use this script to get you started (this is what i used to test the validation api, you'll need to change it to use your api):

```
import openapi_client 
from openapi_client.api import  repositories_api 
from openapi_client.model.api_repository_validation_request import ApiRepositoryValidationRequest


configuration = openapi_client.Configuration(
    host = "http://localhost:8000/api/content_sources/v1.0"
)

# Enter a context with an instance of the API client
api_client = openapi_client.ApiClient(configuration=configuration, header_name="x-rh-identity", header_value="eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IjEiLCJpbnRlcm5hbCI6eyJvcmdfaWQiOiIxIn19fQo=")
api_instance = repositories_api.RepositoriesApi(api_client)


api_repository_request = [ ApiRepositoryValidationRequest(  
    name="name_example",
    url="http://redhat.com/", ),ApiRepositoryValidationRequest()]

api_response = api_instance.validate_repository_parameters(api_repository_request)
```

jlsherrill: actually, in this case i think it works.  (reading the api code) its because the bindings only would return null for Repository  in the case of an error, and technically the object we're returning only actually is returned on success.  The error response is not mapped to an object within the schema.

jlsherrill: in which case, i don't think this part is actaully needed to be changed.  At the core problem, there's no difference in returning: 
```
{
  repository: nil}
}
```
and
```
{
  repository: {
   result: nil
  }
}
```
i don't think?

swadeley:
Hi

Rebase please.

the ci.int.devshift check for this PR only failed on `test_create_update_distro_version_is_unique`, which is fixed by https://github.com/content-services/content-sources-backend/pull/85, so thats a good.

mshriver:
Because this contains openapi spec changes, we won't see the test pass in CI - the client generation is not dynamic. @swadeley can pull the openapi spec into our plugin ahead of merging this once we're confident in the content (above discussions about scope of nil values resolved).

pkg/api/repository.go
rverdile: Oh, yeah. What I pushed is not what I meant to push.

rverdile: I removed the "Response" field which the linter doesn't like, but it makes the output `repository: {}`. I still need to try generating the python bindings and seeing if this helps. I assume we need remove `null` from all responses though.

jlsherrill:
this needs  a rebase.

pkg/api/repository.go
jlsherrill: i see what you're saying that it does produce {}, but do you know why?  It doesn't make much sense to me.  was there some docs around this you found?

I wonder if something like "BulkCreatedRepository" would be a better name than RepositoryResponseJSON?

rverdile: I saw it on a stackoverflow I can't find, but there was no explanation anyway.

My understanding is by wrapping it, then json is marshalling a struct with nothing in it, so it gives `{}`. Where as previously it would marshal a `*RepositoryResponse = nil`, and give `null`. In short, I guess it's `{nil}` vs `nil`.

You could accomplish something similar with a struct where all the fields have the `omitempty` tag. If all the fields are omitted, then `{}` is the result. But here we need RepositoryResponse to return the empty fields in some cases, so that's not an option.

rverdile: It's weird. That's why I'm not sure that we want to use it.

rverdile: I'm going to try overriding the marshal function and see where that gets me

rverdile:
failure looks unrelated: https://github.com/swaggo/swag/issues/1309

jlsherrill (approved):

PR #87, opened 2022-08-24
Fixes 159: Update validation API to support editing

Andrewgdewar:
This adds the ability for the validation API to EXCLUDE the sent UUIDs from its name/url uniqueness check, thus allowing the ability to use this API to validate against within the repository edit screen. 

This solves the problem of false positive (where the uniqueness check would fail) when validating a repo effectively against itself or others in the edit group.

An example: 

Previously this: 
```json
[
   {
        "name": "google",
        "url": "https://www.google.com"
        // "uuid": "a68feaa3-9746-4719-8e33-3ff4688fed85",
    },
    {
        "name": "apple",
        "url": "https://apple.com", 
        //  "uuid": "9d6fb177-ff10-41fc-81d4-ea12ae963042",
    }
]
```

would result in:
```json
[
    {
        "name": {
            "skipped": false,
            "valid": false,
            "error": "A repository with the name 'google' already exists."
        },
        "url": {
            "skipped": false,
            "valid": false,
            "error": "A repository with the URL 'https://www.google.com' already exists.",
            "http_code": 0,
            "metadata_present": false
        }
    },
    {
        "name": {
            "skipped": false,
            "valid": false,
            "error": "A repository with the name 'apple' already exists."
        },
        "url": {
            "skipped": false,
            "valid": false,
            "error": "A repository with the URL 'https://apple.com' already exists.",
            "http_code": 0,
            "metadata_present": false
        }
    }
]
```

Now well editing the above targeted repos, by adding in the commented UUIDs to exclude, we get the following: 
```json
[
    {
        "name": {
            "skipped": false,
            "valid": true,
            "error": ""
        },
        "url": {
            "skipped": false,
            "valid": true,
            "error": "",
            "http_code": 200,
            "metadata_present": true
        }
    },
    {
        "name": {
            "skipped": false,
            "valid": true,
            "error": ""
        },
        "url": {
            "skipped": false,
            "valid": true,
            "error": "",
            "http_code": 200,
            "metadata_present": true
        }
    }
]
```

Which is the expected outcome for validating when editing.

### DIFF
--- a/api/docs.go
+++ b/api/docs.go
@@ -681,6 +681,10 @@ const docTemplate = `{
                 "url": {
                     "description": "URL of the remote yum repository",
                     "type": "string"
+                },
+                "uuid": {
+                    "description": "If set, this is an \"Update\" validation",
+                    "type": "string"
                 }
             }
         },
--- a/api/openapi.json
+++ b/api/openapi.json
@@ -225,6 +225,10 @@
                     "url": {
                         "description": "URL of the remote yum repository",
                         "type": "string"
+                    },
+                    "uuid": {
+                        "description": "If set, this is an \"Update\" validation",
+                        "type": "string"
                     }
                 },
                 "type": "object"
--- a/docs/openapi/Models/api.RepositoryValidationRequest.md
+++ b/docs/openapi/Models/api.RepositoryValidationRequest.md
@@ -5,6 +5,7 @@
 |------------ | ------------- | ------------- | -------------|
 | **name** | **String** | Name of the remote yum repository | [optional] [default to null] |
 | **url** | **String** | URL of the remote yum repository | [optional] [default to null] |
+| **uuid** | **String** | If set, this is an \&quot;Update\&quot; validation | [optional] [default to null] |
 
 [[Back to Model list]](../README.md#documentation-for-models) [[Back to API list]](../README.md#documentation-for-api-endpoints) [[Back to README]](../README.md)
 
--- a/pkg/api/repository_parameters.go
+++ b/pkg/api/repository_parameters.go
@@ -11,6 +11,7 @@ type RepositoryParameterResponse struct {
 type RepositoryValidationRequest struct {
 	Name *string `json:"name"` // Name of the remote yum repository
 	URL  *string `json:"url"`  // URL of the remote yum repository
+	UUID *string `json:"uuid"` // If set, this is an "Update" validation
 }
 
 type RepositoryValidationResponse struct {
--- a/pkg/dao/interfaces.go
+++ b/pkg/dao/interfaces.go
@@ -13,7 +13,7 @@ type RepositoryConfigDao interface {
 	List(orgID string, paginationData api.PaginationData, filterData api.FilterData) (api.RepositoryCollectionResponse, int64, error)
 	Delete(orgID string, uuid string) error
 	SavePublicRepos(urls []string) error
-	ValidateParameters(orgId string, params api.RepositoryValidationRequest) (api.RepositoryValidationResponse, error)
+	ValidateParameters(orgId string, params api.RepositoryValidationRequest, excludedUUIDS []string) (api.RepositoryValidationResponse, error)
 }
 
 type RpmDao interface {
--- a/pkg/dao/repository_config_test.go
+++ b/pkg/dao/repository_config_test.go
@@ -887,6 +887,7 @@ func (suite *RepositoryConfigSuite) TestValidateParameters() {
 		{ // Duplicated name and url
 			Name: &repoConfig.Name,
 			URL:  &repoConfig.Repository.URL,
+			UUID: &repoConfig.UUID,
 		}, { // Not providing any name or url
 			Name: nil,
 			URL:  nil,
@@ -905,8 +906,9 @@ func (suite *RepositoryConfigSuite) TestValidateParameters() {
 		},
 	}
 
-	response, err := dao.ValidateParameters(orgId, parameters[0])
-	require.NoError(t, err)
+	response, err := dao.ValidateParameters(orgId, parameters[0], []string{})
+	assert.NoError(t, err)
+
 	assert.False(t, response.Name.Valid)
 	assert.False(t, response.Name.Skipped)
 	assert.Contains(t, response.Name.Error, "already exists.")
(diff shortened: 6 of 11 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-159

api/docs.go
```diff
@@ -678,6 +678,10 @@ const docTemplate = `{
                 "url": {
                     "description": "URL of the remote yum repository",
                     "type": "string"
+                },
+                "uuid": {
+                    "description": "If set this is an \"Update\" validation",
```
swadeley, line 683:
```suggestion
                    "description": "If set, this is an \"Update\" validation",
```

api/openapi.json
```diff
@@ -225,6 +225,10 @@
                     "url": {
                         "description": "URL of the remote yum repository",
                         "type": "string"
+                    },
+                    "uuid": {
+                        "description": "If set this is an \"Update\" validation",
```
swadeley, line 230:
```suggestion
                        "description": "If set, this is an \"Update\" validation",
```

docs/openapi/Models/api.RepositoryValidationRequest.md
```diff
@@ -5,6 +5,7 @@
 |------------ | ------------- | ------------- | -------------|
 | **name** | **String** | Name of the remote yum repository | [optional] [default to null] |
 | **url** | **String** | URL of the remote yum repository | [optional] [default to null] |
+| **uuid** | **String** | If set this is an \&quot;Update\&quot; validation | [optional] [default to null] |
```
swadeley, line 8:
```suggestion
| **uuid** | **String** | If set, this is an \&quot;Update\&quot; validation | [optional] [default to null] |
```

pkg/api/repository_parameters.go
```diff
@@ -11,6 +11,7 @@ type RepositoryParameterResponse struct {
 type RepositoryValidationRequest struct {
 	Name *string `json:"name"` // Name of the remote yum repository
 	URL  *string `json:"url"`  // URL of the remote yum repository
+	UUID *string `json:"uuid"` // If set this is an "Update" validation
```
swadeley, line 14:
```suggestion
	UUID *string `json:"uuid"` // If set, this is an "Update" validation
```

pkg/handler/repository_parameters.go
```diff
@@ -69,14 +69,21 @@ func (rph *RepositoryParameterHandler) validate(c echo.Context) error {
...
 	errors := make([]error, repoCount)
 
+	UUIDList := []string{}
+	for i := 0; i < repoCount; i++ {
+		if validationParams[i].UUID != nil {
+			UUIDList = append(UUIDList, *validationParams[i].UUID)
+		}
+	}
+
 	//Use go routine here to reduce the api call time length.
```
swadeley, line 79:
```suggestion
	// Use go routine here to reduce the api call time length.
```

pkg/handler/repository_parameters.go
```diff
@@ -69,14 +69,21 @@ func (rph *RepositoryParameterHandler) validate(c echo.Context) error {
...
 	// Each url validation can take seconds to fail in case of a timeout.
 	// So this makes the call roughly take the amount of time as a single timeout at worst case
```
swadeley, line 81:
```suggestion
	// This makes the call roughly take the amount of time as a single timeout at worst.
```
You should avoid starting a sentence with "So"; so, either make the previous sentence end with a comma and change "So" to "so" or accept the simple change proposed above.

swadeley (commented):
Some suggestions

pkg/handler/repository_parameters.go
Andrewgdewar: This wasn't me, happy to change it though.

pkg/dao/interfaces.go
```diff
@@ -13,7 +13,7 @@ type RepositoryDao interface {
 	List(orgID string, paginationData api.PaginationData, filterData api.FilterData) (api.RepositoryCollectionResponse, int64, error)
 	Delete(orgID string, uuid string) error
 	SavePublicRepos(urls []string) error
-	ValidateParameters(orgId string, params api.RepositoryValidationRequest) (api.RepositoryValidationResponse, error)
+	ValidateParameters(orgId string, params api.RepositoryValidationRequest, UUIDList []string) (api.RepositoryValidationResponse, error)
```
rverdile, line 16:
```suggestion
	ValidateParameters(orgId string, params api.RepositoryValidationRequest, uuids []string) (api.RepositoryValidationResponse, error)
```
imo list is implicit in the type and we can just call this uuids

Andrewgdewar: Ahh.. yes. agreed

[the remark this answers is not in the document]
rverdile: on second thought, they may not make sense here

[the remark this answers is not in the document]
Andrewgdewar: I don't see how, as the type is shared with the incoming request. 
I think keeping it separate makes its' purpose much more clear.

pkg/dao/repository.go
```diff
@@ -367,13 +367,17 @@ func (r repositoryDaoImpl) ValidateParameters(orgId string, params api.Repositor
...
 		response.Valid = false
 		response.Error = "Name cannot be blank"
 	} else {
 		found := models.RepositoryConfiguration{}
-		result := r.db.Where("name = ? AND ORG_ID = ?", name, orgId).Find(&found)
+		query := r.db.Where("name = ? AND ORG_ID = ?", name, orgId)
+		if len(uuids) != 0 {
+			query = query.Where("repository_configurations.uuid NOT IN ?", uuids)
+		}
+		result := query.Find(&found)
```
rverdile, line 380: I'm having trouble understanding what this PR is for just from reading the description on the issue. Could you explain what the purpose of this change is? It looks like this query is the meat of the change.

Andrewgdewar: Correct! I will add description... 💯

pkg/dao/repository.go
```diff
@@ -325,7 +325,7 @@ func isTimeout(err error) bool {
 	return false
 }
 
-func (r repositoryDaoImpl) ValidateParameters(orgId string, params api.RepositoryValidationRequest) (api.RepositoryValidationResponse, error) {
+func (r repositoryDaoImpl) ValidateParameters(orgId string, params api.RepositoryValidationRequest, uuids []string) (api.RepositoryValidationResponse, error) {
```
rverdile, line 328:
```suggestion
func (r repositoryDaoImpl) ValidateParameters(orgId string, params api.RepositoryValidationRequest, excludedUUIDs []string) (api.RepositoryValidationResponse, error) {
```

rverdile: or `uuidsToExclude`

pkg/api/repository_parameters.go
```diff
@@ -11,6 +11,7 @@ type RepositoryParameterResponse struct {
 type RepositoryValidationRequest struct {
 	Name *string `json:"name"` // Name of the remote yum repository
 	URL  *string `json:"url"`  // URL of the remote yum repository
+	UUID *string `json:"uuid"` // If set, this is an "Update" validation
```
rverdile, line 14:
```suggestion
	UUID *string `json:"uuid"` // Repositories to exclude from name and URL uniqueness validation
```

rverdile (commented):
can you add tests to `dao/repository_test.go`  and and `handler/repository_parameters_test.go`, that cover the cases with uuid?

Andrewgdewar:
> can you add tests to `dao/repository_test.go` and and `handler/repository_parameters_test.go`, that cover the cases with uuid?

I may need some help with the handler/repository_parameters_test.go, I don't really get what that is supposed to be doing?

pkg/dao/repository_config_test.go
```diff
@@ -905,43 +906,61 @@ func (suite *RepositoryConfigSuite) TestValidateParameters() {
...
 	assert.Contains(t, response.URL.Error, "blank")
 
+	mockExtRDao.Mock.On("ValidRepoMD", *parameters[0].URL).Return(200, nil)
+
+	response, err = dao.ValidateParameters(orgId, parameters[0], []string{*parameters[0].UUID})
+	assert.NoError(t, err)
+
+	assert.True(t, response.Name.Valid)
+	assert.False(t, response.Name.Skipped)
+	assert.True(t, response.URL.Valid)
```
rverdile, line 944: This is good, can you just move it to be underneath the other parameters[0] test? This way the tests are listed in order.

Andrewgdewar: ✅

rverdile (approved):
nice

PR #88, opened 2022-08-25
Fixes 115: Create authentication middleware

avisiedo:
- Add usage of go-platform-middleware.
- Add wrapper to reuse the middleware leetting skiping route configurations.
- Add tests for it.
- Update some current tests.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -181,7 +181,12 @@ func WrapMiddlewareWithSkipper(m func(http.Handler) http.Handler, skip middlewar
 }
 
 func SkipLiveness(c echo.Context) bool {
-	return c.Request().URL.Path == "/ping"
+	p := c.Request().URL.Path
+	if p == "/ping" || (strings.HasPrefix(p, "/api/content_sources/") &&
+		strings.HasSuffix(p, "/ping")) {
+		return true
+	}
+	return false
 }
 
 func ConfigureEcho() *echo.Echo {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -44,8 +44,7 @@ type Certs struct {
 
 // https://stackoverflow.com/questions/54844546/how-to-unmarshal-golang-viper-snake-case-values
 type Options struct {
-	PagedRpmInsertsLimit int  `mapstructure:"paged_rpm_inserts_limit"`
-	Local                bool `mapstructure:"local"`
+	PagedRpmInsertsLimit int `mapstructure:"paged_rpm_inserts_limit"`
 }
 
 const (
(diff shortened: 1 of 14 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-115

pkg/handler/repository_parameters_test.go
```diff
@@ -10,15 +10,17 @@ import (
...
-	"github.com/labstack/echo/v4"
 	"github.com/openlyinc/pointy"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
 )
 
 func serveRepositoryParametersRouter(req *http.Request, mockDao *MockRepositoryDao) (int, []byte, error) {
-	router := echo.New()
+	// TODO use echo.New() and add to the context the key where
+	//      the middleware store the x-rh-identity structure
```
jlsherrill, line 22: are you planning on addressing these TODOs as part of this?  If not, lets move it to an open task in jira

pkg/config/config.go
```diff
@@ -159,16 +162,41 @@ func ConfigureLogging() {
...
+				c.SetResponse(echo.NewResponse(w, c.Echo()))
+				err = next(c)
+			})).ServeHTTP(c.Response(), c.Request())
+			return
+		}
+	}
+}
+
+func SkipLiveness(c echo.Context) bool {
+	return c.Request().URL.Path == "/ping"
```
jlsherrill, line 183: we also need to support  /api/content_sources/v1.0/ping & /api/content_sources/v1/ping

maybe anything that ends in "/ping" ?

jlsherrill: also, when i run:

curl -v localhost:8000/ping
the body of the response is:

```
{"message":"pong"}
Bad Request: missing x-rh-identity header
```

pkg/handler/repository_parameters_test.go
avisiedo: I was at the beginning; but looking at again, I have updated the code to add only what it requires for the test.

pkg/config/config.go
```diff
@@ -84,6 +86,7 @@ func setDefaults(v *viper.Viper) {
 	v.SetDefault("database.name", "")
 	v.SetDefault("certs.cert_path", "")
 	v.SetDefault("options.paged_rpm_inserts_limit", DefaultPagedRpmInsertsLimit)
+	v.SetDefault("options.local", false)
```
jlsherrill, line 89: what does this setting do?

pkg/config/config.go
```diff
@@ -159,16 +162,47 @@ func ConfigureLogging() {
...
+		}
+	}
+}
+
+func SkipLiveness(c echo.Context) bool {
+	p := c.Request().URL.Path
+	if p == "/ping" || (strings.HasPrefix(p, "/api/content_sources/") &&
+		strings.HasSuffix(p, "/ping")) {
+		return true
+	}
```
jlsherrill, line 188: Actually i take this back, maybe should be a little more explicit, as when you do a curl against:
/api/content_sources/v1.0/repositories/ping  it bypasses the check (and throws a nasty panic).

So maybe we should do something like  
```
strings.HasPrefix(p, "/api/content_sources/") && string.split(p, "/").length == 4 &&  string.split(p, "/")[3] == 'ping'
```

pkg/config/config.go
avisiedo: I thought about this as a mechanism to disable the middleware with the configuration, but it is not necessary! I have to remove it! thanks!

pkg/config/config.go
avisiedo: Thanks! I have competed the skipper function with this.

I have replayed the nasty panic and I have added some defensive block to prevent the panic in our code.

jlsherrill (approved):

jlsherrill:
going ahead and merging this as i gotta change the routes

PR #89, opened 2022-08-26
Fixes 145: refactor dao with better naming

jlsherrill:
This renames repository dao to RepositoryConfig
and PublicRepository to Repository


pkg/dao/repositories.go
```diff
@@ -5,52 +5,53 @@ import (
...
+		publicRepos = append(publicRepos, Repository{
 			UUID: repos[i].UUID,
 			URL:  repos[i].URL,
 		})
 	}
 	return nil, publicRepos
 }
 
-func (p publicRepositoryDaoImpl) UpdateRepository(pubRepo PublicRepository) error {
+func (p repositoryDaoImpl) UpdateRepository(pubRepo Repository) error {
```
rverdile, line 54:
```suggestion
func (p repositoryDaoImpl) Update(repo Repository) error {
```

pkg/dao/repositories.go
```diff
@@ -5,52 +5,53 @@ import (
...
-	return nil, PublicRepository{
+	return nil, Repository{
 		UUID:     repo.UUID,
 		URL:      repo.URL,
 		Revision: repo.Revision,
 	}
 }
 
-func (p publicRepositoryDaoImpl) List() (error, []PublicRepository) {
+func (p repositoryDaoImpl) List() (error, []Repository) {
```
rverdile, line 38:
```suggestion
func (p repositoryDaoImpl) ListPublicRepositories() (error, []Repository) {
```
or maybe `ListPublic()`, since isn't listing all repositories, just the ones with the public flag.

pkg/external_repos/introspect.go
```diff
@@ -23,9 +23,9 @@ const (
 // IntrospectUrl Fetch the metadata of a url and insert RPM data
 //  Returns the number of new RPMs inserted system-wide and any error encountered
 func IntrospectUrl(url string) (int64, error) {
-	err, publicRepo := dao.GetPublicRepositoryDao(db.DB).FetchForUrl(url)
+	err, publicRepo := dao.GetRepositoryDao(db.DB).FetchForUrl(url)
```
rverdile, line 26:
```suggestion
	err, repo := dao.GetRepositoryDao(db.DB).FetchForUrl(url)
```

pkg/external_repos/introspect.go
```diff
@@ -89,13 +89,13 @@ func IntrospectAll() (int64, []error) {
...
 	rpmDao := dao.GetRpmDao(db.DB, nil)
-	repoDao := dao.GetPublicRepositoryDao(db.DB)
+	repoDao := dao.GetRepositoryDao(db.DB)
 	result := db.DB.Find(&repos)
 	if result.Error != nil {
 		return 0, []error{result.Error}
 	}
 	for i := 0; i < len(repos); i++ {
-		publicRepo := dao.PublicRepository{
+		publicRepo := dao.Repository{
```
rverdile, line 98:
```suggestion
		repo := dao.Repository{
```

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-145

pkg/dao/repositories.go
```diff
@@ -5,52 +5,53 @@ import (
 	"gorm.io/gorm"
 )
 
-type PublicRepository struct {
+// Repository internal (non-user facing) representation of a repository
+type Repository struct {
 	UUID     string
 	URL      string
 	Revision string
 }
```
rverdile, line 13: what was the reason for using a "copy" struct instead of just using the model directly?

jlsherrill: so that the model layer is separated from any layer using the dao layer.   While the handler layer isn't using thia dao, i kinda consider the 'external_repos' layer similar to the handler layer, it should touch the dao but not the models.

rverdile: oh, it's the touching the models here: https://github.com/content-services/content-sources-backend/blob/main/pkg/external_repos/introspect.go#L86

jlsherrill: ohhh, i guess List() isn't even being used then.  I'll change it to use the Dao List() call and update List() to return all repos?

rverdile: oh. ha :)

makes sense

pkg/dao/repositories.go
jlsherrill: left this alone and change it to return all Repositories, since introspectAll now uses it

pkg/dao/repositories.go
```diff
@@ -5,52 +5,54 @@ import (
...
 		Revision: repo.Revision,
 	}
 }
 
-func (p publicRepositoryDaoImpl) List() (error, []PublicRepository) {
+func (p repositoryDaoImpl) List() (error, []Repository) {
 	var repos []models.Repository
-	var publicRepos []PublicRepository
-	result := p.db.Where("public = true").Find(&repos)
+	var publicRepos []Repository
```
rverdile, line 40: I think we should try to remove the name `publicRepos` since it's not dealing with public repos anymore. Maybe `internalRepos`? Or rename `repos` to `modelRepos` and `publicRepos` to `repos`? Not sure.

pkg/dao/repositories.go
```diff
@@ -5,52 +5,54 @@ import (
...
+			UUID:     repos[i].UUID,
+			URL:      repos[i].URL,
+			Revision: repos[i].Revision,
 		})
 	}
 	return nil, publicRepos
 }
 
-func (p publicRepositoryDaoImpl) UpdateRepository(pubRepo PublicRepository) error {
+func (p repositoryDaoImpl) Update(pubRepo Repository) error {
```
rverdile, line 55: similarly here, rename `pubRepo` to some new name, since it's not a "public repo" anymore

rverdile:
pretty sure those are my last two comments

pkg/dao/repositories.go
jlsherrill: went with dbRepos and repos

PR #93, opened 2022-08-31
Fixes 169: Use swag v1.8.4

rverdile:
For now, to avoid breaking change in v1.8.5

I can't find a simple way to pin package versions with Go. go.mod doesn't work quite like a typical requirements file. It seems the two ways are to download the wanted version manually or keep the wanted version locally and replace the requirement with the local version.

### DIFF
--- a/mk/swag.mk
+++ b/mk/swag.mk
@@ -4,6 +4,7 @@
 ##
 
 SWAG=$(GO_OUTPUT)/swag
+SWAG_VERSION := v1.8.4
 
 .PHONY: install-swag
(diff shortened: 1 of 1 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-169

mshriver:
Looks pretty sane to me, and I'd agree with limiting the scope of complexity to achieve this since the intent is for it to be temporary until a release including the necessary fix is made.

avisiedo (dismissed):
nothing to add; lgtm

mshriver (approved):

PR #95, opened 2022-08-31
Fixes 109: Store and return introspection status metadata

rverdile:
Adds introspection status timestamps, status, and last error to repository. Updated on introspection and returned in list and fetch responses.

To test, create some repositories and list/fetch them. See the new fields. Then, introspect the repositories and list/fetch them again to see how the timestamps/error/status change.

### DIFF
--- a/cmd/external-repos/main.go
+++ b/cmd/external-repos/main.go
@@ -45,14 +45,13 @@ func main() {
 		if len(args) < 3 {
 			log.Fatal().Msg("Usage:  ./external_repos introspect URL")
 		}
-		count, err := external_repos.IntrospectUrl(args[2])
-		if err != nil {
-			log.Panic().Err(err).Msg("Failed to introspect repositories")
+		count, errors := external_repos.IntrospectUrl(args[2])
+		for i := 0; i < len(errors); i++ {
+			log.Panic().Err(errors[i]).Msg("Failed to introspect repository")
 		}
 		log.Debug().Msgf("Successfully Inserted %d packages", count)
 	} else if args[1] == "introspect-all" {
 		count, errors := external_repos.IntrospectAll()
-
 		for i := 0; i < len(errors); i++ {
 			log.Panic().Err(errors[i]).Msg("Failed to introspect repositories")
 		}
--- a/pkg/dao/repositories.go
+++ b/pkg/dao/repositories.go
@@ -1,15 +1,23 @@
 package dao
 
 import (
+	"time"
+
 	"github.com/content-services/content-sources-backend/pkg/models"
 	"gorm.io/gorm"
 )
 
 // Repository internal (non-user facing) representation of a repository
 type Repository struct {
-	UUID     string
-	URL      string
-	Revision string
+	UUID                         string
+	URL                          string
+	Revision                     string
+	Public                       bool
+	LastIntrospectionTime        *time.Time
+	LastIntrospectionSuccessTime *time.Time
+	LastIntrospectionUpdateTime  *time.Time
+	LastIntrospectionError       *string
+	Status                       string
 }
 
 func GetRepositoryDao(db *gorm.DB) RepositoryDao {
@@ -28,11 +36,7 @@ func (p repositoryDaoImpl) FetchForUrl(url string) (error, Repository) {
 	if result.Error != nil {
 		return result.Error, Repository{}
 	}
-	return nil, Repository{
-		UUID:     repo.UUID,
-		URL:      repo.URL,
-		Revision: repo.Revision,
-	}
+	return nil, modelToInternal(repo)
 }
 
 func (p repositoryDaoImpl) List() (error, []Repository) {
@@ -43,25 +47,20 @@ func (p repositoryDaoImpl) List() (error, []Repository) {
 		return result.Error, repos
 	}
 	for i := 0; i < len(dbRepos); i++ {
-		repos = append(repos, Repository{
-			UUID:     dbRepos[i].UUID,
-			URL:      dbRepos[i].URL,
-			Revision: dbRepos[i].Revision,
-		})
+		repos = append(repos, modelToInternal(dbRepos[i]))
 	}
 	return nil, repos
 }
 
 func (p repositoryDaoImpl) Update(repoIn Repository) error {
-	var repo models.Repository
+	var dbRepo models.Repository
 
-	result := p.db.Where("uuid = ?", repoIn.UUID).First(&repo)
+	result := p.db.Where("uuid = ?", repoIn.UUID).First(&dbRepo)
 	if result.Error != nil {
 		return result.Error
 	}
 
-	repo.URL = repoIn.URL
-	repo.Revision = repoIn.Revision
+	repo := internalToModel(repoIn)
 
 	result = p.db.Model(&repo).Updates(repo.MapForUpdate())
 	if result.Error != nil {
@@ -69,4 +68,34 @@ func (p repositoryDaoImpl) Update(repoIn Repository) error {
 	}
 
 	return nil
+}
+
+// modelToInternal returns internal Repository with fields of model
+func modelToInternal(model models.Repository) Repository {
+	var internal Repository
+	internal.UUID = model.UUID
+	internal.URL = model.URL
+	internal.Public = model.Public
+	internal.Revision = model.Revision
+	internal.LastIntrospectionError = model.LastIntrospectionError
+	internal.LastIntrospectionTime = model.LastIntrospectionTime
+	internal.LastIntrospectionUpdateTime = model.LastIntrospectionUpdateTime
+	internal.LastIntrospectionSuccessTime = model.LastIntrospectionSuccessTime
+	internal.Status = model.Status
+	return internal
+}
+
+// internalToModel returns model Repository with fields of internal
+func internalToModel(internal Repository) models.Repository {
+	var model models.Repository
+	model.UUID = internal.UUID
+	model.URL = internal.URL
+	model.Public = internal.Public
+	model.Revision = internal.Revision
+	model.LastIntrospectionError = internal.LastIntrospectionError
+	model.LastIntrospectionTime = internal.LastIntrospectionTime
+	model.LastIntrospectionUpdateTime = internal.LastIntrospectionUpdateTime
+	model.LastIntrospectionSuccessTime = internal.LastIntrospectionSuccessTime
+	model.Status = internal.Status
+	return model
 }
--- a/pkg/dao/rpm.go
+++ b/pkg/dao/rpm.go
@@ -203,13 +203,11 @@ func (r rpmDaoImpl) Search(orgID string, request api.SearchRpmRequest, limit int
 }
 
 // PagedRpmInsert insert all passed in rpms quickly, ignoring any duplicates
-// Returns count of new packages inserted, and any errors
-func (r rpmDaoImpl) PagedRpmInsert(pkgs *[]models.Rpm) (int64, error) {
-	var count int64
+func (r rpmDaoImpl) PagedRpmInsert(pkgs *[]models.Rpm) error {
 	chunk := r.pagedRpmInsertsLimit
 	var result *gorm.DB
 	if len(*pkgs) == 0 {
-		return 0, nil
+		return nil
 	}
 
 	for i := 0; i < len(*pkgs); i += chunk {
@@ -223,11 +221,10 @@ func (r rpmDaoImpl) PagedRpmInsert(pkgs *[]models.Rpm) (int64, error) {
 		}).Create((*pkgs)[i:end])
 
 		if result.Error != nil {
-			return count, result.Error
+			return result.Error
 		}
-		count += result.RowsAffected
 	}
-	return count, result.Error
+	return result.Error
 }
 
 func (r rpmDaoImpl) fetchRepo(uuid string) (models.Repository, error) {
@@ -247,7 +244,6 @@ func (r rpmDaoImpl) fetchRepo(uuid string) (models.Repository, error) {
 //   Returns a count of new RPMs added to the system (not the repo), as well as any error
 func (r rpmDaoImpl) InsertForRepository(repoUuid string, pkgs []yum.Package) (int64, error) {
 	var (
-		rowsAffected      int64
 		err               error
 		repo              models.Repository
 		existingChecksums []string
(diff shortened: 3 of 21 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-109

pkg/models/repository_rpm_test.go
```diff
@@ -9,9 +9,9 @@ func (s *ModelsSuite) TestRepositoriesRpmsValidations() {
 	const spName = "testrepositoriesrpmsvalidations"
 
 	testRepository := Repository{
-		URL:           "https://my-repository.com",
-		LastReadTime:  nil,
-		LastReadError: nil,
+		URL:                    "https://my-repository.com",
```
swadeley, line 12: How about using more [RFC 6761](https://www.rfc-editor.org/rfc/rfc6761) domain names? i.e. "example.com.", "example.net.",  "example.org.", and any names falling within those domains

cmd/external-repos/main.go
```diff
@@ -45,9 +45,11 @@ func main() {
 		if len(args) < 3 {
 			log.Fatal().Msg("Usage:  ./external_repos introspect URL")
 		}
-		count, err := external_repos.IntrospectUrl(args[2])
+		count, errors := external_repos.IntrospectUrl(args[2])
 		if err != nil {
-			log.Panic().Err(err).Msg("Failed to introspect repositories")
+			for i := 0; i < len(errors); i++ {
+				log.Panic().Err(errors[i]).Msg("Failed to introspect repository")
+			}
```
avisiedo, line 52: Unwrap this block from the `if`block and remove the `if err != nil` statement; it is checking a different variable and won't be needed as if no error `len(error)` will be 0 and loop iteration is excuted.

pkg/dao/suite_test.go
```diff
@@ -45,27 +45,34 @@ type RepositorySuite struct {
...
-	LastReadTime:  nil,
-	LastReadError: nil,
-	Public:        true,
+	URL:                          "https://www.redhat.com",
+	Public:                       true,
+	LastIntrospectionTime:        &timestamp,
+	LastIntrospectionUpdateTime:  &timestamp,
+	LastIntrospectionSuccessTime: &timestamp,
+	LastIntrospectionError:       nil,
+	Status:                       "Valid",
```
avisiedo, line 61: I suggest to replace the string by a constant; suggested ones below:

```golang
const (
	StatusValid       = "Valid"       // Repository introspected successfully
	StatusUnavailable = "Unavailable" // Repository introspedted but was unavailable
	StatusInvalid     = "Invalid"     // Repository introspected but is invalid
	StatusPending     = "Pending"     // Repository not introspected yet
)

```

pkg/dao/suite_test.go
```diff
@@ -45,27 +45,34 @@ type RepositorySuite struct {
...
-	LastReadTime:  nil,
-	LastReadError: nil,
-	Public:        false,
+	URL:                          "https://www.redhat.private.com",
+	Public:                       false,
+	LastIntrospectionTime:        &timestamp,
+	LastIntrospectionUpdateTime:  &timestamp,
+	LastIntrospectionSuccessTime: &timestamp,
+	LastIntrospectionError:       nil,
+	Status:                       "Valid",
```
avisiedo, line 75: I suggest to replace the string by a constant

pkg/external_repos/introspect.go
```diff
@@ -132,3 +147,33 @@ func httpClient(useCert bool) (http.Client, error) {
...
+func UpdateIntrospectionStatusMetadata(repo dao.Repository, repoDao dao.RepositoryDao, count int64, err error) error {
+	introspectTimeEnd := time.Now()
+	repo.LastIntrospectionTime = &introspectTimeEnd
+
+	if err != nil {
+		repo.LastIntrospectionError = pointy.String(err.Error())
+
+		switch repo.Status {
+		case "Pending":
+			repo.Status = "Invalid"
```
avisiedo, line 161: I suggest to use a string constant

pkg/external_repos/introspect.go
```diff
@@ -132,3 +147,33 @@ func httpClient(useCert bool) (http.Client, error) {
...
+		default:
+			repo.Status = "Unavailable"
```
avisiedo, line 163: I suggest to use a string constant

pkg/external_repos/introspect.go
```diff
@@ -132,3 +147,33 @@ func httpClient(useCert bool) (http.Client, error) {
...
+		}
+	} else {
+		if count != 0 {
+			repo.LastIntrospectionUpdateTime = &introspectTimeEnd
+		}
+		repo.LastIntrospectionSuccessTime = &introspectTimeEnd
+		repo.LastIntrospectionError = pointy.String("")
+		repo.Status = "Valid"
+	}
```
avisiedo, line 172: I suggest to extract this block in a private function and add a unit test to cover it; suggested block:

```golang
// I am sure a better name for this private function, only to express the idea
func introspectTransition(input dao.Repository, count int64, err error, introspectTimeEnd *time.Time) dao.Repository {
	output := input

	output.LastIntrospectionTime = introspectTimeEnd

	if err == nil {
		if count != 0 {
			output.LastIntrospectionUpdateTime = introspectTimeEnd
		}
		output.LastIntrospectionSuccessTime = introspectTimeEnd
		output.LastIntrospectionError = pointy.String("")
		output.Status = StatusValid
		return output
	}

	output.LastIntrospectionError = pointy.String(err.Error())

	switch input.Status {
	case StatusValid:
		output.Status = StatusUnavailable
		break
	case StatusInvalid:
		output.Status = StatusUnavailable
		break
	case StatusPending:
		output.Status = StatusInvalid
		break
	case StatusUnavailable:
		// Do nothing, but express the status don't change
		break
	default:
		output.Status = StatusUnavailable
		break
	}
	return output
}
```

I know the cases can be simplified, but in this specific case IMO I would leave all the cases to express the state transition.

pkg/handler/repositories_test.go
```diff
@@ -133,13 +133,18 @@ func createRepoCollection(size, limit, offset int) api.RepositoryCollectionRespo
...
+			URL:                          fmt.Sprintf("http://repo-%d.com", i),
+			DistributionVersions:         []string{config.El7},
+			DistributionArch:             config.X8664,
+			AccountID:                    mockAccountNumber,
+			OrgID:                        mockOrgId,
+			LastIntrospectionTime:        "2022-08-31 14:17:50.257623 -0400 EDT",
+			LastIntrospectionSuccessTime: "2022-08-31 14:17:50.257623 -0400 EDT",
+			LastIntrospectionUpdateTime:  "2022-08-31 14:17:50.257623 -0400 EDT",
+			LastIntrospectionError:       "",
+			Status:                       "Valid",
```
avisiedo, line 147: Same suggestion; usage of a constant for `"Valid"`.

pkg/dao/repositories.go
```diff
@@ -43,30 +46,46 @@ func (p repositoryDaoImpl) List() (error, []Repository) {
...
 	}
 	return nil, repos
 }
 
 func (p repositoryDaoImpl) Update(repoIn Repository) error {
-	var repo models.Repository
+	repo := internalToModel(repoIn)
 
-	result := p.db.Where("uuid = ?", repoIn.UUID).First(&repo)
+	result := p.db.Model(&repo).Updates(repo.MapForUpdate())
```
avisiedo, line 57: When I run `make repos-import` I see in the database the repositories are public; after run `./releases/external-repos introspect URL` I get for that repository the Public flag is updated to false into the database; I have tried to modify `MapForUpdate` to comment the line for the `Public` field, and it makes it works, for that situation, but that change will avoid the field can be updated.

avisiedo: :question: is it accurate to say that once the repository is created (Repository record into the database), the indicator Public can not be changed? if the response is yes, then the above could works

pkg/seeds/seeds.go
```diff
@@ -75,7 +75,7 @@ func SeedRepositoryConfigurations(db *gorm.DB, size int, options SeedOptions) er
 	return nil
 }
 
-func randomLastRead() (lastReadTime *time.Time, lastReadError *string) {
+func randomLastRead() (lastIntrospectionTime *time.Time, lastIntrospectionError *string) {
```
avisiedo, line 78: :question: should this return the 5 values? (LastIntrospect* and Status)

pkg/seeds/seeds.go
```diff
@@ -91,12 +91,12 @@ func SeedRepository(db *gorm.DB, size int) error {
 	// Add size randome Repository entries
 	countRecords := 0
 	for i := 0; i < size; i++ {
-		lastReadTime, lastReadError := randomLastRead()
+		lastIntrospectionTime, lastIntrospectionError := randomLastRead()
```
avisiedo, line 94: depending on the above comment, it could need more variables for the returning values.

pkg/models/repository.go
```diff
@@ -69,11 +89,14 @@ func (in *Repository) DeepCopyInto(out *Repository) {
 
 func (r *Repository) MapForUpdate() map[string]interface{} {
 	forUpdate := make(map[string]interface{})
-	forUpdate["LastReadTime"] = r.LastReadTime
-	forUpdate["LastReadError"] = r.LastReadError
 	forUpdate["URL"] = r.URL
 	forUpdate["Public"] = r.Public
```
avisiedo, line 93: This is the line I commented in above comment, and it avoided to update the column to false when the repository is introspected. But it will avoid a repository could be updated from public <> private once it was created.

avisiedo (changes requested):
only a few changes

thanks for this pr

pkg/external_repos/introspect.go
avisiedo: Sorry, I think I wrote wrong the above; I think the transition `Invalid` > `Unavailable` is `Invalid` > `Invalid`.

rverdile: this is a good idea! changed it to this and added tests

pkg/dao/suite_test.go
rverdile: good idea! updated to this. I put the constants in `/dao/repositories.go`

pkg/dao/repositories.go
rverdile: This should be fixed now. Previously, the public field was being lost in the Update function and then defaulting to `false`.

pkg/seeds/seeds.go
rverdile: Yes, I will update this next

rverdile:
Something I noticed, but I don't know if it's a something to fix here or in another issue. 

Let's say there are two repositories with distinct URLs, but the same packages. Then, on introspection, the second repository to be introspected would see no changes to the `rpms` table. In which case, there is considered to be 0 packages inserted.

In the case of this issue, the "LastIntrospectionUpdateTime" would not get updated.

jlsherrill:
@rverdile oh yes, that 'count' was really just an easy way to see that something happened on first introspection.  I wouldn't use it to determine if something changed in a repository.  In fact we can remove it completely if its causing confusion.

pkg/dao/suite_test.go
```diff
@@ -45,27 +45,34 @@ type RepositorySuite struct {
...
 var repoPrivateTest = models.Repository{
 	Base: models.Base{
 		CreatedAt: time.Now(),
 		UpdatedAt: time.Now(),
 	},
-	URL:           "https://www.redhat.private.com",
-	LastReadTime:  nil,
-	LastReadError: nil,
-	Public:        false,
+	URL:                          "https://www.redhat.private.com",
```
swadeley, line 69: This URL is not safe!
You could use, private.redhat.com, but try to use "example.com.", "example.net.", and "example.org." with subdomains . See: https://www.rfc-editor.org/rfc/rfc6761.html

pkg/seeds/seeds.go
```diff
@@ -75,14 +75,39 @@ func SeedRepositoryConfigurations(db *gorm.DB, size int, options SeedOptions) er
...
+	var (
+		state     int
+		timestamp time.Time
+	)
+	state = rand.Intn(4)
+	timestamp = time.Now()
+
+	switch state {
+	case 0:
+		status = "Pending"
```
rverdile, line 93: I couldn't use the constants defined in the dao package here, because then there'd be a circular dependency. I'm not sure where else I could put the constants. I'm open to suggestions :)

pkg/dao/rpm.go
```diff
@@ -284,25 +280,25 @@ func (r rpmDaoImpl) InsertForRepository(repoUuid string, pkgs []yum.Package) (in
...
-		Create(&associations).
-		Error; err != nil {
-		return rowsAffected, fmt.Errorf("failed to Create: %w", err)
+		Create(&associations)
+	if result.Error != nil {
+		return result.RowsAffected, fmt.Errorf("failed to Create: %w", result.Error)
 	}
 
-	return rowsAffected, err
+	return result.RowsAffected, err
```
rverdile, line 301: I changed the count being returned by InsertForRepository to the rows affected on the repositories_rpms table, so it reflects rpm changes per repository instead of rpm changes overall.

pkg/seeds/seeds.go
jlsherrill: config/ maybe?  I know its not great to put it there conceptually, but config can basically be used by everything.

Thats where i put the constants for distribution versions and arch

pkg/external_repos/introspect.go
```diff
@@ -132,3 +150,47 @@ func httpClient(useCert bool) (http.Client, error) {
...
+		output.LastIntrospectionSuccessTime = introspectTimeEnd
+		output.LastIntrospectionError = pointy.String("")
+		output.Status = dao.StatusValid
+		return output
+	}
+
+	// If introspection fails
+	switch input.Status {
+	case dao.StatusValid:
+		output.Status = dao.StatusUnavailable
```
jlsherrill, line 184: this line might be worth a comment, because i can't figure out why it makes sense :)

rverdile: if the status of the repository was valid before it failed, then the status becomes unavailable

jlsherrill (approved):
ACK from me, any further thoughts @avisiedo ?

avisiedo (approved):

swadeley (approved):
LGTM

PR #97, opened 2022-09-06
Refs 151: Add makefile rules for kafka

avisiedo:
- Add kafka-up to start local container to leverage kafka.
- Add kafka-down to stop local container.
- Add kafka-build to build the container.
- Add Dockerfile for the kafka container.
- Configurations are stored at kafka/config
- Some scripts are stored at kafka/scripts

> The configuration passed through environment variables would need some clean-up, but some could be impacting as I got error starting the containers.

> It would need refinement, but this commit start and stop the containers;another commits that I had broke this


jlsherrill:

.gitignore
```diff
@@ -27,3 +27,6 @@ configs/config.yaml
 
 # outputs from build
 Containerfile
+
+# Ignore kafka data
+kafka/data
```
jlsherrill, line 32: when i make kafka-up i get an error that this directory doesn't exist, maybe the make file should create it?

avisiedo: Thanks! I fixed this by checking and creating the directory if necessary.

avisiedo:
I fixed one more stuff related with the directory owner; For any reason the directory that is bound into the container has the mapped user (185 (container) -> 100184 (host)) (I checked the container (podman image inspect ... ) and I didn't see a `USER 185` directive); Not sure where this is comming from and this was evoking situations when starting the container. I have fixed this in the kafka Dockerfile by using `USER 0`, so the root user inside the container (o (container) > $UID (host)) is mapped to my user account out of the container, allowing to write and create directories into the bound data directory inside the container; not sure how it was working previously in the workstation; probably some residual states. now it is fixed.

I have added `kafka-clean` to stop and clean the kafka container (remove the data directory).
I have added `kafka-topics-create` so the command is launched inside the container if necessary (it uses KAFKA_TOPICS variable, we can override at variables.mk or directly from the terminal).
I have added `kafka-topics-list` so we get the list of existing topics.

jlsherrill (approved):
ACK from me.  I haven't actually tested the running server, but the containers start up for me with no issue

jlsherrill:
/retest
[test]

jlsherrill:
while the main issue could be tested, this particular Pr should not require any QE

PR #98, opened 2022-09-06
Fixes 172: fix make build and make run rules

avisiedo:
When a source code is updated or just touched, now `make build` and `make run` detect it has changed and the binaries are rerebuilt.


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-172

jlsherrill (approved):
ACK, now rebuilds on any changes!

PR #101, opened 2022-09-08
Fixes 175: rename api path to content-sources

jlsherrill:
because insights does not like content_sources
This also refactors a bit to only set the app name once in code

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ $ make run
 Hit the api:
 
 ```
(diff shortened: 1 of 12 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-175

mshriver (approved):

rverdile (approved):
lgtm. endpoints still function as normal, including skipping middleware on /ping

PR #102, opened 2022-09-08
Fixes 163: Add Any version/Any architecture options

Andrewgdewar:


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-163

rverdile (commented):
Nice. One thing, I think you should move the validation to the model layer like here: https://github.com/content-services/content-sources-backend/blob/main/pkg/models/repository_configuration.go#L97-L102

Andrewgdewar:
Made suggested changes, added a test. Woot for learning 😝

rverdile (commented):
I think it would be good to have a test that's in the models layer like here: https://github.com/content-services/content-sources-backend/blob/main/pkg/models/repository_configuration_test.go#L63-L82

If you have a test in the models layer, I'm not sure one would be needed in the dao layer, but I'm not necessarily opposed to having both.

pkg/seeds/seeds.go
```diff
@@ -199,7 +199,8 @@ func createVersionArray(existingVersionArray *[]string) []string {
 		return *existingVersionArray
 	}
 	versionArray := make([]string, 0)
-	for k := rand.Intn(4); k < len(config.DistributionVersions); k++ {
+	// The below + 1 ensures that the "any" option is ignored here
```
rverdile, line 202: would it be useful to be able to seed with "any"?

pkg/config/value_constraints.go
```diff
@@ -5,15 +5,20 @@ type DistributionVersion struct {
 	Label string `json:"label"` // Static label of the version
 }
 type DistributionArch struct {
-	Name  string `json:"name" `  // Human-readable form of the arch
-	Label string `json:"label" ` //Static label of the arch
+	Name  string `json:"name"`  // Human-readable form of the arch
+	Label string `json:"label"` // Static label of the arch
 }
 
+const ANY_VERSION = "any" //Required
```
rverdile, line 12: what does the "Required" comment indicate?

Andrewgdewar: I could probably remove that, I suppose at the time it was meant that the other types can change, this one less so.

pkg/seeds/seeds.go
Andrewgdewar: so in cases of array length 0, then you would receive an array with ["any"]

Andrewgdewar: will check this.

rverdile: Right now, the default value I see being set here is `{}`.

The "default" field tag gorm uses will replace any field with that default, if that field is a "zero" value. For a slice, that would be `nil`. However, even when I replace the output of this function with `nil`, the value I get in the database it still `{}`.

I'm not sure what's going on there, but anyway the function isn't working quite like you expect.

pkg/config/value_constraints.go
rverdile: I'm in favor of removing because it confuses me :)

pkg/seeds/seeds.go
rverdile: I should add, you wouldn't see this issue in the handler layer, because the handler calls `FillDefaults()`, which sets the default value before gorm sees it.

rverdile (commented):
If I create/update a repository with `distribution_versions:[]` in the body, it will make versions in the database an empty list.

pkg/models/repository_configuration.go
```diff
@@ -11,8 +11,8 @@ import (
 type RepositoryConfiguration struct {
 	Base
 	Name           string         `json:"name" gorm:"default:null"`
-	Versions       pq.StringArray `json:"version" gorm:"type:text[],default:null"`
-	Arch           string         `json:"arch" gorm:"default:''"`
+	Versions       pq.StringArray `json:"version" gorm:"type:text[],default:['any']"`
```
rverdile, line 14: Not sure if you saw my last comment before you pushed your latest change, but right now it's possible to create/update a repository with `[]` as the distribution_versions. If you can't get this "default" field tag to work, you may need to move filling this default to be part of validation (or as part of the beforecreate/beforeupdate hooks).

pkg/dao/repository_config_test.go
```diff
@@ -89,6 +89,14 @@ func (suite *RepositoryConfigSuite) TestRepositoryCreateAlreadyExists() {
 	}
 }
 
+func (suite *RepositoryConfigSuite) TestRepositoryCreateWithAnyAndOtherVersions() {
```
rverdile, line 92: I feel the other test covers this well enough and this one can be removed.

pkg/models/repository_configuration.go
rverdile: If you do go that route, I think the `createVersionArray` function in seeds could go back to what you had previously.

pkg/api/repository.go
```diff
@@ -27,22 +27,33 @@ type RepositoryBulkCreateResponse struct {
 	Repository *RepositoryResponse `json:"repository"` // Repository object information
 }
 
+func (r *RepositoryRequest) FillNonNilValues() {
+	defaultVersions := []string{"any"}
+	defaultArch := "any"
+	if r.DistributionVersions != nil && len(*r.DistributionVersions) == 0 {
+		r.DistributionVersions = &defaultVersions
+	}
+	if r.DistributionArch != nil && *r.DistributionArch == "" {
```
rverdile, line 36: I don't think arch is needed, right? That field tags works for gorm. Version was the only problematic one.

pkg/handler/repositories.go
```diff
@@ -198,9 +198,13 @@ func (rh *RepositoryHandler) update(c echo.Context, fillDefaults bool) error {
 	if err := c.Bind(&repoParams); err != nil {
 		return echo.NewHTTPError(http.StatusBadRequest, "error binding params: "+err.Error())
 	}
+
 	if fillDefaults {
 		repoParams.FillDefaults()
+	} else {
+		repoParams.FillNonNilValues()
```
rverdile, line 205: Hmm. So I suggested putting this at part of the beforecreate/beforeupdate hooks because then this would happen closer to the database level. 

With this way (the handler level), if I'm in the dao layer (somehow - maybe in a test) I could still create `versions = []`. For practical use cases I guess it's not important, but what do you think?

pkg/models/repository_configuration.go
rverdile, line 14: since the `"['any']"` field tag isn't working for versions, it might actually be more clear to remove the default from here. Otherwise, someone might expect that this is doing something.

pkg/api/repository.go
rverdile: In which, this function could be `FillDefaultVersions()` or something like that.

Andrewgdewar: Without this I was seeing the arch on creation being saved as "". 
I don't those tags do anything 😐

pkg/handler/repositories.go
Andrewgdewar: Will try this.

pkg/api/repository.go
rverdile: Gotcha. It turns out this default field tag only works on Create in gorm.

pkg/api/repository.go
```diff
@@ -36,20 +36,12 @@ func (r *RepositoryRequest) FillDefaults() {
...
 	}
 	if r.URL == nil {
 		r.URL = &defaultUrl
 	}
-	if r.DistributionVersions == nil {
-		r.DistributionVersions = &defaultVersions
-	}
-	if r.DistributionArch == nil {
-		r.DistributionArch = &defaultArch
-	}
```
rverdile, line 52: This is still necessary

For example:
- I have a repository, `foo` with arch `x86_64`. 
- I PUT update `foo` with just `{ "name": "new foo"}`
- I expect the arch of foo to become `any`, but it's still `x86_64`.

You should also add a test to cover this scenario

pkg/models/repository_configuration.go
```diff
@@ -73,6 +79,16 @@ func (rc *RepositoryConfiguration) DedupeVersions(tx *gorm.DB) error {
 	return nil
 }
 
+func (rc *RepositoryConfiguration) ReplaceNonNilValues(tx *gorm.DB) error {
```
rverdile, line 82:
```suggestion
func (rc *RepositoryConfiguration) ReplaceEmptyValues(tx *gorm.DB) error {
```
Empty is slightly more precise

pkg/api/repository.go
Andrewgdewar: This makes sense, but I think a bigger issue with the put update you describe is that the name/url are not required. If this is truly a PUT, you should likely require a set of variables no? What would happen to the default URL in such a situation? Should the version array switch to the default? The existing GPGkey? and validation setting (that is still to come)?

rverdile: Name/URL are essentially required. The default URL/name is `""`, and the API will return an error because validation will catch that, like here: https://github.com/content-services/content-sources-backend/blob/main/pkg/models/repository_configuration.go#L78. Perhaps they should be marked as required in the API spec, if that's what you mean?

Version array should default to `["any"]`, just like it does for create. 

gpgKey and verification default to `""` and `false`.

The idea is to reset whatever values are specified in the PUT request. The body is supposed to be a complete resource, so anything missing gets its default value.

Andrewgdewar: Got it. I'll reset these to their previous values.

rverdile (approved):
nice! reminder not to merge, this still needs qe testing

pkg/api/repository.go
```diff
@@ -36,8 +36,8 @@ func (r *RepositoryRequest) FillDefaults() {
 	//Fill in default values in case of PUT request, doesn't have to be valid, let the db validate that
```
swadeley, line 36:
```suggestion
	// Fill in default values in case of PUT request, doesn't have to be valid, let the db validate that
```

swadeley (commented):
lgtm

PR #104, opened 2022-09-12
Fixes 152: Smarter introspection job

avisiedo:
- It runs every 8 hours, starting at 0:00 hours.
- It executes for every repository with Status != 'Valid'.
- It does not execute for a repository with Status == 'Valid' hat was introspected into the last 24 hours.
- Unit tests for the above.
- Removed: Add livenessProbe to the pod job specification: it was stepped back.

### DIFF
--- a/cmd/external-repos/main.go
+++ b/cmd/external-repos/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"flag"
 	"os"
 	"sort"
 
@@ -47,21 +48,14 @@ func main() {
 		log.Debug().Msg("Successfully loaded external repositories.")
 	} else if args[1] == "introspect" {
 		if len(args) < 3 {
-			log.Panic().Err(err).Msg("Usage:  ./external_repos introspect [--force] URL")
+			log.Panic().Err(err).Msg("Usage:  ./external_repos introspect URL [--force]")
 			os.Exit(1)
 		}
-		// TODO Quick option implemented, refactor to use some library
-		var url string
-		if len(args) == 4 {
-			if args[2] != "--force" {
-				log.Info().Err(err).Msgf("Expected --force but got '%s'", args[2])
-				log.Panic().Err(err).Msg("Usage:  ./external_repos introspect [--force] URL")
-				os.Exit(1)
-			}
-			forceIntrospect = true
-			url = args[3]
-		} else {
-			url = args[2]
+		url := args[2]
+		if len(args) > 3 {
+			flagset := flag.NewFlagSet("introspect", flag.ExitOnError)
+			flagset.BoolVar(&forceIntrospect, "force", false, "Force introspection even if not needed")
+			flagset.Parse(args[3:])
 		}
 		count, errors := external_repos.IntrospectUrl(url, forceIntrospect)
 		for i := 0; i < len(errors); i++ {
@@ -69,9 +63,10 @@ func main() {
 		}
 		log.Debug().Msgf("Inserted %d packages", count)
 	} else if args[1] == "introspect-all" {
-		// TODO Quick option implemented, refactor to use some library
-		if len(args) == 3 && args[2] == "--force" {
-			forceIntrospect = true
+		if len(args) > 2 {
+			flagset := flag.NewFlagSet("introspect-all", flag.ExitOnError)
+			flagset.BoolVar(&forceIntrospect, "force", false, "Force introspection even if not needed")
+			flagset.Parse(args[2:])
 		}
 		count, errors := external_repos.IntrospectAll(forceIntrospect)
 		for i := 0; i < len(errors); i++ {
--- a/cmd/external-repos/main.go
+++ b/cmd/external-repos/main.go
@@ -55,7 +55,10 @@ func main() {
 		if len(args) > 3 {
 			flagset := flag.NewFlagSet("introspect", flag.ExitOnError)
 			flagset.BoolVar(&forceIntrospect, "force", false, "Force introspection even if not needed")
-			flagset.Parse(args[3:])
+			if err = flagset.Parse(args[3:]); err != nil {
+				log.Panic().Err(err).Msg("Usage:  ./external_repos introspect URL [--force]")
+				os.Exit(1)
+			}
 		}
 		count, errors := external_repos.IntrospectUrl(url, forceIntrospect)
 		for i := 0; i < len(errors); i++ {
@@ -66,7 +69,10 @@ func main() {
 		if len(args) > 2 {
 			flagset := flag.NewFlagSet("introspect-all", flag.ExitOnError)
 			flagset.BoolVar(&forceIntrospect, "force", false, "Force introspection even if not needed")
-			flagset.Parse(args[2:])
+			if err = flagset.Parse(args[2:]); err != nil {
+				log.Panic().Err(err).Msg("Usage:  ./external_repos introspect-all [--force]")
+				os.Exit(1)
+			}
 		}
 		count, errors := external_repos.IntrospectAll(forceIntrospect)
 		for i := 0; i < len(errors); i++ {
--- a/cmd/external-repos/main.go
+++ b/cmd/external-repos/main.go
@@ -67,7 +67,7 @@ func main() {
 		for i := 0; i < len(errors); i++ {
 			log.Panic().Err(errors[i]).Msg("Failed to introspect repository")
 		}
-		log.Debug().Msgf("Successfully Inserted %d packages", count)
+		log.Debug().Msgf("Inserted %d packages", count)
 	} else if args[1] == "introspect-all" {
 		// TODO Quick option implemented, refactor to use some library
 		if len(args) == 3 && args[2] == "--force" {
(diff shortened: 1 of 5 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-152

jlsherrill:
This is good to rebase now that ryan's pr has gone in

pkg/external_repos/introspect.go
```diff
@@ -106,6 +107,11 @@ func IntrospectAll() (int64, []error) {
 		return 0, []error{err}
 	}
 	for i := 0; i < len(repos); i++ {
+		hasToIntrospect, reason := needsIntrospect(&repos[i])
```
rverdile, line 110: I wonder if `needsIntrospect` should be called by `Introspect` so that the behavior also applies to `IntrospectUrl`. From a user perspective, I think I would expect `introspect [url]` and `introspect-all` to have the same behavior, except `introspect-all` applies that behavior to all the repositories.

What do you think?

pkg/external_repos/introspect.go
```diff
@@ -106,6 +107,11 @@ func IntrospectAll() (int64, []error) {
...
+		log.Info().Msg(reason)
+		if !hasToIntrospect {
+			continue
+		}
```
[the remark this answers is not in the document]
avisiedo: I like it, and I think it is interesting to have! something like `./release/external-repos introspect-all --force`; so if the flag exists, `hasToIntrospect` will return always true

pkg/external_repos/introspect.go
avisiedo: I like it, and I like more combined with the previous comment; it would provide consistency with both commands, in the other hand launching for an individual url could be useful to quickly troubleshoot the reason a specific repo is not introspected without launch it for all the repos.

jlsherrill: oh i see @avisiedo suggested --force above, i like that more!

cmd/external-repos/main.go
```diff
@@ -43,15 +47,33 @@ func main() {
 		log.Debug().Msg("Successfully loaded external repositories.")
 	} else if args[1] == "introspect" {
 		if len(args) < 3 {
-			log.Fatal().Msg("Usage:  ./external_repos introspect URL")
+			log.Panic().Err(err).Msg("Usage:  ./external_repos introspect [--force] URL")
+			os.Exit(1)
 		}
-		count, errors := external_repos.IntrospectUrl(args[2])
+		// TODO Quick option implemented, refactor to use some library
```
rverdile, line 53: dbmigrate handles flags with this: https://github.com/content-services/content-sources-backend/blob/main/cmd/dbmigrate/main.go#L52-L56

It's not a great library but it's simple

cmd/external-repos/main.go
```diff
@@ -43,15 +47,33 @@ func main() {
...
+			forceIntrospect = true
+			url = args[3]
+		} else {
+			url = args[2]
+		}
+		count, errors := external_repos.IntrospectUrl(url, forceIntrospect)
 		for i := 0; i < len(errors); i++ {
 			log.Panic().Err(errors[i]).Msg("Failed to introspect repository")
 		}
 		log.Debug().Msgf("Successfully Inserted %d packages", count)
```
rverdile, line 70:
```suggestion
		log.Debug().Msgf("Inserted %d packages", count)
```

This message outputs even if introspection is skipped because of the thresholds. Maybe for now we can just remove the "successfully", because that makes it seem like introspection ran.

I'm not sure this card is the place for it, but I feel like the output messaging for introspect could use some refactoring. It can be confusing, especially when there's multiple errors.

pkg/external_repos/introspect.go
```diff
@@ -124,6 +141,27 @@ func IntrospectAll() (int64, []error) {
 	return total, errors
 }
 
+func needsIntrospect(repo *dao.Repository) (bool, string) {
+	if repo == nil {
+		return false, "Cannot introspect nil Repository"
+	}
+
+	if repo.Status != config.StatusValid {
+		return true, fmt.Sprintf("The Status field content differs from '%s' for Repository.UUID = %s", config.StatusValid, repo.UUID)
```
rverdile, line 150:
```suggestion
		return true, fmt.Sprintf("Introspection started: the Status field content differs from '%s' for Repository.UUID = %s", config.StatusValid, repo.UUID)
```
Add "Introspection started" to the front, just so it is clear that introspection was run

pkg/external_repos/introspect.go
```diff
@@ -124,6 +141,27 @@ func IntrospectAll() (int64, []error) {
...
+	}
+
+	if repo.LastIntrospectionTime == nil {
+		return true, fmt.Sprintf("Not expected LastIntrospectionTime = nil for Repository.UUID = %s", repo.UUID)
```
rverdile, line 154:
```suggestion
		return true, fmt.Sprintf("Introspection started: not expected LastIntrospectionTime = nil for Repository.UUID = %s", repo.UUID)
```

pkg/external_repos/introspect.go
```diff
@@ -124,6 +141,27 @@ func IntrospectAll() (int64, []error) {
...
+	}
+
+	threshold := repo.LastIntrospectionTime.Add(IntrospectTimeInterval)
+	if threshold.After(time.Now()) {
+		return false, fmt.Sprintf("Last instrospection happened before the threshold for Repository.UUID = %s", repo.UUID)
```
rverdile, line 159:
```suggestion
		return false, fmt.Sprintf("Introspection skipped: Last instrospection happened before the threshold for Repository.UUID = %s", repo.UUID)
```

pkg/external_repos/introspect.go
```diff
@@ -124,6 +141,27 @@ func IntrospectAll() (int64, []error) {
...
+	}
+
+	return true, fmt.Sprintf("Last introspection happened after the threshold for Repository.UUID = %s", repo.UUID)
```
rverdile, line 162:
```suggestion
	return true, fmt.Sprintf("Introspection started: last introspection happened after the threshold for Repository.UUID = %s", repo.UUID)
```

rverdile:
I left a couple more comments about the output messaging, but otherwise seems to function well

avisiedo:
sorry for the delay! I am on this pr! thanks for the comments! looking into them! :+1:

pkg/external_repos/introspect.go
avisiedo: Looking at the library suggested by @rverdile to use it! thanks mates!

pkg/external_repos/external_repos.json
```diff
@@ -1,4 +1,52 @@
 [
+    {
+        "base_url": ""
+    },
+    {
+        "base_url": ""
+    },
```
rverdile, line 7: There's a bunch of new blank base_url's here, likely by mistake

cmd/external-repos/main.go
```diff
@@ -43,15 +48,33 @@ func main() {
...
+				os.Exit(1)
+			}
 		}
-		count, errors := external_repos.IntrospectUrl(args[2])
+		count, errors := external_repos.IntrospectUrl(url, forceIntrospect)
 		for i := 0; i < len(errors); i++ {
 			log.Panic().Err(errors[i]).Msg("Failed to introspect repository")
 		}
-		log.Debug().Msgf("Successfully Inserted %d packages", count)
+		log.Debug().Msgf("Inserted %d packages", count)
```
rverdile, line 67: I can't leave a comment on it because it's not a change, but "successfully" should be removed from the introspect-all result as well

pkg/external_repos/external_repos.json
jlsherrill: something is up with the 'make repos-download' command.  I'd just leave this out for now and i'll take a look

avisiedo: Thanks mate! I didn't realize about this! I have removed it.

cmd/external-repos/main.go
avisiedo: Updated! :+1:

rverdile (approved):
lgtm!

pkg/external_repos/introspect_test.go
```diff
@@ -273,3 +273,112 @@ func TestUpdateIntrospectionStatusMetadata(t *testing.T) {
...
+					LastIntrospectionTime: nil,
+				},
+				expected: TestCaseExpected{
+					result: true,
+					reason: "Introspection started: not expected LastIntrospectionTime = nil for Repository.UUID = ",
+				},
+			},
+			// When Status is Valid
+			//  and LastIntrospectionTime does not reach the threshold interval (24hours)
+			// It returns false indicating that no introspection is needed
```
swadeley, line 351:
```suggestion
			//  it returns false indicating that no introspection is needed
```

pkg/external_repos/introspect_test.go
```diff
@@ -273,3 +273,112 @@ func TestUpdateIntrospectionStatusMetadata(t *testing.T) {
...
+					LastIntrospectionTime: &thresholdBefore24,
+				},
+				expected: TestCaseExpected{
+					result: false,
+					reason: "Introspection skipped: Last instrospection happened before the threshold for Repository.UUID = ",
+				},
+			},
+			// When Status is Valid
+			//  and LastIntrospectionTime does reach the threshold interval (24hours)
+			// It returns true indicating that an introspection is needed
```
swadeley, line 364:
```suggestion
			//  it returns true indicating that an introspection is needed
```

mayurilahane:
note: Verified repo with (status != valid )

PR #92, opened 2022-08-30
Fixes 111: record package count

jlsherrill:

PR #110, opened 2022-09-20
Fixes 185: Save revision properly

rverdile:
The `Update` function in the repository dao was not working correctly when passed zero values. It would update the model with the zero values.

It may be best to have the [internal Repository object](https://github.com/content-services/content-sources-backend/blob/main/pkg/dao/repositories.go#L10) split into two the way that the RepoConfig layer has RepositoryRequest and RepositoryResponse. This way we can distinguish between what should be updated and what shouldn't be with `nil`. I can make a change like that here if we want.

For now I fixed this in the least disruptive way possible because it's blocking #92 

To test:
Introspect and repository and the Revision field should be saved correctly. Without this change, it would be blank.

### DIFF
--- a/pkg/dao/repositories.go
+++ b/pkg/dao/repositories.go
@@ -12,7 +12,6 @@ type Repository struct {
 	UUID                         string
 	URL                          string
 	Revision                     string
-	Public                       bool
 	LastIntrospectionTime        *time.Time
 	LastIntrospectionSuccessTime *time.Time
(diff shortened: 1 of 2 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-185

jlsherrill:
@rverdile mind filing a ticket to fix this in a more destructive way?

jlsherrill:
needs a rebase now, but looks good and works well

PR #112, opened 2022-09-21
Fixes 76: Add fetchGpgKey API

Andrewgdewar:


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-66

pkg/api/repository_parameters.go
```diff
@@ -2,6 +2,15 @@ package api
 
 import "github.com/content-services/content-sources-backend/pkg/config"
 
+// FetchGPGKeyResponse holds data returned by the FetchGpgKey API response
+type FetchGPGKeyResponse struct {
+	GpgKey string `json:"gpg_key" ` //The downloaded gpg-key from the provided url.
```
swadeley, line 7:
```suggestion
	GpgKey string `json:"gpg_key" ` // The downloaded gpg-key from the provided url.
```

pkg/dao/repository_config_test.go
```diff
@@ -754,7 +754,7 @@ func (suite *RepositoryConfigSuite) TestListFilterMultipleVersions() {
 	assert.Equal(t, quantity, len(response.Data))
 	assert.Equal(t, int64(quantity), count)
 
-	// By setting SortBy to "version asc" we expect the first page versions lengths to be half and half or 1 and 3
+	// By setting the above seed values and SortBy to "version asc", we expect the first page to contain 10 versions of length 1 and 10 versions of length 3
```
Andrewgdewar, line 757: Fixed this as per @swadely's recommendations from a previous PR. (Had github automerge issues)

Andrewgdewar, line 757: Fixed this as per @swadely's recommendations from a previous PR. (Had github automerge issues)

pkg/dao/external_resource.go
```diff
@@ -14,6 +18,36 @@ func GetExternalResourceDao() ExternalResourceDao {
 	return ExternalResourceDaoImpl{}
 }
 
+func (erd ExternalResourceDaoImpl) FetchGpgKey(url string) (string, error) {
+	timeout := 1 * time.Second // 10 Second timout
```
rverdile, line 22:
```suggestion
	timeout := 10 * time.Second // 10 Second timout
```
typo :)

Andrewgdewar: Oh lol.

pkg/dao/external_resource.go
```diff
@@ -14,6 +18,36 @@ func GetExternalResourceDao() ExternalResourceDao {
 	return ExternalResourceDaoImpl{}
 }
 
+func (erd ExternalResourceDaoImpl) FetchGpgKey(url string) (string, error) {
+	timeout := 10 * time.Second // 10 Second timout
```
jlsherrill, line 22: we probably want a lower timeout?

pkg/handler/repository_parameters_test.go
```diff
@@ -16,16 +16,32 @@ import (
...
+func (erd *MockExternalResourceDao) ValidRepoMD(url string) (int, error) {
+	args := erd.Called(url)
+	return args.Int(0), args.Error(1)
+}
+
+func (erd *MockExternalResourceDao) FetchGpgKey(url string) (string, error) {
+	args := erd.Called(url)
+	return args.String(), args.Error(1)
+}
+
```
jlsherrill, line 36: hrmm, we already have a mockExternalResource in pkg/dao/repository_config_test.go   I wonder if we should just move it to a new package like  'test' ?

pkg/dao/external_resource.go
```diff
@@ -46,3 +80,17 @@ func UrlToRepomdUrl(urlIn string) (string, error) {
...
+func validateGpgKeyString(bodyString string) error {
+	httpError := echo.NewHTTPError(http.StatusNotAcceptable, "Received response is not a valid gpg-key")
+	if len(bodyString) < 36 {
+		return httpError
+	}
+	firstLine := strings.ToLower(bodyString[0:36])
+
+	if strings.Contains(firstLine, "begin pgp public key block") {
+		return nil
+	}
```
jlsherrill, line 93: should we actuall try to parse it?

pkg/handler/repository_parameters_test.go
Andrewgdewar: Sure! can you tell me exactly where to move it from "xxx" to "xxx/xxx". 
Most of this structure still doesn't make a lot of sense to me, despite starting to see the pattern in the layout.

pkg/dao/external_resource.go
Andrewgdewar: Can you point to an example of what you mean?

rverdile:
this should be "Fixes 76", 66 is the story

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-76

pkg/dao/external_resource.go
jlsherrill: i guess i mean using https://github.com/ProtonMail/gopenpgp to parse the key and ensure at least one valid key is present: https://gist.github.com/jvehent/9994311#file-pgpkeydetail-go-L17

https://pkg.go.dev/golang.org/x/crypto/openpgp#ReadArmoredKeyRing

pkg/handler/repository_parameters_test.go
jlsherrill: so right now there is a mockExternalResource in pkg/dao/repository_config_test.go and now you've added a duplicate in pkg/handler/repository_parameters_test.go.  lets just have one and put it somewhere like:

.pkg/tests/mocks.go  ?

jlsherrill: and i could see us moving all the mock implementation there, so maybe it should be ./pkg/test/mocks/external_resources.go ?

pkg/dao/external_resource.go
jlsherrill: actually sorry, i think you'd use  ReadKeyRing maybe for an arbitrary string of keys?

jlsherrill: actually i'm not sure which we need, i'm dealing with this now on my task, so i'll let you know what i figure out

pkg/handler/repository_parameters_test.go
Andrewgdewar: I like it. Will do.

pkg/dao/external_resource.go
Andrewgdewar: The reason I went the route I did, was just to do a quick basic check on the string to make sure it had the correct syntax. If there is an equivalent of  .checkValid(publicKey) :boolean I'd love that.

pkg/handler/repository_parameters.go
```diff
@@ -12,15 +12,48 @@ import (
...
+	repoDao *dao.RepositoryConfigDao,
+	externalDao *dao.ExternalResourceDao,
+) {
+	rph := RepositoryParameterHandler{RepositoryDao: *repoDao, ExternalResourceDao: *externalDao}
 	engine.GET("/repository_parameters/", rph.listParameters)
+	engine.GET("/repository_parameters/external_gpg_key/", rph.fetchGpgKey)
 	engine.POST("/repository_parameters/validate/", rph.validate)
 }
 
+// FerchGpgKeys godoc
```
rverdile, line 30:
```suggestion
// FetchGpgKeys godoc
```

pkg/dao/external_resource.go
```diff
@@ -46,3 +80,17 @@ func UrlToRepomdUrl(urlIn string) (string, error) {
 	u.Path = path.Join(u.Path, "/repodata/repomd.xml")
 	return u.String(), nil
 }
+
+func validateGpgKeyString(bodyString string) error {
+	httpError := echo.NewHTTPError(http.StatusNotAcceptable, "Received response is not a valid gpg-key")
```
rverdile, line 85: Having echo in the dao layer feels illegal. Maybe this can be done in such a way that echo is not imported here?

pkg/dao/external_resource.go
Andrewgdewar: @jlsherrill, any update on this?

pkg/dao/external_resource.go
rverdile: Looking in `dao/repository_config.go`, we use this [error struct](https://github.com/content-services/content-sources-backend/blob/main/pkg/dao/error.go) along with this [DBErrorToApi](https://github.com/content-services/content-sources-backend/blob/main/pkg/dao/repository_config.go#L26) function. Then in the handler we use this [httpCodeForError](https://github.com/content-services/content-sources-backend/blob/main/pkg/handler/repositories.go#L136) function to map that to an http error. I think you could leverage the same functions here.

pkg/dao/external_resource.go
jlsherrill: its as simple as this:  `openpgp.ReadArmoredKeyRing(strings.NewReader(gpgKey))`  where gpgKey is a string with the gpg key.  Here's some sample usage:
```go
	gpgKey := "-----BEGIN PGP PUBLIC KEY BLOCK-----\nVersion: GnuPG v1.4.7 (GNU/Linux)\n\nmQENBFYxWIwBCADAKoZhZlJxGNGWzqV+1OG1xiQeoowKhssGAKvd+buXCGISZJwT\nLXZqIcIiLP7pqdcZWtE9bSc7yBY2MalDp9Liu0KekywQ6VVX1T72NPf5Ev6x6DLV\n7aVWsCzUAF+eb7DC9fPuFLEdxmOEYoPjzrQ7cCnSV4JQxAqhU4T6OjbvRazGl3ag\nOeizPXmRljMtUUttHQZnRhtlzkmwIrUivbfFPD+fEoHJ1+uIdfOzZX8/oKHKLe2j\nH632kvsNzJFlROVvGLYAk2WRcLu+RjjggixhwiB+Mu/A8Tf4V6b+YppS44q8EvVr\nM+QvY7LNSOffSO6Slsy9oisGTdfE39nC7pVRABEBAAG0N01pY3Jvc29mdCAoUmVs\nZWFzZSBzaWduaW5nKSA8Z3Bnc2VjdXJpdHlAbWljcm9zb2Z0LmNvbT6JATUEEwEC\nAB8FAlYxWIwCGwMGCwkIBwMCBBUCCAMDFgIBAh4BAheAAAoJEOs+lK2+EinPGpsH\n/32vKy29Hg51H9dfFJMx0/a/F+5vKeCeVqimvyTM04C+XENNuSbYZ3eRPHGHFLqe\nMNGxsfb7C7ZxEeW7J/vSzRgHxm7ZvESisUYRFq2sgkJ+HFERNrqfci45bdhmrUsy\n7SWw9ybxdFOkuQoyKD3tBmiGfONQMlBaOMWdAsic965rvJsd5zYaZZFI1UwTkFXV\nKJt3bp3Ngn1vEYXwijGTa+FXz6GLHueJwF0I7ug34DgUkAFvAs8Hacr2DRYxL5RJ\nXdNgj4Jd2/g6T9InmWT0hASljur+dJnzNiNCkbn9KbX7J/qK1IbR8y560yRmFsU+\nNdCFTW7wY0Fb1fWJ+/KTsC4=\n=J6gs\n-----END PGP PUBLIC KEY BLOCK-----\n"

	keyring, err := openpgp.ReadArmoredKeyRing(strings.NewReader(gpgKey))
	if err != nil {
		return err  //Bad key
	}
       //good key
```

Andrewgdewar: openpgp is deprecated, I used another package, but similar fix. Cheers

pkg/dao/external_resource.go
Andrewgdewar: That seems like some workaround. I have used a different package and caught the error higher up in the tree to avoid this instead.

pkg/handler/repository_parameters.go
```diff
@@ -12,15 +12,49 @@ import (
...
+	url := c.QueryParam("url")
+
+	if url == "" {
+		return echo.NewHTTPError(http.StatusBadRequest, "url cannot be blank")
+	}
+
+	gpgKey, err := rh.ExternalResourceDao.FetchGpgKey(url)
+
+	if err != nil {
+		httpError := echo.NewHTTPError(http.StatusNotAcceptable, "Received response was not a valid gpg-key")
```
Andrewgdewar, line 49: Is this response useful for the user? 
Screen Shot 2022-09-28 at 3 16 31 PM

Andrewgdewar:
Suggestions for a meaningful test for this without actually calling an external URL?

jlsherrill:
here's some examples:
http://yum.theforeman.org/RPM-GPG-KEY-foreman
http://yum.theforeman.org/releases/1.21/RPM-GPG-KEY-foreman
https://packages.microsoft.com/keys/microsoft.asc

pkg/dao/external_resource_test.go
```diff
@@ -65,3 +55,49 @@ func TestValidRepoMD(t *testing.T) {
...
+	ZWFzZSBzaWduaW5nKSA8Z3Bnc2VjdXJpdHlAbWljcm9zb2Z0LmNvbT6JATUEEwEC
+	AB8FAlYxWIwCGwMGCwkIBwMCBBUCCAMDFgIBAh4BAheAAAoJEOs+lK2+EinPGpsH
+	/32vKy29Hg51H9dfFJMx0/a/F+5vKeCeVqimvyTM04C+XENNuSbYZ3eRPHGHFLqe
+	MNGxsfb7C7ZxEeW7J/vSzRgHxm7ZvESisUYRFq2sgkJ+HFERNrqfci45bdhmrUsy
+	7SWw9ybxdFOkuQoyKD3tBmiGfONQMlBaOMWdAsic965rvJsd5zYaZZFI1UwTkFXV
+	KJt3bp3Ngn1vEYXwijGTa+FXz6GLHueJwF0I7ug34DgUkAFvAs8Hacr2DRYxL5RJ
+	XdNgj4Jd2/g6T9InmWT0hASljur+dJnzNiNCkbn9KbX7J/qK1IbR8y560yRmFsU+
+	NdCFTW7wY0Fb1fWJ+/KTsC4=
+	=J6gs
+	-----END PGP PUBLIC KEY BLOCK-----
```
jlsherrill, line 78:
```suggestion
Version: GnuPG v1.4.7 (GNU/Linux)

mQENBFYxWIwBCADAKoZhZlJxGNGWzqV+1OG1xiQeoowKhssGAKvd+buXCGISZJwT
LXZqIcIiLP7pqdcZWtE9bSc7yBY2MalDp9Liu0KekywQ6VVX1T72NPf5Ev6x6DLV
7aVWsCzUAF+eb7DC9fPuFLEdxmOEYoPjzrQ7cCnSV4JQxAqhU4T6OjbvRazGl3ag
OeizPXmRljMtUUttHQZnRhtlzkmwIrUivbfFPD+fEoHJ1+uIdfOzZX8/oKHKLe2j
H632kvsNzJFlROVvGLYAk2WRcLu+RjjggixhwiB+Mu/A8Tf4V6b+YppS44q8EvVr
M+QvY7LNSOffSO6Slsy9oisGTdfE39nC7pVRABEBAAG0N01pY3Jvc29mdCAoUmVs
ZWFzZSBzaWduaW5nKSA8Z3Bnc2VjdXJpdHlAbWljcm9zb2Z0LmNvbT6JATUEEwEC
AB8FAlYxWIwCGwMGCwkIBwMCBBUCCAMDFgIBAh4BAheAAAoJEOs+lK2+EinPGpsH
/32vKy29Hg51H9dfFJMx0/a/F+5vKeCeVqimvyTM04C+XENNuSbYZ3eRPHGHFLqe
MNGxsfb7C7ZxEeW7J/vSzRgHxm7ZvESisUYRFq2sgkJ+HFERNrqfci45bdhmrUsy
7SWw9ybxdFOkuQoyKD3tBmiGfONQMlBaOMWdAsic965rvJsd5zYaZZFI1UwTkFXV
KJt3bp3Ngn1vEYXwijGTa+FXz6GLHueJwF0I7ug34DgUkAFvAs8Hacr2DRYxL5RJ
XdNgj4Jd2/g6T9InmWT0hASljur+dJnzNiNCkbn9KbX7J/qK1IbR8y560yRmFsU+
NdCFTW7wY0Fb1fWJ+/KTsC4=
=J6gs
-----END PGP PUBLIC KEY BLOCK-----
```

pkg/dao/external_resource.go
```diff
@@ -14,6 +17,39 @@ func GetExternalResourceDao() ExternalResourceDao {
...
+
+	if err != nil {
+		return "", err
+	}
+
+	gpgKeyString := string(bodyBytes)
+
+	// print(">>>>>>>>>>>>>>>>>>>", gpgKeyString)
+
+	_, cryptoErr := crypto.NewKeyFromArmored(gpgKeyString)
```
jlsherrill, line 44: This will only load the first key.  It looks like there is a version of the old library that this library is still using (but presumably still maintained).  In your imports:

```
openpgp "github.com/ProtonMail/go-crypto/openpgp"
```
then here:
```
_, cryptoErr := openpgp.ReadArmoredKeyRing(strings.NewReader(gpgKeyString))
```

pkg/api/repository_parameters.go
```diff
@@ -2,6 +2,14 @@ package api
 
 import "github.com/content-services/content-sources-backend/pkg/config"
 
+type FetchGPGKeyResponse struct {
+	GpgKey string `json:"gpg_key" ` // The downloaded gpg-key from the provided url.
```
jlsherrill, line 6: Should probably refer to them as GPG Keys

pkg/api/repository_parameters.go
```diff
@@ -2,6 +2,14 @@ package api
...
+}
+
+type FetchGPGKeyRequest struct {
+	URL *string `json:"url"` // The url from which to download the gpg-key.
+}
```
jlsherrill, line 11: same here

jlsherrill:
small wording comment, otherwise looks good!

pkg/handler/repository_parameters.go
```diff
@@ -12,15 +12,49 @@ import (
...
-func RegisterRepositoryParameterRoutes(engine *echo.Group, repoDao *dao.RepositoryConfigDao) {
-	rph := RepositoryParameterHandler{RepositoryDao: *repoDao}
+func RegisterRepositoryParameterRoutes(
+	engine *echo.Group,
+	repoDao *dao.RepositoryConfigDao,
+	externalDao *dao.ExternalResourceDao,
+) {
+	rph := RepositoryParameterHandler{RepositoryDao: *repoDao, ExternalResourceDao: *externalDao}
 	engine.GET("/repository_parameters/", rph.listParameters)
+	engine.GET("/repository_parameters/external_gpg_key/", rph.fetchGpgKey)
```
jlsherrill, line 26: i kinda wonder if this should be a POST instead of a GET.   Since its fetching data from external servers, a mean person could embed a bunch of links on some page, which forces the visitor's browser to request those URLS.  This could slow our site, or potentially do something more malicious.

jlsherrill: i realized this while reviewing your frontend pr

Andrewgdewar: ahh... how would we prevent that with a POST??

jlsherrill: So for example, here's a get request example, create an html file with:

```
<html>
<img src="https://console.stage.redhat.com/api/content-sources/v1/ping"
</html>
```

If you login to console.stage.redhat.com within your browser and then browse to this html page, you'll see the ping request gets executed and you get a pong response.

If we try a similar thing with a post:

```
<html>

<script>

let xhr = new XMLHttpRequest();
xhr.open("POST", "https://console.stage.redhat.com/api/content-sources/v1/repositories/");
xhr.setRequestHeader("Accept", "application/json");
xhr.setRequestHeader("Content-Type", "application/json");

xhr.onreadystatechange = function () {
  if (xhr.readyState === 4) {
    console.log(xhr.status);
    console.log(xhr.responseText);
  }};

let data = `{
  "name": "fooobar34",
  "url": "http://yum.theforeman.org/pulpcore/3.14/el8/x86_64/",
}`;

xhr.send(data);
</script>

</html>
```                                     
It will fail due to cross site protection i believe, you get a forbidden with:  
Referrer Policy: strict-origin-when-cross-origin

Maybe this is related to the configuration of console dot.  But since i think this particular request probably shouldn't be cache-able, it also makes sense to be something other than a GET.

pkg/handler/repository_parameters.go
```diff
@@ -12,15 +12,49 @@ import (
...
 
+// FetchGpgKeys godoc
+// @Summary      Fetch gpgkey from URL
+// @ID           fetchGpgKey
+// @Description  Fetch gpgkey from URL
+// @Tags         gpgKey
+// @Accept       json
+// @Produce      json
+// @Success      200 {object} api.FetchGPGKeyResponse
+// @Router       /repository_parameters/external_gpg_key [get]
```
jlsherrill, line 38: this needs to be changed to post, and apidocs regenerated

pkg/handler/repository_parameters.go
jlsherrill: GPG Key :)

jlsherrill (approved):

rverdile:
I wonder why this didn't get marked "Ready for QE"?

PR #118, opened 2022-09-30
Fixes 187: speed up seeds_test

Andrewgdewar:
Currently seed test is the slowest test file, taking over a second.  

Some of this is due to 'batch size' for inserting, and a desire to test a number greater than that.  

This ticket makes the batch size configurable so the seeding can test with a loweramount


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-187

jlsherrill (approved):

PR #119, opened 2022-10-03
Fixes 186: Refactor test suites to run individual test files

rverdile:
This PR makes it possible (in GoLand) to right-click on a testify-suite-using test file like `dao/repository_config.go`, click "Run", and only run the tests in that file.

This PR should get a review from both a GoLand user and a VSCode user. I'm not sure how it affects VSCode.

Also, I refactored the dao test suite to embedded a new `DaoSuite` type, which deduplicates some code. I did the same thing in the models test suite, but it's a more obvious change there.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -64,9 +64,10 @@ func Get() *Configuration {
 }
 
 func readConfigFile(v *viper.Viper) {
-	v.SetConfigName("config")
+	v.SetConfigName("config.yaml")
 	v.SetConfigType("yaml")
 	v.AddConfigPath("./configs/")
+	v.AddConfigPath("../../configs/")
 
 	path, set := os.LookupEnv("CONFIG_PATH")
 	if set {
--- a/pkg/dao/suite_test.go
+++ b/pkg/dao/suite_test.go
@@ -1,46 +1,21 @@
 package dao
 
 import (
-	"log"
-	"os"
-	"testing"
 	"time"
 
 	"github.com/content-services/content-sources-backend/pkg/config"
-	"github.com/content-services/content-sources-backend/pkg/db"
 	"github.com/content-services/content-sources-backend/pkg/models"
 	"github.com/content-services/content-sources-backend/pkg/seeds"
 	"github.com/lib/pq"
 	"github.com/stretchr/testify/suite"
 	"gorm.io/gorm"
-	"gorm.io/gorm/logger"
 )
 
-type RepositoryConfigSuite struct {
+type DaoSuite struct {
 	suite.Suite
 	db                        *gorm.DB
 	tx                        *gorm.DB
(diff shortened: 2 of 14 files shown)


pkg/config/config.go
```diff
@@ -64,9 +64,10 @@ func Get() *Configuration {
 }
 
 func readConfigFile(v *viper.Viper) {
-	v.SetConfigName("config")
+	v.SetConfigName("config.yaml")
 	v.SetConfigType("yaml")
 	v.AddConfigPath("./configs/")
+	v.AddConfigPath("../../configs/")
```
rverdile, line 70: I'm not sure reading this config ever worked? Unless I'm breaking something here I don't know about. These changes let me use this locally.

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-186

pkg/dao/suite_test.go
```diff
@@ -1,48 +1,23 @@
...
 	"github.com/content-services/content-sources-backend/pkg/models"
 	"github.com/content-services/content-sources-backend/pkg/seeds"
 	"github.com/lib/pq"
 	"github.com/stretchr/testify/suite"
 	"gorm.io/gorm"
-	"gorm.io/gorm/logger"
 )
 
-type RepositoryConfigSuite struct {
+type ModelsSuite struct {
```
rverdile, line 14: This should actually be DaoSuite

pkg/config/config.go
jlsherrill: what didn't work without these changes?

rverdile: running tests using the variables in config file

jlsherrill:
This all looks good and works great in both goland and vscode.  Just one question and then i think its good to go

rverdile:
> This all looks good and works great in both goland and vscode. Just one question and then i think its good to go

I think I answered that question

pkg/config/config.go
jlsherrill: how did you run the test?

jlsherrill: `go test ./pkg/config/ `  seems to work fine for me without these two lines.

rverdile: hmm, I'm not sure why that behaves differently, but if you run `go test ./...`, tests that require db credentials will fail.

jlsherrill: oh i see!  yes, our make file is specifying CONFIG_PATH="/home/jlsherri/git/content-sources-backend/configs/"  which makes it work, but go test by itself didn't.  +1 to this change!

jlsherrill (approved):

pkg/config/config.go
rverdile: Oh I think it's different because it sets an environment variable, so it's not reading from the config

PR #120, opened 2022-10-04
Fixes 202: Log info and debug requests

rverdile:
Now when a successful request is made, there should be "debug" and/or "info" logs in the console. Previously, only errors were logged.


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-202

jlsherrill (approved):

PR #121, opened 2022-10-07
Fixes 203: Add URL trailing slash on creation

rverdile:
Adds URL trailing slash on creation, if missing, and removes extra trailing slashes (e.g. http://example/path/////)

Also:
* Trims leading/trailing whitespace from URL before creation
* Reject requests where URLs contain other whitespace

### DIFF
--- a/pkg/dao/repository_config.go
+++ b/pkg/dao/repository_config.go
@@ -32,6 +32,8 @@ func DBErrorToApi(e error) *Error {
 			switch pgError.ConstraintName {
 			case "repo_and_org_id_unique":
 				dupKeyName = "URL"
+			case "repositories_unique_url":
+				dupKeyName = "URL"
 			case "name_and_org_id_unique":
 				dupKeyName = "name"
 			}
@@ -419,11 +421,12 @@ func (r repositoryConfigDaoImpl) ValidateParameters(orgId string, params api.Rep
 	if params.URL == nil {
 		response.URL.Skipped = true
 	} else {
-		err = r.validateUrl(orgId, *params.URL, &response, excludedUUIDS)
+		url := models.CleanupURL(*params.URL)
+		err = r.validateUrl(orgId, url, &response, excludedUUIDS)
 		if err != nil {
 			return response, err
 		} else if response.URL.Valid {
-			repomd := r.validateMetadataPresence(*params.URL, &response)
+			repomd := r.validateMetadataPresence(url, &response)
 			if response.URL.MetadataPresent {
 				r.checkSignaturePresent(&params, repomd, &response)
 			}
@@ -436,23 +439,26 @@ func (r repositoryConfigDaoImpl) validateName(orgId string, name string, respons
 	if name == "" {
 		response.Valid = false
 		response.Error = "Name cannot be blank"
-	} else {
-		found := models.RepositoryConfiguration{}
-		query := r.db.Where("name = ? AND ORG_ID = ?", name, orgId)
-		if len(excludedUUIDS) != 0 {
-			query = query.Where("repository_configurations.uuid NOT IN ?", excludedUUIDS)
-		}
-		result := query.Find(&found)
-		if result.Error != nil {
-			response.Valid = false
-			return result.Error
-		} else if found.UUID != "" {
-			response.Valid = false
-			response.Error = fmt.Sprintf("A repository with the name '%s' already exists.", name)
-		} else {
-			response.Valid = true
-		}
+		return nil
 	}
+
+	found := models.RepositoryConfiguration{}
+	query := r.db.Where("name = ? AND ORG_ID = ?", name, orgId)
+	if len(excludedUUIDS) != 0 {
+		query = query.Where("repository_configurations.uuid NOT IN ?", excludedUUIDS)
+	}
+	if err := query.Find(&found).Error; err != nil {
+		response.Valid = false
+		return err
+	}
+
+	if found.UUID != "" {
+		response.Valid = false
+		response.Error = fmt.Sprintf("A repository with the name '%s' already exists.", name)
+		return nil
+	}
+
+	response.Valid = true
 	return nil
 }
 
@@ -460,25 +466,35 @@ func (r repositoryConfigDaoImpl) validateUrl(orgId string, url string, response 
 	if url == "" {
 		response.URL.Valid = false
 		response.URL.Error = "URL cannot be blank"
-	} else {
-		found := models.RepositoryConfiguration{}
-		query := r.db.Preload("Repository").
-			Joins("inner join repositories on repository_configurations.repository_uuid = repositories.uuid").
-			Where("Repositories.URL = ? AND ORG_ID = ?", url, orgId)
-		if len(excludedUUIDS) != 0 {
-			query = query.Where("repository_configurations.uuid NOT IN ?", excludedUUIDS)
-		}
-		result := query.Find(&found)
-		if result.Error != nil {
-			response.URL.Valid = false
-			return result.Error
-		} else if found.UUID != "" {
-			response.URL.Valid = false
-			response.URL.Error = fmt.Sprintf("A repository with the URL '%s' already exists.", url)
-		} else {
-			response.URL.Valid = true
-		}
+		return nil
 	}
+
+	found := models.RepositoryConfiguration{}
+	query := r.db.Preload("Repository").
+		Joins("inner join repositories on repository_configurations.repository_uuid = repositories.uuid").
+		Where("Repositories.URL = ? AND ORG_ID = ?", url, orgId)
+	if len(excludedUUIDS) != 0 {
+		query = query.Where("repository_configurations.uuid NOT IN ?", excludedUUIDS)
+	}
+	if err := query.Find(&found).Error; err != nil {
+		response.URL.Valid = false
+		return err
+	}
+
+	if found.UUID != "" {
+		response.URL.Valid = false
+		response.URL.Error = fmt.Sprintf("A repository with the URL '%s' already exists.", url)
+		return nil
+	}
+
+	containsWhitespace := strings.ContainsAny(strings.TrimSpace(url), " \t\n\v\r\f")
+	if containsWhitespace {
+		response.URL.Valid = false
+		response.URL.Error = "URL cannot contain whitespace."
+		return nil
+	}
+
+	response.URL.Valid = true
 	return nil
 }
 
--- a/pkg/dao/repository_config_test.go
+++ b/pkg/dao/repository_config_test.go
@@ -43,7 +43,7 @@ func TestRepositoryConfigSuite(t *testing.T) {
 
 func (suite *RepositoryConfigSuite) TestCreate() {
 	name := "Updated"
-	url := "http://someUrl.com"
+	url := "http://example.com/"
 	orgID := seeds.RandomOrgId()
 	accountId := seeds.RandomAccountId()
 	distributionArch := "x86_64"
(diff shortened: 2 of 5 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-203

jlsherrill:
one small testing comment  otherwise this all looks good!

pkg/models/repository_test.go
```diff
@@ -34,4 +35,43 @@ func (s *RepositorySuite) TestRepositoriesCreate() {
...
+			URL: testCases[i].given,
+		}
+
+		tx.Create(&repo)
+		assert.Nil(s.T(), tx.Error)
+
+		err := tx.Where("url = ?", testCases[i].expected).Find(&found).Error
+		assert.Nil(s.T(), err)
+		assert.NotEmpty(s.T(), found.UUID)
+	}
```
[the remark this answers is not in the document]
rverdile: What you're seeing there is, when you update from the API, either the Repository exists or a new one is created. The Repository column/model itself is never updated (there's not even a BeforeUpdate hook there). So far the only thing that updates repositories is introspect, but I don't think that would change the URL.

I can add a hook and stuff to the Repository model if you think it should be there?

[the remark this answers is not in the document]
jlsherrill: ohh right, updating a repo config creates a new repository, makes sense!  I think its okay for now

pkg/dao/repository_config.go
```diff
@@ -444,48 +446,62 @@ func (r repositoryConfigDaoImpl) ValidateName(orgId string, name string, respons
...
-		} else {
-			response.URL.Valid = true
-		}
+		return nil
+	}
+
+	found := models.RepositoryConfiguration{}
+	query := r.db.Preload("Repository").
+		Joins("inner join repositories on repository_configurations.repository_uuid = repositories.uuid").
+		Where("Repositories.URL = ? AND ORG_ID = ?", url, orgId)
```
jlsherrill, line 482: wouldn't you want to call CleanupURL on the url to make sure the check is checking the same as what we're creating?

pkg/dao/repository_config.go
```diff
@@ -444,48 +446,62 @@ func (r repositoryConfigDaoImpl) ValidateName(orgId string, name string, respons
...
+		return nil
 	}
+
+	foundRepository := found.Repository
+	containsWhitespace := strings.ContainsAny(strings.TrimSpace(foundRepository.URL), " \t\n\v\r\f")
+	if containsWhitespace {
+		response.URL.Valid = false
+		response.URL.Error = "URL cannot contain whitespace."
+		return nil
+	}
```
jlsherrill, line 503: since we're stripping whitespace at creation time, is this check needed?

+1 to the small refactor here

rverdile: This is checking for non leading/trailing whitespace. Only the leading/trailing whitespace gets stripped.

pkg/dao/repository_config.go
rverdile: you're right, I'll adjust this

pkg/dao/repository_config.go
```diff
@@ -418,11 +420,12 @@ func (r repositoryConfigDaoImpl) ValidateParameters(orgId string, params api.Rep
 	if params.URL == nil {
 		response.URL.Skipped = true
 	} else {
-		err = r.ValidateUrl(orgId, *params.URL, &response, excludedUUIDS)
+		url := models.CleanupURL(*params.URL)
```
rverdile, line 423: moved this here

pkg/dao/repository_config.go
jlsherrill: ah!

pkg/dao/repository_config.go
```diff
@@ -444,48 +447,62 @@ func (r repositoryConfigDaoImpl) ValidateName(orgId string, name string, respons
...
+	}
+
+	if found.UUID != "" {
+		response.URL.Valid = false
+		response.URL.Error = fmt.Sprintf("A repository with the URL '%s' already exists.", url)
+		return nil
 	}
+
+	foundRepository := found.Repository
+	containsWhitespace := strings.ContainsAny(strings.TrimSpace(foundRepository.URL), " \t\n\v\r\f")
```
jlsherrill, line 499:
```suggestion
	containsWhitespace := strings.ContainsAny(strings.TrimSpace(url), " \t\n\v\r\f")
```

jlsherrill: otherwise it pulls what was found from the db, which if nothing is found would be ""

pkg/models/repository.go
```diff
@@ -27,12 +28,41 @@ func (r *Repository) BeforeCreate(tx *gorm.DB) (err error) {
 	if err := r.Base.BeforeCreate(tx); err != nil {
 		return err
 	}
+	if err := r.validate(); err != nil {
```
Andrewgdewar, line 31: Question,  so when a user enters a URL that has " https://some.url.com " 
Would this "validate" throw an error? 

As I understand it your CleanupURL function would remove the above leading/trailing whitespace.
So would you not want to first check for empty string, and then run cleanupURL, then run the stringContainsWhitespace (if at all) to check for internal whitespace? 

I'm probably misunderstanding this.

rverdile: stringContainsWhitespace also trims the leading/trailing whitespace

Andrewgdewar: ahh.. I see the strings.TrimSpace(s), this makes more sense now.

Andrewgdewar (approved):
This looks good to me!

pkg/models/repository.go
Andrewgdewar: An optional change you could do would be to change the name from stringContainsWhitespace to something like: 
stringContainsInternalWhitespace?

Andrewgdewar: I'm bad with names..

pkg/dao/repository_config_test.go
```diff
@@ -967,13 +967,13 @@ func (suite *RepositoryConfigSuite) TestValidateParameters() {
 			URL:  pointy.String(""),
 		}, { // Providing a valid url & name
 			Name: pointy.String("Some Other Name"),
-			URL:  pointy.String("http://foobar.com"),
+			URL:  pointy.String("http://foobar.com/"),
```
swadeley, line 970: The gives ERROR 403: Forbidden. I could not find a special use domain name that provides that. Do you need that error here?

pkg/dao/repository_config_test.go
```diff
@@ -967,13 +967,13 @@ func (suite *RepositoryConfigSuite) TestValidateParameters() {
...
 		}, { // Providing a bad url that doesn't have a repo
 			Name: pointy.String("Some bad repo!"),
-			URL:  pointy.String("http://badrepo.com"),
+			URL:  pointy.String("http://badrepo.com/"),
```
swadeley, line 973: "badrepo.com
is parked free, courtesy of GoDaddy.com."
Can you use 
```suggestion
			URL:  pointy.String("http://badrepo.test/"),
```

pkg/dao/repository_config_test.go
```diff
@@ -1016,7 +1016,7 @@ func (suite *RepositoryConfigSuite) TestValidateParameters() {
 	assert.False(t, response.URL.Skipped)
 	assert.Contains(t, response.URL.Error, "blank")
 
-	mockExtRDao.Mock.On("ValidRepoMD", "http://foobar.com").Return(200, nil)
+	mockExtRDao.Mock.On("ValidRepoMD", "http://foobar.com/").Return(200, nil)
```
swadeley, line 1019: http://foobar.com/ returns 403 not 200
Try example.com if you want 200 responce code.

pkg/dao/repository_config_test.go
```diff
@@ -1027,7 +1027,7 @@ func (suite *RepositoryConfigSuite) TestValidateParameters() {
 	assert.True(t, response.URL.MetadataPresent)
 	assert.False(t, response.URL.Skipped)
 
-	mockExtRDao.Mock.On("ValidRepoMD", "http://badrepo.com").Return(404, nil)
+	mockExtRDao.Mock.On("ValidRepoMD", "http://badrepo.com/").Return(404, nil)
```
swadeley, line 1030: see comments above, can you use
```suggestion
	mockExtRDao.Mock.On("ValidRepoMD", "http://badrepo.test/").Return(404, nil)
```

pkg/dao/repository_config_test.go
```diff
@@ -1044,7 +1044,7 @@ func (suite *RepositoryConfigSuite) TestValidateParameters() {
 		Timeout: true,
 	}
 
-	mockExtRDao.Mock.On("ValidRepoMD", "http://timemeout.com").Return(0, timeoutErr)
+	mockExtRDao.Mock.On("ValidRepoMD", "http://timemeout.com/").Return(0, timeoutErr)
```
swadeley, line 1047: I could not find a special use domain name that provides a time out. Hope the domain does not change hands soon.

pkg/models/repository_test.go
```diff
@@ -34,4 +35,43 @@ func (s *RepositorySuite) TestRepositoriesCreate() {
...
+	var found Repository
+
+	type TestCase struct {
+		given    string
+		expected string
+	}
+
+	testCases := []TestCase{
+		{
+			given:    "https://example1.com",
```
swadeley, line 52: can you use .test or .invalid?
```suggestion
			given:    "https://example1.test",
```
I see that domain has some text, `example.com` would also display some test.

pkg/models/repository_test.go
```diff
@@ -34,4 +35,43 @@ func (s *RepositorySuite) TestRepositoriesCreate() {
...
+			expected: "https://example1.com/",
+		},
+		{
+			given:    "   https://example2.com   ",
```
swadeley, line 56: Can you use a special use domain name here?
https://www.iana.org/assignments/special-use-domain-names/special-use-domain-names.xhtml

pkg/models/repository_test.go
```diff
@@ -34,4 +35,43 @@ func (s *RepositorySuite) TestRepositoriesCreate() {
...
+			expected: "https://example2.com/",
+		},
+		{
+			given:    "https://example3.com/path/////",
```
swadeley, line 60: Can yo use a special use domain name here?
https://www.iana.org/assignments/special-use-domain-names/special-use-domain-names.xhtml

swadeley (changes requested):
Please review the use of domain names with unknown owners.

pkg/dao/repository_config_test.go
```diff
@@ -42,7 +42,7 @@ func TestRepositoryConfigSuite(t *testing.T) {
 
 func (suite *RepositoryConfigSuite) TestCreate() {
 	name := "Updated"
-	url := "http://someUrl.com"
+	url := "http://someUrl.com/"
```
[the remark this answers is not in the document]
rverdile: I can change the names of the URLs here, but none of them are being reached out to. The names are completely arbitrary.

pkg/dao/repository_config_test.go
rverdile: I can change this, but the URL doesn't need to return anything here. The 200 response code is being mocked. In other words, this line means "return 200 for this URL".

pkg/dao/repository_config_test.go
rverdile: In this case we're not actually using the URL for its response, we're mocking the response. What we use it for is to parse a string with URL-style syntax. So it's okay that the URL doesn't work.

pkg/dao/repository_config_test.go
jlsherrill: we could just use timeout.example.com. In this instance the actual domain doesn't matter

rverdile:
updated. everything should be addressed now. most of the renames were resolved just by rebasing

swadeley (approved):
lgtm

PR #123, opened 2022-10-13
Fixes 227: package_count=0 after introspect

avisiedo:
- update Introspect to pass dao.Repository by reference, and update callers.

> Sorry, I didn't realize about this during the pr reviews

TODO:

- [x] Add unit test for this change.

### DIFF
--- a/pkg/external_repos/introspect_test.go
+++ b/pkg/external_repos/introspect_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/openlyinc/pointy"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestIsRedHatUrl(t *testing.T) {
@@ -92,6 +93,7 @@ func TestIntrospect(t *testing.T) {
 		}
 	}))
 	defer server.Close()
+
 	mockRepoDao := MockRepositoryDao{}
 
 	repoUUID := uuid.NewString()
@@ -107,18 +109,20 @@ func TestIntrospect(t *testing.T) {
 	mockRepoDao.On("Update", repoUpdate).Return(nil).Times(1)
 
 	count, err := Introspect(
-		dao.Repository{
-			UUID: repoUUID,
-			URL:  server.URL + "/content",
+		&dao.Repository{
+			UUID:         repoUUID,
+			URL:          server.URL + "/content",
+			PackageCount: 0,
 		},
 		&mockRepoDao,
 		MockRpmDao{})
 	assert.NoError(t, err)
 	assert.Equal(t, int64(13), count)
+	assert.Equal(t, 13, expected.PackageCount)
 
 	// Without any changes to the repo, there should be no package updates
 	count, err = Introspect(
-		dao.Repository{
+		&dao.Repository{
 			UUID:         repoUUID,
 			URL:          server.URL + "/content",
 			Revision:     revisionNumber,
@@ -128,6 +132,8 @@ func TestIntrospect(t *testing.T) {
 		MockRpmDao{})
 	assert.NoError(t, err)
 	assert.Equal(t, int64(0), count)
+	assert.Equal(t, 13, expected.PackageCount)
+
 	mockRepoDao.Mock.AssertExpectations(t)
 }
 
@@ -155,14 +161,15 @@ func TestUpdateIntrospectionStatusMetadata(t *testing.T) {
 	}
 
 	type TestCase struct {
+		name     string
 		given    TestCaseGiven
 		expected dao.Repository
 	}
 
 	timestamp := time.Now()
 	testCases := []TestCase{
-		// Status change from Pending to Valid
 		{
+			name: "Status change from Pending to Valid",
 			given: TestCaseGiven{
 				status: config.StatusPending,
 				count:  1,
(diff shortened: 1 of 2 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-227

pkg/external_repos/introspect_test.go
```diff
@@ -174,10 +181,11 @@ func TestUpdateIntrospectionStatusMetadata(t *testing.T) {
 				LastIntrospectionUpdateTime:  &timestamp,
 				LastIntrospectionError:       pointy.String(""),
 				Status:                       config.StatusValid,
+				PackageCount:                 *pointy.Int(100),
```
rverdile, line 184: I haven't tested it, but I'm pretty sure `*pointy.Int(100)` can just be `100`

There are a few places where this `*pointy.Int(100)` exists, but I'll just leave the comment here.

pkg/external_repos/introspect_test.go
```diff
@@ -230,10 +241,59 @@ func TestUpdateIntrospectionStatusMetadata(t *testing.T) {
...
+				LastIntrospectionTime:        &timestamp,
+				LastIntrospectionUpdateTime:  nil,
+				LastIntrospectionSuccessTime: nil,
+				LastIntrospectionError:       pointy.String("Error ramins Unavailable"),
+				Status:                       config.StatusUnavailable,
+				PackageCount:                 *pointy.Int(100),
+			},
+		},
+		{
+			name: "Status is set to Unavailable when error an any other current status",
```
rverdile, line 280: What if you rename this to "Status change to Unavailable" and make the TestCaseGiven status "AnythingExceptPending", then remove the test case for Valid to Unavailable (since this overlaps with that).

pkg/external_repos/introspect_test.go
```diff
@@ -230,10 +241,59 @@ func TestUpdateIntrospectionStatusMetadata(t *testing.T) {
...
+			name: "Status remains as Invalid",
+			given: TestCaseGiven{
+				status: config.StatusInvalid,
+				count:  1,
+				err:    fmt.Errorf("Error remains, keep it as Invalid"),
+			},
+			expected: dao.Repository{
+				LastIntrospectionTime:        &timestamp,
+				LastIntrospectionUpdateTime:  nil,
+				LastIntrospectionSuccessTime: nil,
```
rverdile, line 257: Since these are nil by default, I think it's okay to leave them out.

pkg/external_repos/introspect_test.go
avisiedo: you are right!! abusing the IDE features! :+1:  I have removed all of them.

pkg/external_repos/introspect_test.go
avisiedo: thanks mate! I have tried it, but it leaves the line below not covered :

```golang
	case config.StatusValid:
		output.Status = config.StatusUnavailable // Repository introspected successfully at least once, but now errors
```

so leaving as it is.

pkg/external_repos/introspect_test.go
```diff
@@ -244,34 +304,61 @@ func TestUpdateIntrospectionStatusMetadata(t *testing.T) {
...
+			testCase.given.count,
+			testCase.given.err,
+			&timestamp)
+
+		if testCase.expected.LastIntrospectionError != nil {
+			require.NotNil(t, result.LastIntrospectionError)
+			assert.Equal(t, *testCase.expected.LastIntrospectionError, *result.LastIntrospectionError)
+		} else {
+			assert.Nil(t, result.LastIntrospectionError)
+		}
```
rverdile, line 337: What is the purpose of the if/else blocks here? Isn't it sufficient just to test that expected and result are equal? If expected is nil, result should be nil. If expected as some value, result should have that value.

pkg/external_repos/introspect_test.go
```diff
@@ -230,10 +241,57 @@ func TestUpdateIntrospectionStatusMetadata(t *testing.T) {
...
+			name: "Status remains as Unavailable",
+			given: TestCaseGiven{
+				status: config.StatusUnavailable,
+				count:  1,
+				err:    fmt.Errorf("Error ramins Unavailable"),
+			},
+			expected: dao.Repository{
+				LastIntrospectionTime:        &timestamp,
+				LastIntrospectionUpdateTime:  nil,
+				LastIntrospectionSuccessTime: nil,
```
rverdile, line 271: also remove nils here

pkg/external_repos/introspect_test.go
```diff
@@ -230,10 +241,57 @@ func TestUpdateIntrospectionStatusMetadata(t *testing.T) {
...
+			name: "Status change to Unavailable",
+			given: TestCaseGiven{
+				status: "AnythingExceptPending",
+				count:  1,
+				err:    fmt.Errorf("Error set to Unavailable"),
+			},
+			expected: dao.Repository{
+				LastIntrospectionTime:        &timestamp,
+				LastIntrospectionUpdateTime:  nil,
+				LastIntrospectionSuccessTime: nil,
```
rverdile, line 287: also remove nils here

pkg/external_repos/introspect_test.go
rverdile: there are a couple of other places where the nils are being explicitly defined, so I left two more comments to point them out.

pkg/external_repos/introspect_test.go
rverdile: oh I see! It's testing the default case

pkg/external_repos/introspect_test.go
avisiedo: you are right; I was thinking about them that only the pointer were being compared; and in the case of leverage the data referenced, I was adding a require before. I have dig into the Equal, and it makes a deepequal of the fields using reflection, so when the pointers differs, it try to compare the data they are pointing out. That simplify many of the tests. Thanks mate!

rverdile (approved):
looks good!

PR #124, opened 2022-10-17
Fixes 212: Add sort/search to rpm List

Andrewgdewar:
Adds sort/search functionality to the rpm list endpoint. 

This PR is paired with the front-end PR found [here.](https://github.com/content-services/content-sources-frontend/pull/46)

Still needs: 
- [x] Tests

### DIFF
--- a/pkg/api/repository_rpm.go
+++ b/pkg/api/repository_rpm.go
@@ -17,6 +17,12 @@ type RepositoryRpmCollectionResponse struct {
 	Links Links            `json:"links"` // Links to other pages of results
 }
 
+type RepositoryRpmRequest struct {
+	UUID   string `param:"uuid"`    // Identifier of the repository
+	Search string `query:"search"`  // Search string based query to optionally filter-on
+	SortBy string `query:"sort_by"` // SortBy sets the sort order of the result
+}
+
 type SearchRpmRequest struct {
 	URLs   []string `json:"urls,omitempty"`  // URLs of repositories to search
 	UUIDs  []string `json:"uuids,omitempty"` // List of RepositoryConfig UUIDs to search
--- a/pkg/dao/helpers.go
+++ b/pkg/dao/helpers.go
@@ -1,0 +1,40 @@
+package dao
+
+import "strings"
+
+func convertSortByToSQL(SortBy string, SortMap map[string]string) string {
+	sqlOrderBy := ""
+
+	sortByArray := strings.Split(SortBy, ",")
+	lengthOfSortByParams := len(sortByArray)
+
+	for i := 0; i < lengthOfSortByParams; i++ {
+		sortBy := sortByArray[i]
+
+		split := strings.Split(sortBy, ":")
+		ascOrDesc := " asc"
+
+		if len(split) > 1 && split[1] == "desc" {
+			ascOrDesc = " desc"
+		}
+
+		sortField, ok := SortMap[strings.TrimSpace(split[0])]
+
+		// Only update if the SortMap above returns a valid value
+		if ok {
+			// Concatenate (e.g. "url desc," + "name" + " asc")
+			sqlOrderBy = sqlOrderBy + sortField + ascOrDesc
+
+			// Add a comma if this isn't the last item in the "sortByArray".
+			if i+1 < lengthOfSortByParams {
+				sqlOrderBy = sqlOrderBy + ", "
+			}
+		}
+	}
+
+	if sqlOrderBy == "" {
+		sqlOrderBy = "name asc"
+	}
+
+	return sqlOrderBy
+}
--- a/pkg/dao/helpers_test.go
+++ b/pkg/dao/helpers_test.go
@@ -1,0 +1,37 @@
+package dao
+
+import (
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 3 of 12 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-212

pkg/api/repository_rpm.go
```diff
@@ -17,6 +17,12 @@ type RepositoryRpmCollectionResponse struct {
 	Links Links            `json:"links"` // Links to other pages of results
 }
 
+type RepositoryRpmRequest struct {
+	UUID   string `param:"uuid"`    // Identifier of the rpm
```
jlsherrill, line 21: This is the repository uuid right?

pkg/handler/repository_rpm.go
```diff
@@ -74,16 +76,17 @@ func (rh *RepositoryRpmHandler) searchRpmPreprocessInput(input *api.SearchRpmReq
 // @Router       /repositories/{uuid}/rpms [get]
```
jlsherrill, line 76: need to add @Param options here similar to the search on repository list

Andrewgdewar: roger! 🖖

pkg/dao/rpm.go
```diff
@@ -417,3 +429,47 @@ func FilteredConvert(yumPkgs []yum.Package, excludeChecksums []string) []models.
...
+		if ok {
+			// Concatenate (e.g. "url desc," + "name" + " asc")
+			sqlOrderBy = sqlOrderBy + sortField + ascOrDesc
+
+			// Add a comma if this isn't the last item in the "sortByArray".
+			if i+1 < lengthOfSortByParams {
+				sqlOrderBy = sqlOrderBy + ", "
+			}
+		}
+	}
```
jlsherrill, line 468: can we pull most of this to a helper function that could also be used by  dao/repository_config.go?

Andrewgdewar: I was thinking this too, but where would that helper function live?

pkg/handler/repository_rpm.go
Andrewgdewar: add this here and to repository list... where I may have neglected that in the past. 🙄

pkg/dao/rpm.go
rverdile: I don't think there's a precedent for a function shared between two dao files (except in tests). Maybe a `dao/helpers.go` file?

pkg/dao/helpers.go
```diff
@@ -0,0 +1,40 @@
+package dao
+
+import "strings"
+
+func convertSortByToSQL(SortBy string, SortMap map[string]string) string {
```
rverdile, line 5: you could add a test just for this function

Andrewgdewar: Rog!

pkg/handler/repository_rpm.go
```diff
@@ -70,20 +72,25 @@ func (rh *RepositoryRpmHandler) searchRpmPreprocessInput(input *api.SearchRpmReq
 // @Accept       json
 // @Produce      json
 // @Param		 uuid	path string true "Identifier of the Repository"
+// @Param		 limit query int false "Limit the number of items returned"
+// @Param		 offset query int false "Offset into the list of results to return in the response"
+// @Param		 search query string false "Search term for name and url."
```
jlsherrill, line 77:
```suggestion
// @Param		 search query string false "Search term for name."
```

pkg/dao/rpm.go
jlsherrill: works4me!

pkg/dao/helpers_test.go
```diff
@@ -0,0 +1,34 @@
...
+		"name":          "name",
+		"url":           "url",
+		"package_count": "banana",
+		"status":        "status",
+	}
+
+	asc := ":asc"
+	desc := ":desc"
+
+	result := convertSortByToSQL("name", sortMap)
```
rverdile, line 20: looking at the function, am I right that it supports multiple sort parameters i.e. "name:asc,version:desc"? If so I think there should be a test for that

rverdile: could also test passing it an empty string

Andrewgdewar: That gave me a good idea about trimming whitespace..

pkg/dao/helpers_test.go
```diff
@@ -0,0 +1,37 @@
...
+	result = convertSortByToSQL("url"+asc, sortMap)
+	assert.Equal(t, "url asc", result)
+
+	result = convertSortByToSQL("package_count", sortMap)
+	assert.Equal(t, "banana asc", result)
+
+	result = convertSortByToSQL("status"+desc, sortMap)
+	assert.Equal(t, "status desc", result)
+
+	result = convertSortByToSQL(" status , name:desc", sortMap)
```
Andrewgdewar, line 35: This tests when multiple sortable items are listed, and also tests the new trimming of whitespace.

rverdile (commented):
Testing changes lgtm. I tested this with the frontend and it seems to be working. I think that's all from me.

jlsherrill (approved):
ack from me, needs a rebase though

PR #127, opened 2022-10-21
Fixes 207: return 404 when uuid syntax invalid

rverdile:
To test:
Make a PUT/PATCH/DELETE request where the UUID has invalid syntax i.e. /repostories/bad-uuid

Previously, the API would return 500 error and complain about invalid UUID syntax. Now, it returns a 404.


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-207

Andrewgdewar (approved):
Make it so!

swadeley (commented):
lgtm

PR #131, opened 2022-10-27
Fixes 242: set minReplicas to 3

avisiedo:
- set `minReplicas` to 3
- add new `replicas` field with the same value.
- add note to delete `minReplicas` when the new field is available in all the environments.


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-242

jlsherrill (approved):

PR #133, opened 2022-10-31
Fixes 235: add /ping for kafka consumer

jlsherrill:
and use a single binary for api and consumer


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-235

jlsherrill:
To test:

1.  You can now use 'make run' for both api & kafka
2. alternatively, you can run:   `go run cmd/content_sources/main.go api consumer`
3. i'd test those out individually as well:     `go run cmd/content_sources/main.go  api`  &    `go run cmd/content_sources/main.go  consumer`   (although you'll probably get an error trying to run them on the same machine, as they will both try to listen for /ping on the same port)
4.  verify that `curl localhost:8000/ping`  works  for all 3 versions of the command

pkg/event/consumer.go
```diff
@@ -104,8 +99,8 @@ func NewConsumerEventLoop(consumer *kafka.Consumer, handler Eventable) func() {
 					}
 
 					select {
-					case sig := <-sigchan:
-						log.Logger.Info().Msgf("Caught signal %v: terminating\n", sig)
+					case <-ctx.Done():
+						log.Logger.Info().Msgf("Caught signal, terminating consumer\n")
```
avisiedo, line 103:
```suggestion
						log.Logger.Info().Msgf("Context done for NewConsumerEventLoop")
```
I suggest to update the message to indicate the source is the context.

pkg/config/config.go
```diff
@@ -227,6 +232,17 @@ func ConfigureEcho() *echo.Echo {
...
+func ConfigureEchoShutdown(e *echo.Echo) {
+	quit := make(chan os.Signal, 1)
+	signal.Notify(quit, syscall.SIGINT, syscall.SIGTERM)
+	<-quit
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	if err := e.Shutdown(ctx); err != nil {
+		e.Logger.Fatal(err)
+	}
+}
```
avisiedo, line 248:
```suggestion
```
I think this function is not used; if that is the case I suggest to remove it.

avisiedo (changes requested):
:+1: for the changes on this pr; only a few changes requested.

thinks I have learned:
- I have learned and dig more about mechanisms to synchronize in golang and how to use the context to broadcast the termination.
- I have learned how to debug this stuffs from vscode in a productive way; the below is not in the pr, but I found very helpful when debugging the changes; I suggest to add the below to `.vscode/launch.json` for the service:

```raw
    "console": "integratedTerminal"
```

This open a terminal to launch the debug process, and it allows to invoke the signal handler when ctl+c is pressed into that terminal, sending a sigint to the process. This is important because from the integrated debugger panel in vscode only a sigkill is possible to be sent to the process (when we click stop, press the key-binding to stop debugger, or click the menu item to stop the debugger).

Furthermore; number of replicas for the kafka consumer verified in ephemeral environment:

```raw
$ oc get deployments
NAME                                     READY   UP-TO-DATE   AVAILABLE   AGE
content-sources-backend-kafka-consumer   3/3     3            3           4m30s
content-sources-backend-service          3/3     3            3           4m30s
```

And ping to kafka consumer verified locally and from the ephemeral environment.

jlsherrill:
@avisiedo   Updated with suggestions! thank you!

avisiedo:
lgtm

PR #138, opened 2022-11-03
Refs 264: Fix one missed case in the unit test

avisiedo:
- fix test "force error when message content fails validation", it was providing
  two errors which order is not deterministic; this change fix that providing only
  one failure, making the result deterministic.


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-264

jlsherrill (approved):

PR #139, opened 2022-11-03
Fixes 262: ignore metalinks in external repos download

jlsherrill:
image builder started supporting fedora but is using metalinks which we do not support.  This ignores those and also fixes a bug & deprecated method use.


jlsherrill:
so in the end, there was no new repositories added, this just fixes the download/save process

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-262

pkg/external_repos/image_builder.go
```diff
@@ -3,16 +3,16 @@ package external_repos
...
 
 // IBUrlsFromDir Import Repo Urls from image builder's git repo
 // The dirPath should point to GITDIR/distributions/ which contains
-// 		the repo directories.  Each directory contains a json file
-//		with the same name as the directory itself:
-//		./distributions/rhel-90/rhel-90.json
+//
+//	the repo directories.  Each directory contains a json file
+//	with the same name as the directory itself:
+//	./distributions/rhel-90/rhel-90.json
```
rverdile, line 15:
```suggestion
// the repo directories.  Each directory contains a json file
// with the same name as the directory itself:
// ./distributions/rhel-90/rhel-90.json
```

pkg/external_repos/image_builder.go
```diff
@@ -56,18 +56,22 @@ func extractUrls(filePath string) ([]string, error) {
...
 	}
 	return repoUrls, nil
 }
 
 // Scans the IB Distributions directory for repo json files
-//   This assumes that some directory "distributions/foo/" contains
-// 	 a json file  distributions/foo/foo.json
+//
+//	  This assumes that some directory "distributions/foo/" contains
+//		 a json file  distributions/foo/foo.json
```
rverdile, line 70:
```suggestion
// This assumes that some directory "distributions/foo/" contains
// a json file  distributions/foo/foo.json
```

rverdile (approved):
two small comments otherwise tested and lgtm!

pkg/external_repos/image_builder.go
jlsherrill: so go fmt is doing this.  I turned on 'run go fmt' automatically in my ide, and if you run `go fmt` against a file manually it seems to do this too.  Let me see if i can figure out the reasoning of why its doing this.

jlsherrill: ah, i see why!

PR #141, opened 2022-11-04
Fixes 261: add limits when returning rpm searches

avisiedo:
- This update the request structure to add limit field.
- By default it set the value of 100.
- The maximum threshold is 500.

TODO
- [x] Add unit test for the handler.


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-261

swadeley:
@avisiedo Please rebase

pkg/handler/repository_rpms_test.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+func (suite *RpmSuite) TearDownTest() {
+	//Rollback and reset db.DB
+	db.DB.Rollback()
+	db.DB = suite.savedDB
+	require.NoError(suite.T(), suite.echo.Shutdown(context.Background()))
+}
+
+func (suite *RpmSuite) TestRegisterRepositoryRpmRoutes() {
```
rverdile, line 81: is this already tested implicitly by `serveRpmsRouter`? I see that function calls `RegisterRepositoryRpmRoutes(pathPrefix, &rh.Dao)`

pkg/handler/repository_rpms_test.go
```diff
@@ -0,0 +1,330 @@
...
+	response := rr.Result()
+	defer response.Body.Close()
+
+	body, err := ioutil.ReadAll(response.Body)
+	return response.StatusCode, body, err
+}
+
+type RpmSuite struct {
+	suite.Suite
+	savedDB *gorm.DB
```
rverdile, line 60: I'm pretty sure this database stuff is unused, because we're mocking the dao layer, and can be removed. I see it's also in `repositories_test.go`, but that was added before we started mocking the dao layer.

Andrewgdewar (approved):
Functionally works really well, code wise, looks like @rverdile has a few comments. 
This gets an ack from though!

pkg/handler/repository_rpms_test.go
avisiedo: thanks mate! you are right that both are covering the code!

I used to add one function/method, one test for it; `serveRpmsRouter` is a helper to make the unit tests simpler; but the responsibility for `serveRpmsRouter` is not to cover `RegistryRepositoryRpmsRoutes`, even if it is covered indirectly; in this case it is true that it does not make any difference, but it allows to grow the unit test independent from `serveRpmsRouter` if more complex behavior is added to `RegistryRepositoryRpmsRoute`, and keep the responsibility for each function separated (helper and unit test).

I would leave it as it is because of the above vision, but if I am overlooking something, please let me know; I am human, I do mistakes (many), and I learn from them.

pkg/handler/repository_rpms_test.go
avisiedo: Nice catch! I didn't see at all! updating! :+1:

avisiedo:
thanks mates for the review! @rverdile I have added some comments; If I am overlooking something please let me know, happy to update the pr if needed.

I will create a new card to cover the unit test for the second handler `listRepositoriesRpm` (or if there some card related with it, it could be an opportunity to add it), so the coverage is increased.

Sorry for not providing the unit tests for the handlers when I created this code, at that moment I didn't know very well how to add them; now I am more capable to add unit tests for the http handlers.

pkg/handler/repository_rpms_test.go
rverdile: I guess my instinct would be to add a unit test for `RegistryRepositoryRpmsRoutes` if/when it becomes more complex. But I understand what you're saying, I'm okay leaving it like this.

pkg/test/handler/utils.go
```diff
@@ -0,0 +1,30 @@
+package handler
```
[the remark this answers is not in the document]
rverdile: I agree that `handlertest` is redundant. My thinking was these two packages go together and `handlertest` gives immediate context to what the package is for (testing handler). And since we're using an alias anyway, it felt to me like there was an unnecessary package name collision.

But we can leave this for now, maybe I'll think of a different name change later.

rverdile:
@avisiedo addressed all my concerns :)

PR #142, opened 2022-11-08
yummy refactor

rverdile:
Companion to this PR: https://github.com/content-services/yummy/pull/7

Moves the functionality in `dao/external_resources.go` to yummy. Refactors to adjust to yummy changes. 

I haven't really tested this a lot yet, so it might be broken. Feedback welcome, architectural feedback in particular.

TODO

- [ ] Fix tests.


pkg/dao/repository_config.go
```diff
@@ -393,9 +394,11 @@ func (r repositoryConfigDaoImpl) ValidateParameters(orgId string, params api.Rep
 		if err != nil {
 			return response, err
 		} else if response.URL.Valid {
-			repomd := r.validateMetadataPresence(url, &response)
+			repo := r.yumRepo.Repository()
+			repo.URL = url
+			r.validateMetadataPresence(&response)
 			if response.URL.MetadataPresent {
-				r.checkSignaturePresent(&params, repomd, &response)
+				r.checkSignaturePresent(&params, &response)
```
rverdile, line 401: For the purposes of mocking, it seems I have to dependency inject the YumRepository interface and also return the Repository struct with the `Repository()` method. 

This seems more cumbersome than what we had before, so open to other ideas.

pkg/dao/repository_config_test.go
```diff
@@ -996,8 +997,9 @@ func (suite *RepositoryConfigSuite) TestValidateParameters() {
 		UUID: &repoConfig.UUID,
 	}
 
-	mockExtRDao.Mock.On("FetchRepoMd", repoConfig.Repository.URL).Return(pointy.String("<XMLFILE>"), 200, nil)
-	mockExtRDao.Mock.On("FetchSignature", repoConfig.Repository.URL).Return(pointy.String("sig"), 200, nil)
+	mockExtRDao.Mock.On("FetchRepomd").Return(200, nil)
```
rverdile, line 1000: Only updated this one test as a demonstration for now because I'm not confident about how this is architect-ed.

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-200

pkg/dao/repository_config.go
rverdile: Resolved this with yumRepo.Configure(settings yum.YummySettings) method.

pkg/dao/external_resource.go
```diff
@@ -1,3 +1,5 @@
+//TODO remove this file. Everything here handled by yummy
+
```
jlsherrill, line 2: can probably go ahead and do this

pkg/dao/repository_config_test.go
```diff
@@ -996,8 +997,8 @@ func (suite *RepositoryConfigSuite) TestValidateParameters() {
 		UUID: &repoConfig.UUID,
 	}
 
-	mockExtRDao.Mock.On("FetchRepoMd", repoConfig.Repository.URL).Return(pointy.String("<XMLFILE>"), 200, nil)
-	mockExtRDao.Mock.On("FetchSignature", repoConfig.Repository.URL).Return(pointy.String("sig"), 200, nil)
+	mockExtRDao.Mock.On("Repomd").Return(&yum.Repomd{}, 200, nil)
```
jlsherrill, line 1000: I'd probably rename mockExtRDao to mockYumRepo or something

pkg/dao/repository_config_test.go
```diff
@@ -1135,16 +1136,17 @@ func (suite *RepositoryConfigSuite) TestValidateParametersGpgKey() {
...
 
-	mockExtRDao.Mock.On("FetchRepoMd", *parameters.URL).Return(pointy.String(test.SignedRepomd()), 200, nil)
-	mockExtRDao.Mock.On("FetchSignature", *parameters.URL).Return(pointy.String(test.RepomdSignature()), 200, nil)
+	mockExtRDao.Mock.On("Repomd").Return(test.Repomd, 200, nil)
+	mockExtRDao.Mock.On("Signature").Return(test.RepomdSignature(), 200, nil)
 
 	response, err := dao.ValidateParameters(repoConfig.OrgID, parameters, []string{})
 	assert.NoError(t, err)
 	assert.True(t, response.GPGKey.Valid)
+	test.SignedRepomd()
```
jlsherrill, line 1149: is this doing anything?

jlsherrill:
Overall, a big +1, i don't have any major comments

jlsherrill (approved):
ACK, we'll just need to tag yummy and then update the go.mod here

rverdile:
not sure if CI failure here is related

rverdile:
Current test failure is because the `yum.YumRepository` object in our backend is instantiated when the handler is registered.  This means that repeated calls to yummy do not fetch new data because the handler is registered only when the app first loads.

Opened this PR to yummy: https://github.com/content-services/yummy/pull/8, so that we have a method to clear the data.

pkg/dao/repository_configs.go
```diff
@@ -393,10 +395,13 @@ func (r repositoryConfigDaoImpl) ValidateParameters(orgId string, params api.Rep
 		err = r.validateUrl(orgId, url, &response, excludedUUIDS)
 		if err != nil {
 			return response, err
-		} else if response.URL.Valid {
-			repomd := r.validateMetadataPresence(url, &response)
+		}
+		if response.URL.Valid {
+			r.yumRepo.Clear()
```
jlsherrill, line 400: This method gets called via a go routine in the handler, which likely means if you pass in 4 repositories to validate, it uses the same instance of the dao and the same instance of yumRepo?   If i'm right, i wonder if its causing any problems?

rverdile: I hadn't considered that. I tried validating with multiple repos with different URLs. It seems to work but I'm kinda surprised that it does.

jlsherrill (approved):

PR #144, opened 2022-11-09
Fixes 266: use gorm config for batch creation

jlsherrill:
Gorm has a built in way to batch creations, but we were not using it.  By using it, we can handle all creations, and introspect larger repos that were failing before

### DIFF
--- a/pkg/dao/rpm.go
+++ b/pkg/dao/rpm.go
@@ -5,7 +5,6 @@ import (
 	"strings"
 
 	"github.com/content-services/content-sources-backend/pkg/api"
-	"github.com/content-services/content-sources-backend/pkg/config"
 	"github.com/content-services/content-sources-backend/pkg/models"
 	"github.com/content-services/yummy/pkg/yum"
 	"gorm.io/gorm"
@@ -13,8 +12,7 @@ import (
 )
 
 type rpmDaoImpl struct {
-	db                   *gorm.DB
-	pagedRpmInsertsLimit int
+	db *gorm.DB
 }
 
 type RpmDaoOptions struct {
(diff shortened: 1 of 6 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-266

jlsherrill:
To test:
Create a repo with this url:  https://jlsherrill.fedorapeople.org/fake-repos/rhel7/
wait a minute or two, verify that the repo is introspected and shows packages.

jlsherrill:
rebased!

pkg/dao/rpm.go
```diff
@@ -5,44 +5,24 @@ import (
...
-		if value > 0 {
-			pagedRpmInsertsLimit = value
-		}
-	}
-
 	// Return DAO instance
 	return rpmDaoImpl{
-		db:                   db,
-		pagedRpmInsertsLimit: pagedRpmInsertsLimit,
+		db: db,
```
rverdile, line 25: I would remove the options parameter since `GetRpmDao` is no longer using it

pkg/handler/api.go
```diff
@@ -74,9 +72,7 @@ func RegisterRoutes(engine *echo.Echo) {
 		RegisterRepositoryRoutes(group, &daoRepo, &introspectRequest)
 		RegisterRepositoryParameterRoutes(group, &daoRepo, &externalRepo)
 
-		daoRpm := dao.GetRpmDao(db.DB, &dao.RpmDaoOptions{
-			PagedRpmInsertsLimit: pointy.Int(pagedRpmInsertsLimit),
-		})
+		daoRpm := dao.GetRpmDao(db.DB, &dao.RpmDaoOptions{})
```
rverdile, line 75: As far as I can tell, RpmDaoOptions is no longer being used anywhere, so I think that can be removed from here and wherever else it still remains,

PR #145, opened 2022-11-14
Revert "Fixes 235: add /ping for kafka consumer (#133)"

jlsherrill:


pkg/config/config.go
```diff
@@ -83,7 +83,7 @@ func readConfigFile(v *viper.Viper) {
 
 func setDefaults(v *viper.Viper) {
 	v.SetDefault("Loaded", true)
-	// In viper you have to set defaults, otherwise loading from ENV doesn't work
+	//In viper you have to set defaults, otherwise loading from ENV doesn't work
```
mshriver, line 86: Don't think removing this whitespace is intentional, but is inconsequential.

jlsherrill: this is just a straight revert of the previous change,  if we can figure out if this is breaking stage, we'll put it back

mshriver (approved):
acking as a straight revert, not a mechanism QE was relying on.

PR #148, opened 2022-11-24
Fixes 276: Add "Popular repositories" endpoint

Andrewgdewar:
Adds the "/common_repositories/" endpoint with 4 repositories in the common_repositories.cfg for now.

- [x] Still needs tests

### DIFF
--- a/api/docs.go
+++ b/api/docs.go
@@ -20,6 +20,49 @@ const docTemplate = `{
     "host": "{{.Host}}",
     "basePath": "{{.BasePath}}",
     "paths": {
+        "/popular_repositories/": {
+            "get": {
+                "description": "Get popular repositories",
+                "consumes": [
+                    "application/json"
+                ],
+                "produces": [
+                    "application/json"
+                ],
+                "tags": [
+                    "popular",
+                    "repositories"
+                ],
+                "summary": "List Popular Repositories",
+                "operationId": "listPopularRepositories",
+                "responses": {
+                    "200": {
+                        "description": "OK",
+                        "schema": {
+                            "$ref": "#/definitions/api.PopularRepositoriesCollectionResponse"
+                        }
+                    },
+                    "400": {
+                        "description": "Bad Request",
+                        "schema": {
+                            "$ref": "#/definitions/errors.ErrorResponse"
+                        }
+                    },
+                    "404": {
+                        "description": "Not Found",
+                        "schema": {
+                            "$ref": "#/definitions/errors.ErrorResponse"
+                        }
+                    },
+                    "500": {
+                        "description": "Internal Server Error",
+                        "schema": {
+                            "$ref": "#/definitions/errors.ErrorResponse"
+                        }
+                    }
+                }
+            }
+        },
         "/repositories/": {
             "get": {
                 "description": "list repositories",
@@ -761,6 +804,70 @@ const docTemplate = `{
                 },
                 "prev": {
                     "description": "Path to previous page of results",
+                    "type": "string"
+                }
+            }
+        },
+        "api.PopularRepositoriesCollectionResponse": {
+            "type": "object",
+            "properties": {
+                "data": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/api.PopularRepositoryResponse"
+                    }
+                },
+                "links": {
+                    "description": "Links to other pages of results",
+                    "$ref": "#/definitions/api.Links"
+                },
+                "meta": {
+                    "description": "Metadata about the request",
+                    "$ref": "#/definitions/api.ResponseMetadata"
+                }
+            }
+        },
+        "api.PopularRepositoryResponse": {
+            "type": "object",
+            "properties": {
+                "distribution_arch": {
+                    "description": "Architecture to restrict client usage to",
+                    "type": "string",
+                    "example": "x86_64"
+                },
+                "distribution_versions": {
+                    "description": "Versions to restrict client usage to",
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    },
+                    "example": [
+                        "7",
+                        "8"
+                    ]
+                },
+                "existing_name": {
+                    "description": "Existing reference name for repository",
+                    "type": "string"
+                },
+                "gpg_key": {
+                    "description": "GPG key for repository",
+                    "type": "string"
+                },
+                "metadata_verification": {
+                    "description": "Verify packages",
+                    "type": "boolean"
+                },
+                "suggested_name": {
+                    "description": "Suggested name of the popular repository",
+                    "type": "string"
+                },
+                "url": {
+                    "description": "URL of the remote yum repository",
+                    "type": "string"
+                },
+                "uuid": {
+                    "description": "UUID of the repository if it exists for the user",
                     "type": "string"
                 }
             }
--- a/pkg/api/popular_repositories.go
+++ b/pkg/api/popular_repositories.go
@@ -1,0 +1,25 @@
+package api
+
+// RepositoryParameterResponse holds data returned by a repositories API response
+
+type PopularRepositoryResponse struct {
+	UUID                 string   `json:"uuid"`                                // UUID of the repository if it exists for the user
+	ExistingName         string   `json:"existing_name"`                       // Existing reference name for repository
+	SuggestedName        string   `json:"suggested_name"`                      // Suggested name of the popular repository
+	URL                  string   `json:"url"`                                 // URL of the remote yum repository
+	DistributionVersions []string `json:"distribution_versions" example:"7,8"` // Versions to restrict client usage to
+	DistributionArch     string   `json:"distribution_arch" example:"x86_64"`  // Architecture to restrict client usage to
+	GpgKey               string   `json:"gpg_key"`                             // GPG key for repository
+	MetadataVerification bool     `json:"metadata_verification"`               // Verify packages
+}
+
+type PopularRepositoriesCollectionResponse struct {
+	Data  []PopularRepositoryResponse `json:"data"`  //
+	Meta  ResponseMetadata            `json:"meta"`  // Metadata about the request
+	Links Links                       `json:"links"` // Links to other pages of results
+}
+
+func (r *PopularRepositoriesCollectionResponse) SetMetadata(meta ResponseMetadata, links Links) {
+	r.Meta = meta
+	r.Links = links
+}
--- a/pkg/external_repos/introspect_test.go
+++ b/pkg/external_repos/introspect_test.go
@@ -3,7 +3,7 @@ package external_repos
 //nolint:gci
 import (
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"testing"
@@ -46,7 +46,7 @@ func TestIntrospect(t *testing.T) {
 				if response, err = http.DefaultClient.Get(url); err != nil {
 					t.Errorf(err.Error())
 				}
-				if body, err = ioutil.ReadAll(response.Body); err != nil {
+				if body, err = io.ReadAll(response.Body); err != nil {
 					t.Errorf(err.Error())
 				}
 				response.Body.Close()
--- a/pkg/handler/popular_repositories.go
+++ b/pkg/handler/popular_repositories.go
@@ -1,0 +1,99 @@
+package handler
+
+import (
+	"embed"
+	"encoding/json"
+	"strings"
+
+	"github.com/content-services/content-sources-backend/pkg/api"
+	"github.com/content-services/content-sources-backend/pkg/dao"
+	ce "github.com/content-services/content-sources-backend/pkg/errors"
+	"github.com/labstack/echo/v4"
+)
+
+//go:embed "popular_repositories.json"
+
+var fs embed.FS
+
+type PopularRepositoriesHandler struct {
+	RepositoryDao dao.RepositoryConfigDao
+}
+
+func RegisterPopularRepositoriesRoutes(
+	engine *echo.Group,
+	dao *dao.RepositoryConfigDao,
+) {
+	rph := PopularRepositoriesHandler{RepositoryDao: *dao}
+	engine.GET("/popular_repositories/", rph.listPopularRepositories)
+}
+
+// ListPopularRepositories godoc
+// @Summary      List Popular Repositories
+// @ID           listPopularRepositories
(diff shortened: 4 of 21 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-276

avisiedo:
I suggest to run 'make openapi' to re-generate the `api/openapi.json` and `api/docs.go` files.

pkg/handler/common_repositories.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	if repos.Data[0].Name != "" {
+		repo.ExistingName = repos.Data[0].Name
+	}
+
+	return repo, nil
+}
+
+var CommonRepositories = [...]api.CommonRepositoryResponse{
```
avisiedo: I suggest to add the additional repositories.

pkg/handler/common_repositories.go
```diff
@@ -0,0 +1,99 @@
...
+// ListCommonRepositories godoc
+// @Summary      List Common Repositories
+// @ID           listCommonRepositories
+// @Description  get common repositories
+// @Tags         common,repositories
+// @Accept       json
+// @Produce      json
+// @Success      200 {object} api.CommonRepositoriesCollectionResponse
+// @Router       /common_repositories/ [get]
+func (rh *CommonRepositoriesHandler) listCommonRepositories(c echo.Context) error {
```
avisiedo, line 34: I suggest to add a unit test to provide coverage to the new handler.

pkg/handler/common_repositories.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (rh *CommonRepositoriesHandler) updateNameIfExists(c echo.Context, repo api.CommonRepositoryResponse) (api.CommonRepositoryResponse, error) {
+	_, orgID := getAccountIdOrgId(c)
+	repos, _, err := rh.RepositoryDao.List(orgID, api.PaginationData{}, api.FilterData{Search: repo.URL})
+
+	if err != nil {
+		return api.CommonRepositoryResponse{}, echo.NewHTTPError(httpCodeForError(err), "Error listing repositories: "+err.Error())
+	}
+
+	if repos.Data[0].Name != "" {
```
avisiedo: I suggest to check `len(repos.Data)` before using the slice to prevent panic.

pkg/external_repos/introspect_test.go
```diff
@@ -25,7 +25,7 @@ func TestIsRedHatUrl(t *testing.T) {
 }
 
 // https://packages.cloud.google.com/yum/repos/google-compute-engine-el8-x86_64-stable/repodata/repomd.xml
-//go:embed "test_files/test_repomd.xml"
+// go:embed "test_files/test_repomd.xml"
```
avisiedo, line 28: I think this space is evoking the current error into the tests: https://github.com/content-services/content-sources-backend/actions/runs/3543726621/jobs/5950420530#step:5:117

avisiedo (changes requested):
Thanks for the changes :+1: a few suggestions in the pr review.

I could not replay the linter hints from the macos environment; I will try in x86 environment to see if I get something different; very likely I have something wrong in my environment.

If the data is going to be totally static, maybe embed the content in a variable from an external yaml file could be nice (https://pkg.go.dev/embed); on this point, in regarding to the GpgKey field, it could be easier to write the gpg key in a yaml file than in a json file (this reference could be useful: https://yaml-multiline.info)

pkg/external_repos/introspect_test.go
Andrewgdewar: Yeah, I noticed this.. Fixed

pkg/handler/common_repositories.go
Andrewgdewar: I know, waiting on Justin's feedback. Then will add tests.

pkg/handler/common_repositories.go
```diff
@@ -0,0 +1,74 @@
...
+
+// ListCommonRepositories godoc
+// @Summary      List Common Repositories
+// @ID           listCommonRepositories
+// @Description  get common repositories
+// @Tags         common,repositories
+// @Accept       json
+// @Produce      json
+// @Success      200 {object} api.CommonRepositoriesCollectionResponse
+// @Router       /common_repositories/ [get]
```
jlsherrill, line 33: You'll want to detail the errors as well: like this:   https://github.com/content-services/content-sources-backend/blob/main/pkg/handler/repositories.go#L115-L117

pkg/config/common_repositories.json
```diff
@@ -0,0 +1,50 @@
...
+        "metadata_verification": false
+    }
+]
```
jlsherrill, line 50: uuid & existing_name aren't part of this structure, as they'd be generated at run time, are they needed in the json?

pkg/handler/common_repositories.go
```diff
@@ -0,0 +1,74 @@
...
+func (rh *CommonRepositoriesHandler) listCommonRepositories(c echo.Context) error {
+	jsonConfig, err := os.ReadFile("./pkg/config/common_repositories.json")
```
jlsherrill, line 35: I don't think this will work within the container without actually copying the .json to  /pkg/config/common_repositories.json

Usually you'd have to embed like this:  https://github.com/content-services/content-sources-backend/blob/be2bf7c4d0f4588b6f6e028bf6b78a89ccd78d1a/pkg/external_repos/store.go#L11-L14

and then read it from there:  https://github.com/content-services/content-sources-backend/blob/be2bf7c4d0f4588b6f6e028bf6b78a89ccd78d1a/pkg/external_repos/store.go#L42-L58

pkg/handler/common_repositories.go
```diff
@@ -0,0 +1,74 @@
...
+		return ce.NewErrorResponseFromError("listCommonRepositories could not Get RepositoryDao.List", err)
+	}
+
+	// If the URL exists update the "existingName" field
+	if len(repos.Data) > 0 && repos.Data[0].Name != "" {
+		repo.ExistingName = repos.Data[0].Name
+		repo.UUID = repos.Data[0].UUID
+	}
+
+	return nil
```
jlsherrill, line 73: a much more efficient way to do this, would be to do 1 query for all urls, and the collate the lists together.  Obviously for 4 repos it doesn't matter too much, but you might at least leave a comment hinting at that solution

pkg/handler/common_repositories.go
```diff
@@ -0,0 +1,74 @@
...
+
+		if err != nil {
+			return ce.NewErrorResponseFromError("listCommonRepositories could not Get RepositoryDao.List", err)
+		}
+	}
+
+	return c.JSON(200, setCollectionResponseMetadata(&api.CommonRepositoriesCollectionResponse{Data: configData}, c, int64(len(configData))))
+}
+
+func (rh *CommonRepositoriesHandler) updateDataIfExists(c echo.Context, repo *api.CommonRepositoryResponse) error {
```
jlsherrill, line 59: nit pick ,but i think data is a little verbose, maybe just 'updateIfExists' ?

pkg/handler/common_repositories.go
```diff
@@ -0,0 +1,74 @@
...
+
+	for i := 0; i < len(configData); i++ {
+		err := rh.updateDataIfExists(c, &configData[i])
+
+		if err != nil {
+			return ce.NewErrorResponseFromError("listCommonRepositories could not Get RepositoryDao.List", err)
+		}
+	}
+
+	return c.JSON(200, setCollectionResponseMetadata(&api.CommonRepositoriesCollectionResponse{Data: configData}, c, int64(len(configData))))
```
jlsherrill, line 56: +1 to returning the full list, while leaving the possibility for pagination in the future!

jlsherrill:
overall this is looking great!

pkg/handler/common_repositories.go
```diff
@@ -0,0 +1,74 @@
...
+// @Tags         common,repositories
+// @Accept       json
+// @Produce      json
+// @Success      200 {object} api.CommonRepositoriesCollectionResponse
+// @Router       /common_repositories/ [get]
+func (rh *CommonRepositoriesHandler) listCommonRepositories(c echo.Context) error {
+	jsonConfig, err := os.ReadFile("./pkg/config/common_repositories.json")
+
+	if err != nil {
+		return ce.NewErrorResponseFromError("listCommonRepositories could not read common_repositories.json", err)
```
rverdile, line 38:
```suggestion
		return ce.NewErrorResponseFromError("could not read common_repositories.json", err)
```
since these error messages are potentially ones users would see, is it better to remove references to function names?

pkg/handler/common_repositories.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	configData := []api.CommonRepositoryResponse{}
+
+	err = json.Unmarshal([]byte(jsonConfig), &configData)
+	if err != nil {
+		return ce.NewErrorResponseFromError("listCommonRepositories could not read common_repositories.json", err)
```
rverdile, line 45:
```suggestion
		return ce.NewErrorResponseFromError("could not read common_repositories.json", err)
```

pkg/handler/common_repositories.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	for i := 0; i < len(configData); i++ {
+		err := rh.updateDataIfExists(c, &configData[i])
+
+		if err != nil {
+			return ce.NewErrorResponseFromError("listCommonRepositories could not Get RepositoryDao.List", err)
```
rverdile, line 52:
```suggestion
			return ce.NewErrorResponseFromError("could not get repository list", err)
```

pkg/handler/common_repositories.go
Andrewgdewar: Added this, but not sure if I captured the correct errors (I basically did a straight up copy).

pkg/handler/common_repositories.go
Andrewgdewar: I could leave a comment or write a separate db query. Which would you prefer here?

pkg/handler/popular_repositories.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	configData := []api.PublicRepositoryResponse{}
+
+	err = json.Unmarshal([]byte(jsonConfig), &configData)
+	if err != nil {
+		return ce.NewErrorResponseFromError("Could not read popular_repositories.json", err)
+	}
+
+	// We should likely call the db directly here to reduce this down to one query if this list get's larger.
```
Andrewgdewar, line 55: I've added this comment for now.

pkg/handler/popular_repositories.json
```diff
@@ -0,0 +1,42 @@
+[
+    {
+        "suggested_name": "Epel-9-Everything-x86_64",
```
jlsherrill, line 3: I'd probably change these names to   "EPEL 9 Everything x86_64"

I do wonder if we should go with the full "Extra Packages for Enterprise Linux 9 Everything x86_64", but that seems too long :)

pkg/handler/popular_repositories.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/labstack/echo/v4"
+)
+
+//go:embed "popular_repositories.json"
+
+var fs embed.FS
+
+type PublicRepositoriesHandler struct {
+	RepositoryDao dao.RepositoryConfigDao
+}
```
jlsherrill, line 20: Should this be "Popular" and not "Public"?

pkg/api/popular_repositories.go
```diff
@@ -0,0 +1,25 @@
...
+type PublicRepositoriesCollectionResponse struct {
+	Data  []PublicRepositoryResponse `json:"data"`  //
+	Meta  ResponseMetadata           `json:"meta"`  // Metadata about the request
+	Links Links                      `json:"links"` // Links to other pages of results
+}
+
+func (r *PublicRepositoriesCollectionResponse) SetMetadata(meta ResponseMetadata, links Links) {
+	r.Meta = meta
+	r.Links = links
+}
```
jlsherrill, line 25: same here, shouldn't this be PopularRepositories?

pkg/handler/popular_repositories.json
Andrewgdewar: Added the first for now.

api/docs.go
```diff
@@ -20,6 +20,49 @@ const docTemplate = `{
     "host": "{{.Host}}",
     "basePath": "{{.BasePath}}",
     "paths": {
+        "/popular_repositories/": {
+            "get": {
+                "description": "get public repositories",
```
swadeley, line 25:
```suggestion
                "description": "Get public repositories",
```
Most other descriptions start with a capital letter

pkg/handler/popular_repositories_test.go
```diff
@@ -0,0 +1,136 @@
...
+	err = json.Unmarshal(body, &response)
+
+	assert.Nil(t, err)
+	assert.Equal(t, http.StatusOK, code)
+	mockDao.AssertExpectations(t)
+	assert.Equal(t, 0, response.Meta.Offset)
+	assert.Equal(t, int64(4), response.Meta.Count)
+	assert.Equal(t, 10, response.Meta.Limit)
+	assert.Equal(t, 4, len(response.Data))
+	assert.Equal(t, response.Data[0].ExistingName, "")
```
jlsherrill, line 79: it'd be good to check the condition where it does find a match.  So for example you could modify this test such that the last List() call actually returns a repo, and then check that its ExistingName matches what we'd expect (non-blank string)

jlsherrill:
Oh and it'd probably help qe to write a few curl commands to test

jlsherrill:
tested everything else its good to go! just one test addition request

pkg/handler/popular_repositories_test.go
Andrewgdewar: roger!

jlsherrill (approved):

api/docs.go
```diff
@@ -20,6 +20,49 @@ const docTemplate = `{
...
+                        "schema": {
+                            "$ref": "#/definitions/errors.ErrorResponse"
+                        }
+                    }
+                }
+            }
+        },
         "/repositories/": {
             "get": {
                 "description": "list repositories",
```
mshriver, line 68: I'd get more specific here, now that there are specific repository types.  Rhetorically, does this include popular repositories?

Not sure about the value of going all the way to rename this custom_repositories, but at least this description should indicate a distinction.

PR #149, opened 2022-11-24
Fixes 470: Add RBAC integration

avisiedo:
- Add Rbac middleware.
- Move middlewares to `pkg/middleware`
- Generate mock for the rbac client.
- Add initial unit test structure.
- Add configurations manage when enable middleware, base url and timeout.
- Refactor ConfigureEcho to the main package to avoid cycle depencies; not happy with this change, open to alternatives.
- add mock service to simulate clients response; this needs a re-thinking.

TODO

- [x] add use cases to cover all the paths for the rbac middleware and helper functions.
- [ ] refine the mock service client at least for rbac service.
- [x] compare with the results from ephemeral environment and update according to them.
- [x] Update deployment.yaml descriptor to consider CLIENTS_RBAC_BASE_URL


cmd/mock/rbac.go
```diff
@@ -0,0 +1,95 @@
...
+	            "key": "cost-management.aws.account",
+	            "operation": "equal",
+	            "value": "123456"
+	          }
+	        }
+	      ]
+	    }
+	  ]
+	}
+*/
```
avisiedo, line 40: Clean-up this block

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-240

cmd/mock/main.go
```diff
@@ -0,0 +1,59 @@
...
+		if err := e.Shutdown(ctx); err != nil {
+			log.Fatal().Msgf("error shutting down mock service: %s", err.Error())
+		}
+		cancel()
+	}()
+
+	<-quit
+	cancel()
+	wg.Wait()
+}
```
jlsherrill, line 59: instead of adding a new command, what about adding it to the content-sources command?  

`go run ./cmd/content-sources/main.go  api consumer rbac`  ?

cmd/mock/rbac.go
```diff
@@ -0,0 +1,136 @@
...
+
+	switch {
+	case xRhIdentity.OrgID == "12345" && xRhIdentity.AccountNumber == "11111":
+		return c.JSON(http.StatusOK, outputAdmin)
+	case xRhIdentity.OrgID == "12345" && xRhIdentity.AccountNumber == "22222":
+		return c.JSON(http.StatusOK, outputDefault)
+	default:
+		return c.JSON(http.StatusOK, outputEmpty)
+	}
+}
```
jlsherrill, line 136: This might be a hot take, but what do you think about moving a lot of this rbac code to a different github project?   

content-services/mock-rbac  ?

jlsherrill: and we could still spin up the server from our existing  content-sources/main.go  command, it would just that as a dependency

pkg/middleware/rbac.go
```diff
@@ -0,0 +1,103 @@
...
+		return client.RbacVerbWrite
+	case echo.PUT:
+		return client.RbacVerbWrite
+	case echo.DELETE:
+		return client.RbacVerbWrite
+
+	default:
+		return client.RbacVerbUndefined
+	}
+}
```
jlsherrill, line 83: i think our routes don't fully match this, for example:
```
    "method": "POST",
    "path": "/api/content-sources/v1/rpms/names",
```
would be associated with a Read permission.  It'd be nice if we had some method that could register the route alongside the permission:

```
	engine.GET("/repositories/", rh.listRepositories, rbac.RbacVerbRead)
	engine.GET("/repositories/:uuid", rh.fetch, rbac.RbacVerbRead)
	engine.PUT("/repositories/:uuid", rh.fullUpdate, rbac.RbacVerbWrite)
        etc...
```

cmd/mock/rbac.go
jlsherrill: i guess this is small enough that it doesn't really matter, i  retract this unless you think its a great idea

cmd/mock/rbac.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	switch {
+	case xRhIdentity.OrgID == "12345" && xRhIdentity.AccountNumber == "11111":
+		return c.JSON(http.StatusOK, outputAdmin)
+	case xRhIdentity.OrgID == "12345" && xRhIdentity.AccountNumber == "22222":
+		return c.JSON(http.StatusOK, outputDefault)
+	default:
+		return c.JSON(http.StatusOK, outputEmpty)
+	}
```
jlsherrill, line 135: could you base this off orgId instead of account number?    I'd probably suggest allow all for most all OrgIds, but have some more restricted ones for testing:  

```
switch {
	case xRhIdentity.OrgID == "12345" && xRhIdentity.AccountNumber == "readonly":
		return c.JSON(http.StatusOK, outputDefault)
	case xRhIdentity.OrgID == "12345" && xRhIdentity.AccountNumber == "noperms":
		return c.JSON(http.StatusOK, outputEmpty)
	default:
		return c.JSON(http.StatusOK, outputAdmin)
	}
```

I think this will make things simpler to develop with (especially when working in the UI where you don't control the orgId).  (also i get that orgId is typically a number, but i think it'd work as a string?)

cmd/mock/rbac.go
avisiedo: happy to move it into a new repo! I like it!

pkg/middleware/rbac.go
avisiedo: thanks! inspiring! I have a solution in mind that would works!

cmd/mock/rbac.go
avisiedo: sure! thanks for that! I didn't see from that point of view!

pkg/middleware/rbac.go
avisiedo: I have added a mechanism to map the api requests to the permissions.
- I create a new type which is a map[string]map[string]Permission.
- First map map the request method.
- Second map map the api path once its prefix is removed.
- The value stored is a struct for the permission name and verb for that http operation.

By default, if no mapping is found, it is rejected with unauthorized.

If a mapping is found, then the rbac client is called with the permission to check for the operation, and reject or continuous with the operation depending on the result.

deployments/deployment.yaml
```diff
@@ -1,6 +1,6 @@
 # https://consoledot.pages.redhat.com/clowder/dev/api_reference.html
 ---
-apiVersion: v1
+apiVersion: template.openshift.io/v1
```
avisiedo, line 3: I got the error message below when checking the template with `oc process`:

```sh
oc process -f deployments/deployment.yaml -p "IMAGE=quay.io/avisied0/content-sources-backend"        -p "IMAGE_TAG=somehash" -n "$( oc project -q )" -o yaml -p "ENV_NAME=env-$(oc project -q )"
error: failed to read input object (not a Template?): resource mapping not found for name: "content-sources-backend" namespace: "" from "deployments/deployment.yaml": no matches for kind "Template" in version "v1"
```

cmd/mock/main.go
avisiedo: thanks updated! if the configuration is setting `clients.rbac_enabled: false` then the rbac mock is not started even if it is indicated into the arguments.

mk/mock.mk
```diff
@@ -0,0 +1,24 @@
...
+.PHONY: mock-down
+mock-down: ## Stop mock services
+	[ -e "$(MOCK_PID)" ] || ( echo "mock service is not up currently"; exit 1 )
+	/usr/bin/kill -SIGTERM "$$(<"$(MOCK_PID)")"
+	[ ! -e "$(MOCK_PID)" ] || rm -f "$(MOCK_PID)"
+
+.PHONY: mock-clean
+mock-clean:  ## Clean the mock service
+	[ ! -e "$(MOCK_PID)" ] || rm -f "$(MOCK_PID)"
+	[ ! -e "$(MOCK_LOG)" ] || rm -f "$(MOCK_LOG)"
```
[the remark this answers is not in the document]
avisiedo: thanks updated! I have removed this makefile rules.

avisiedo:
**test specs**

```raw
Scenario 1: check write actions for admin user

Given an admin user
When it create, update or delete a repository
Then the operation is successful

----

Scenario 2: check read actions for admin user

Given an admin user
When it list or get a repository
Then the operation is successful

----

Scenario 3: check write rejected for normal user

Given a normal user (no write permissions)
When it create, update or delete a repository
Then the operation is unauthorized

----

Scenario 4: check write rejected for user with no read nor write permissions

Given a user with no read or write permissions
When it create, update, delete, list or get a repository
Then the operation is unauthorized

----

Scenario 5: check read for normal user

Given a normal user (no write permissions)
When it list or retrieve a repository
Then the operation is authorized

```

cmd/mock/rbac.go
avisiedo: sorry, I didn't accommodate this change; the "mock rback service" is launched when a "mock_rbac" is indicated into the argument list of the main binary. I will create a ticket for future considerations.

README.md
```diff
@@ -134,6 +134,27 @@ To open the prometheus web UI, once the container is up, run the below:
...
+  clients:
+    rbac_enabled: True
+    rbac_base_url: http://localhost:8800/api/rbac/v1
+    rbac_timeout: 30
+  mocks:
+    my_org_id: "12345"
+    rbac:
+      account_admin: "12345"
+      account_viewer: "123456"
+  ```
```
jlsherrill, line 151: only having a single org have access might be a tad annoying when development.  What about adding some more option:

```
mocks:
  rbac:
    denied_org_ids: ["1","2","3"]
    viewer_org_ids: ["4", "5"]
```

and then the logic could be:

1) if in denied org_ids list, send empty list of permission
2) if in viewer_org_ids, only allow for view only
3) else allow all.

pkg/test/mocks/rbac/rbac.go
```diff
@@ -0,0 +1,181 @@
...
+		log.Debug().Msgf("returning permissions for admin")
+		return c.JSON(http.StatusOK, outputAdmin)
+	case xRhIdentity.Identity.AccountNumber == accountViewer:
+		log.Debug().Msgf("returning permissions for a viewer")
+		return c.JSON(http.StatusOK, outputDefault)
+	default:
+		log.Debug().Msgf("returning empty permissions")
+		return c.JSON(http.StatusOK, outputEmpty)
+	}
+}
```
jlsherrill, line 181: and i would avoid using account numbers, just stick to org ids

README.md
jlsherrill: and you could set a default in the config.yaml.exmple of 

    denied_org_ids: ["denied"]
    viewer_org_ids: ["view"]

pkg/router/route.go
```diff
@@ -0,0 +1,66 @@
+package router
```
jlsherrill, line 1: i kinda question if this needs to be its own package versus another file in config.  What drew you to make it a package?

pkg/middleware/rbac_test.go
```diff
@@ -0,0 +1,554 @@
...
+	})
+	require.Panics(t, func() {
+		NewRbac(Rbac{
+			BaseUrl:        "http://localhost:8800/api/rbac/v1",
+			Skipper:        skipper,
+			PermissionsMap: pm,
+			Client:         nil,
+		})
+	})
+}
```
jlsherrill, line 129: Could we either add a startup check or a unit test that checks that all registered routes have something defined in ServicePermissions ?  

This way a dev won't ever forget to add a new route to ServicePermissions.

pkg/router/route.go
avisiedo: I got some cycle references; at the beginning I moved some code to the `cmd/content-sources/main.go` and it resolved, but I was not happy with that solution. Visiting some repository I saw that the routing for the api endpoints where set in a different package; at the beginning I thought "ey! I like how it looks", and I saw it resolved the dependency; but a deeper thinking reveals what solve the situation:

- in golang the dependency granularity happens at package level, so moving the code into a file in the same package would improve the clashing with changes into the repository, but it would not fix the dependency situation.
- the situation is that some code in `pkg/config` depends on `pkg/handler` and `pkg/handler` depends on `pkg/config`.
- The code that configure the route has the dependency on `pkg/handler` (and `pkg/middleware`), so when I extracted the code into a new package, now I have the below dependencies:
  - `pkg/routes` > `pkg/config`, `pkg/handler`, `pkg/middleware`, `pkg/instrumentation`, `pkg/client` (I moved them out of `pkg/config`).
  - `pkg/handler` > `pkg/config`, ...  (but not on `pkg/routes`).
  - `pkg/config` > `pkg/errors`, `pkg/event` (I can see a refactor here, moving back the kafka configuration to `pkg/config`, as I think that probably `pkg/event` now can depends on `pkg/config` with no cycle references; the tool that @rverdile mentioned recently will arise a better big picture, I have to try it).
- `pkg/config` is broadly referenced into the repository (below what I get with a quick `grep -HRn pkg/config pkg/`):

```raw
pkg/api
pkg/handler
pkg/models
pkg/external_repos
pkg/middleware
pkg/seeds
pkg/db
pkg/dao
pkg/route
```

So with moving the code to a new package, many risky cycle dependencies from `pkg/config` have been extracted.

Thanks for the question, that let me think deeper about this.

> Probably I am missing some dependency into the above, I didn't check all the import for the files at `pkg/config` package.

pkg/middleware/rbac_test.go
avisiedo: I like the idea, and I tried that, unfortunately I didn't find a way (I don't know if it exists) to do the check into the reverse direction; happy to know about it, as that would provide a powerful check detecting that situations in a very early stage (before push the commits) that is very interesting.

cmd/mock/rbac.go
avisiedo: sorry, I didn't move to a new repo this! I would create a ticket for enhance it in a future change.

avisiedo: https://issues.redhat.com/browse/HMSCONTENT-341

pkg/router/route.go
rverdile: I see the dependency on `pkg/handler` in `pkg/router`, but I'm curious why handler needs to be called there. `cmd/main.go` is registering the routes already, so why does it need they also need to be registered in the `ConfigureEcho()`?

With that said `pkg/router` would still be needed to avoid a circular dependency. There would be a circular dependency between `pkg/config` and `pkg/middleware` without `pkg/router`. 

I've been trying some ideas, but I'm not sure really sure what the best way to handle that is. If we want a config package, a middleware package, but no third package to handle the circular dependency, maybe the middleware can be registered in `cmd/main.go` after `ConfigureEcho() is called.

README.md
```diff
@@ -134,6 +134,27 @@ To open the prometheus web UI, once the container is up, run the below:
...
+
+- Now run: `make mock-start`
+- Run the application by: `make run`
+- Make some request using: `./scripts/header.sh 12345 12345` for admin or `./scripts/header.sh 12345 123456` for viewer.
```
jlsherrill, line 155: I think we want to update header.sh to support usernames, so you can run:  './scripts/header.sh 12345 admin'

jlsherrill: meaning we'd get rid of the account id completely

pkg/test/mocks/rbac/rbac.go
```diff
@@ -0,0 +1,192 @@
...
+			},
+		},
+	}
+
+	mocksConfig := config.Get().Mocks
+	orgId := mocksConfig.MyOrgId
+	if orgId != "" && orgId != xRhIdentity.Identity.OrgID {
+		log.Debug().Msgf("rbac requested for another org_id; returning an empty list of permissions")
+		return c.JSON(http.StatusOK, outputEmpty)
+	}
```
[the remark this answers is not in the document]
jlsherrill: meaning, the mock service would only care about username, and really not care about org_id

pkg/client/rbac.go
```diff
@@ -0,0 +1,88 @@
+package client
+
+/*
+Usage example:
+
+type myhandler struct {
+	clientRbac client.Rbac
+}
```
rverdile, line 8: Since `pkg/client` seems to be specific to Rbac, I'm curious why not have this code in `pkg/middleware/rbac.go` instead?

rverdile: To elaborate, is this code in `pkg/client` only intended to be used with the middleware code, or is it something you think could be used independently?

configs/config.yaml.example
```diff
@@ -49,6 +49,24 @@ certs:
...
+
+# Configuration for the mocks
+mocks:
+  # Update this with your org_id value
+  my_org_id: "12345"
+  rbac:
+    # Update with your account number for admin
+    user_write: ["jdoe@example.com"]
+    # Update with yout account number for viewer
+    user_read: ["tdoe@example.com"]
```
jlsherrill, line 72: make it simpler :) 
```
user_write: ["write"]
user_read: ["read]
```

I still think i'd prefer to allow write by all unless explicitly denied, so you'd have just:
```
user_denied: ["baduser"]
user_read:["reader"]
```

That way user opt-in to testing the rbac when they want, rather than be restricted to one or two users all the time.  i can see an argument that since this is an rbac mocking service, maybe it shouldn't default to open, but it is just a mocked service, never meant to be really used.

configs/config.yaml.example
```diff
@@ -49,6 +49,24 @@ certs:
...
+# the client services
+clients:
+  rbac_enabled: True
+  rbac_base_url: http://localhost:8800/api/rbac/v1
+  rbac_timeout: 30
+
+# Configuration for the mocks
+mocks:
+  # Update this with your org_id value
+  my_org_id: "12345"
```
jlsherrill, line 67: no need to have the org_id either :) simplify!

README.md
```diff
@@ -134,6 +134,27 @@ To open the prometheus web UI, once the container is up, run the below:
...
+    rbac_base_url: http://localhost:8800/api/rbac/v1
+    rbac_timeout: 30
+  mocks:
+    my_org_id: "12345"
+    rbac:
+      account_admin: "12345"
+      account_viewer: "123456"
+  ```
+
+- Now run: `make mock-start`
```
jlsherrill, line 153: i think this is a out of date and can be removed.

deployments/deployment.yaml
```diff
@@ -224,5 +228,11 @@ parameters:
   - name: MEMORY_REQUESTS
     value: 100Mi
   - name: LOGGING_LEVEL
-    value: warn
+    # value: warn
+    value: debug
```
rverdile, line 232: should this be changed back to warn before merging? just wanted to point this out so it's not forgotten about.

pkg/client/rbac.go
avisiedo: 1st comment
- I thought that a logical layer to break down is the client code to interact with third party platform services.
- I could have created a new file `pkg/middleware/rbac_client.go`, but I thought about it in the future as a common group of components with a common responsibility (the line above).
- The above line align with future feature integrating pulp service, so it would provide a place into the repository to hold that at `pkg/client`.
- If in the future we have an additional interaction for other platform service, we only had to add the code to the `pkg/client` package.

2nd comment

- I think it will be used only with the middleware code, but I see two logical responsibilities:
  - the middleware provide a centralised place to check the permissions.
  - the client provide the communication with the third party component (in this case it is only an adapter to the public client library). If it were generated from the rbac api, it would have generated a complete component to interact with the whole rbac api.
- I don't think it will be used independently.

configs/config.yaml.example
avisiedo: I have updated with the below:

```raw
user_read_write: ["user1","user2"]
user_read: ["user3"]
```

that match with the permissions returned.

to don't use the rbac check during developing other functionalities, it can be disabled by:

```yaml
clients:
  rbac_enabled: False
```

so the configuration for the allowed read/write and read is kept, and we can use it when we need, or disable when we are developing other features; by default it is enabled.

> I will update `configs.yaml.example` file to add comments about the above.

avisiedo: as `configs/config.yaml` is used only for local development, I have set `rbac_enabled: False` there, so if a new configuration is generated from it, by default the rbac will be disabled for the local environment.

I have added comments to clarify what happens when the value is True and False.

deployments/deployment.yaml
avisiedo: Yes it should be changed! many thanks!

jlsherrill:
https://issues.redhat.com/browse/HMS-470

jlsherrill:
https://issues.redhat.com/browse/HMS-240

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-470

jlsherrill (approved):

pkg/middleware/rbac_test.go
```diff
@@ -0,0 +1,554 @@
...
+		{
+			Name: "Skipper return true",
+			Given: TestCaseGiven{
+				GenerateIdentity: false,
+				Request: TestCaseGivenRequest{
+					Method: http.MethodGet,
+					Path:   testPath + "/repositories/",
+				},
+				MockResponse: TestCaseGivenRbac{
+					// Resource: "repositories",
```
mshriver, line 170: Why is this commented out here?

avisiedo: thanks mate! This is commented out because the mocked interface is not called for that test case; If I fill the information, it is considered that should be called, the mocked interface is prepared, and the test execution was failing because no call was received on the mocked instance.

I have added a note about it and cleaned-up the comments.

PR #151, opened 2022-11-30
Fixes 267: Add a metrics endpoint

avisiedo:
Currently:
- Provide and endpoint to pull the metrics (instrumentation service).
- Provide local infra to check the metrics from prometheus ui (make prometheus-{up,down,clean,logs,ui}).
- Provide local prometheus configuration to check it locally.
- Update dependencies.
- Fix a wrong behavior for containers in macos systems (healthcheck) when podman and docker are installed into the same system.
- Create a common place to set the metrics to export at `pkg/instrumentation/metrics.go`.

TODO

- [ ] Some metrics seems absolute values for the whole service; how could this impact when every pod export the same metric?
- [ ] If the response to the above point is that it can impact, then study and ask about decompose in some discrete metrics that could be used for a prometheus query that finally report the expected values.
- [x] About the introspection result to use it for calculate repositories not introspected into the last 24 hours, it currently happens in a different pod; now that they could be launched from the same binary, could make sense to use only one pod and start the kafka-consumer from the unique pod? Keeping api and kafka consumer running in different pods.
- [x] Declare in app-interface the metric endpoint for the service.
- [x] Unit tests.
- [ ] Top50 repositories is not implemented on this pr; I have not found a way to add that information with prometheus metric; we store counter, specific value that go up and down, histogram and summaries, but for the 50 repositories, not clear; cross my mind to store the position as the value, and the url repository as a label of that position; as soon we have 50 repositories or more, when a repository is out of the list, the last value scrapped is that remain for that label, which it does not represent the reality.
- [x] unit test for `collector.go` file.


jlsherrill:
>  About the introspection result to use it for calculate repositories not introspected into the last 24 hours, it currently happens in a different pod; now that they could be launched from the same binary, could make sense to use only one pod and start the kafka-consumer from the unique pod?

I would assume this would be a db query, so it doesn't matter which pod is asked this question?  They would all return the same answer. (but maybe i'm missing something)

pkg/instrumentation/metrics.go
```diff
@@ -0,0 +1,45 @@
...
+			Help: "The number of repositories introspected with failure",
+		}),
+	}
+
+	reg.MustRegister(metrics.introspectionSuccess)
+	reg.MustRegister(metrics.introspectionFailure)
+	reg.MustRegister(metrics.repositoriesCreated)
+	reg.MustRegister(metrics.repositoriesDeleted)
+
+	reg.MustRegister(collectors.NewBuildInfoCollector())
```
rverdile, line 42: Playing around with prometheus, I learned that using `promauto` will automatically register the metrics with the `DefaultRegisterer`. In which case, I don't think you need to call `MustRegister` here.

cmd/content-sources/main.go
```diff
@@ -96,3 +109,36 @@ func apiServer(ctx context.Context, wg *sync.WaitGroup, allRoutes bool) {
...
+			EnableOpenMetrics: true,
+			// Pass custom registry
+			Registry: reg,
+		},
+	)))
+	e.HideBanner = true
+	go func() {
+		defer wg.Done()
+		log.Logger.Info().Msgf("Starting instrumentation")
+		if err := e.Start(":9000"); err != nil && err != http.ErrServerClosed {
```
rverdile, line 130: Just curious, why start the metrics endpoint on another port? I assumed it would be the same as our API server, but just a new endpoint. 

I couldn't find documentation anywhere to suggest one way or the other, so I was wondering why you chose this way.

avisiedo: thanks for this question;  I am not sure if I have used something that I read from another repository (very likely); but now your concern is my concern :) and it made me dig about this and I have found a response that remove my uncertainty about this.

In a practical way, I have found the information about the port and the endpoint into the clowdenvirontment resource (I have reserved a namespace in the ephemeral environment and I have printed out the environment configuration by `oc get clowdenvironments.cloud.redhat.com/env-$( oc project -q ) -o yaml | yq --yaml-output '.spec.providers.metrics'`). The above provide the port (`9000`) and the path (`/metrics`) that will be used for the instrumentation.

The above information is injected by clowder, so I have to read it from the configuration and set up the endpoint according to it.

Thanks a million, I will update that!

pkg/instrumentation/metrics.go
avisiedo: You are right about `promauto`! but for any reason that I don't understand yet, when I integrated that into the service, I couldn't get the metric into the response; it was when I used this approach when the metric started to be reported; I don't know the root cause honestly.

avisiedo: I have just seen that invokation to `promauto.NewCounter` is invoking `With(prometheus.DefaultRegisterer).NewCounter`; I am doing something wrong here.

avisiedo:
@jlsherrill @rverdile thanks for the comments! I will update the code based on the comments:
- I will add a go routine to update periodically the metrics from the database.
- I will try to avoid the leverage of DefaultRegisterer.
- I will use the path and port indicated by clowder.

cmd/content-sources/main.go
rverdile: I've looked through some other apps, and I see some examples where metrics is an endpoint on the same server:
- [image builder](https://github.com/osbuild/image-builder/blob/d8d15007839a1a555d1fe0cd1d9ec1495d913140/internal/v1/server.go)
- [enititlements-api-go](https://github.com/RedHatInsights/entitlements-api-go/blob/c7c29cc8c8b00c8a8a1dbf1e7d16dc4584448e34/server/routes.go)

I see what you mean, but I wonder if what clowder injects is configurable from somewhere. I think if we might want to avoid it for simplicity.

pkg/instrumentation/metrics.go
rverdile: If you wanted to avoid DefaultRegisterer, I believe you can do `prometheus.NewCounter` instead.

cmd/content-sources/main.go
```diff
@@ -7,19 +7,27 @@ import (
...
+	m "github.com/content-services/content-sources-backend/pkg/instrumentation"
+	"github.com/labstack/echo/v4"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/rs/zerolog/log"
 )
 
 func main() {
+	reg := prometheus.NewRegistry()
+	metrics := m.NewMetrics(reg)
```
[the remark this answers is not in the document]
rverdile: I see what you mean, thanks. When I did this, I used a global variable, which I liked because it kept everything really simple. But if it's important to avoid that, I see why dependency injection would be useful. I also wasn't thinking about unit tests :)

mk/prometheus.mk
```diff
@@ -0,0 +1,51 @@
...
+export PROMETHEUS_CONFIG
+export PROMETHEUS_VERSION
+
+.PHONY: prometheus-up
+prometheus-up: ## Start prometheus service (local access at http://localhost:9090)
+	$(DOCKER) volume inspect prometheus &> /dev/null || $(DOCKER) volume create prometheus
+	$(DOCKER) container inspect prometheus &> /dev/null || \
+	$(DOCKER) run -d \
+	  --name prometheus \
+	  --restart always \
```
rverdile, line 29: is this container stoppable, even through Docker, if `--restart` is set to `always`? Reading here, it seems like it isn't https://docs.docker.com/engine/reference/commandline/run/#restart-policies---restart

avisiedo: thanks mate! I have seen the prometheus container was not removed when calling `make prometheus-down`. I have updated both rules `prometheus-up` to remove `--restart always` (I added this to trace some situations when spinning up the container) and added the `--rm` option to remove the container when it is stopped.

You could need to run `make prometheus-clean`; if after run the above you see the prometheus container when running `podman container ls -a`, then run `podman container rm prometheus`. After this, and rebasing with the changes into the make rules, the expected lifecycle for the container using make rules should works properly (checked with podman, I have to check with docker).

pkg/instrumentation/middleware.go
```diff
@@ -0,0 +1,60 @@
...
+			method := ctx.Request().Method
+			path := matchedRoute(ctx)
+
+			config.Metrics.HttpTotalRequests.Inc()
+			err := next(ctx)
+			if ctx.Response().Status >= 500 {
+				config.Metrics.HttpTotalFailedRequests.Inc()
+			}
+			status := fmt.Sprintf("%d", ctx.Response().Status)
+			config.Metrics.HttpRequestLatency.WithLabelValues(status, method, path).Observe(float64(time.Since(start).Nanoseconds()) / 1000000)
```
rverdile, line 56: Check my math, but I think this is milliseconds (1 nanosecond = 1e-6 ms). But the metric is seconds. [ObserveDuration](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Timer.ObserveDuration) would return seconds.

Or there is a Milliseconds [function](https://pkg.go.dev/time#Duration.Milliseconds) and we could use milliseconds for the metric.

avisiedo: :+1: Nice catch! I have to divide by 1000000000.

I have just realized that the `.Seconds()` method is implemented like the below:

```go
func (d Duration) Seconds() float64 {
	sec := d / Second
	nsec := d % Second
	return float64(sec) + float64(nsec)/1e9
}
```

So I could simplify the expression with just:

```go
config.Metrics.
    HttpRequestLatency.
    WithLabelValues(status, method, path).
    Observe(time.Since(start).Seconds())
```

pkg/config/config.go
```diff
@@ -248,8 +268,65 @@ func CustomHTTPErrorHandler(err error, c echo.Context) {
 	}
 }
 
-func ConfigureEcho() *echo.Echo {
-	e := echo.New()
+func getResource(path string) string {
```
rverdile, line 271: What is the function for? I can't figure it out :)

pkg/config/config.go
```diff
@@ -248,8 +268,65 @@ func CustomHTTPErrorHandler(err error, c echo.Context) {
...
+}
+
+func metricsMiddlewareSkipper(ctx echo.Context) bool {
+	path := ctx.Path()
+	switch {
+	case strings.HasPrefix(path, "/ping/"):
+		return true
+	case strings.HasPrefix(path, "/metrics"):
+		return true
+	case hasResources(path, "validation", "ping"):
```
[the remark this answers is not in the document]
avisiedo: yes, "validation" means to skip the `repository/validate` endpoint! thanks! I have to change it.

pkg/config/config.go
avisiedo: It was intended to retrieve the main resource of the API request: `repositories` for `/api/v1/repositories` or `/api/v1/repositories/{uuid}` or `/beta/api/v1/repositories/{uuid}`
but I have to update it; thanks mate

deployments/deployment.yaml
```diff
@@ -118,6 +118,11 @@ objects:
                   - introspect
                   - https://cdn.redhat.com/content/dist/layered/rhel8/x86_64/ansible/2/os
             image: ${IMAGE}:${IMAGE_TAG}
+            command:
+              - /content-sources
+              - api
+              - consumer
+              - instrumentation
```
jlsherrill, line 125: maybe this was a temporary change, but i don't think we want consumer to run as part of this right?

avisiedo: thanks! I changed locally and I have committed at some point!

pkg/config/config.go
```diff
@@ -248,8 +269,35 @@ func CustomHTTPErrorHandler(err error, c echo.Context) {
...
+
+func createMetricsMiddleware(metrics *instrumentation.Metrics) echo.MiddlewareFunc {
+	return instrumentation.MetricsMiddlewareWithConfig(
+		&instrumentation.MetricsConfig{
+			Skipper: metricsMiddlewareSkipper,
+			Metrics: metrics,
+		})
+}
+
+func configureEchoCommon(e *echo.Echo, metrics *instrumentation.Metrics) *echo.Echo {
```
rverdile, line 299: Since `configureEchoCommon` takes metrics but does not use it, I suggest changing the echo configuration functions to [these](https://gist.github.com/rverdile/150de14374ee08d0d182daa16f8e5033). 

With this there's only two functions: one called when creating the API server, and the other called when creating the metrics server. Hopefully that makes sense.

pkg/config/config.go
```diff
@@ -248,8 +269,35 @@ func CustomHTTPErrorHandler(err error, c echo.Context) {
...
+		return true
+	case path == "/metrics" || path == "/metrics/":
+		return true
+	}
+	pathItems := handler_utils.NewPathWithString(path).RemovePrefixes()
+	if pathItems.StartWithResources(
+		[]string{"ping"},
+		[]string{"repositories", "validate"},
+		[]string{"repository_parameters", "validate"},
+	) {
```
rverdile, line 285: I was thinking about what needs to be skipped. Since the `/metrics` endpoint is on a different port, I don't think it needs to be skipped here. It would be an unregistered route on this server. In which case, only `/ping` needs to be skipped. 

Also, I think you could just hard code this and avoid the tricky path parsing.

pkg/instrumentation/metrics.go
```diff
@@ -0,0 +1,42 @@
...
+func NewMetrics(reg *prometheus.Registry) *Metrics {
+	if reg == nil {
+		panic("reg cannot be nil")
+	}
+	metrics := &Metrics{
+		reg: reg,
+		HttpStatusHistogram: *promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+			Name:    HttpStatusHistogram,
+			Help:    "http status histogram",
+			Buckets: prometheus.LinearBuckets(0.01, 0.05, 10),
```
rverdile, line 31: I think you should add `Namespace: "content_sources"` here. This will prepend "content_sources_" to the metric. This will make it easier to identify when using the shared prometheus server.

pkg/instrumentation/metrics.go
```diff
@@ -0,0 +1,42 @@
...
+// See: https://prometheus.io/docs/tutorials/understanding_metric_types/#types-of-metrics
+func NewMetrics(reg *prometheus.Registry) *Metrics {
+	if reg == nil {
+		panic("reg cannot be nil")
+	}
+	metrics := &Metrics{
+		reg: reg,
+		HttpStatusHistogram: *promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+			Name:    HttpStatusHistogram,
+			Help:    "http status histogram",
```
rverdile, line 30:
```suggestion
			Help:    "Duration of HTTP requests",
```

pkg/instrumentation/metrics.go
rverdile, line 31: I think the buckets should have more general/widespread values. There's [DefBuckets](https://github.com/prometheus/client_golang/blob/v1.14.0/prometheus/histogram.go#L265) provided by prometheus, or you could define the buckets manually.

pkg/config/config.go
```diff
@@ -54,11 +58,22 @@ type Options struct {
...
+	// metric traffic.
+	Port int `mapstructure:"port"`
+}
+
 const (
 	DefaultPagedRpmInsertsLimit = 500
 )
 
 var LoadedConfig Configuration
+var configMutex *sync.Mutex = &sync.Mutex{}
```
rverdile, line 76: It looks like this is unused. You may have forgotten to remove it.

pkg/dao/interfaces.go
```diff
@@ -34,3 +34,12 @@ type ExternalResourceDao interface {
...
 }
+
+type MetricsDao interface {
+	RepositoriesCount() int
+	RepositoryConfigsCount() int
+	PublicRepositoriesNotIntrospectedLas24HoursCount() int
+	PublicRepositoriesFailedIntrospectionCount() int
+	NonPublicRepositoriesNonIntrospectedLast24HoursCount() int
+	Top50Repositories() int
+}
```
rverdile, line 45: I'm not sure if you planned to finish implementing these, but I suggest we leave additional metrics for another issue.

rverdile (commented):
I know you're still working on this, so some of my comments might already be outdated by changes you've made (or plan to make) locally. Some of the comments are also on small things.

I'm just leaving a bunch since I'll be out tomorrow :).

pkg/dao/interfaces.go
avisiedo: :+1: I will add them into a new ticket; I forgot about the `interfaces.go`! thanks mate!

avisiedo:
I have updated this pr by adding the custom-metrics but the top50 repositories, and their unit tests, but for the collector.

deployments/deployment.yaml
```diff
@@ -195,6 +199,9 @@ objects:
           port: 8000
           protocol: TCP
           targetPort: 8000
+        - name: metrics
+          port: 9000
+          targetPort: 9000
```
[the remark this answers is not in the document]
avisiedo: my guess is this is not necessary; the `content-sources-backend-service` is created that reference the ports indicated here; I tried to create the ServiceMonitor resource into ephemral environment and it was created with no errors related with a port name not existing (if that check were made). I can see several services related with the pods are created when deploying in the ephemeral environment.

pkg/config/config.go
avisiedo: thanks mate! I tracked down information about the path and port for the metrics, and if it is actually hardcoded into a clowdenvironment resource, if that information change, would make wrong several things; I have to update all the matchings with '/metrics' to use the path read from configuration and the same for the metrics port.

pkg/dao/interfaces.go
avisiedo: I have implemented all but Top50Repositories, and I have added the remaining metric in a new ticket.

pkg/dao/metrics_test.go
```diff
@@ -0,0 +1,255 @@
...
+	// s.repoPrivate = repoPrivate
+
+	// // Create repository configuration entry
+	// repoConfig := repoConfigTest1.DeepCopy()
+	// // repoConfig.Repository = *repoPublic
+	// repoConfig.RepositoryUUID = repoPublic.Base.UUID
+	// if err := s.tx.Create(&repoConfig).Error; err != nil {
+	// 	s.FailNow("Preparing RepositoryConfiguration record UUID=" + repoConfig.UUID)
+	// }
+	// s.repoConfig = repoConfig
```
jlsherrill, line 57: some cleanup here needed :)

pkg/instrumentation/custom/collector.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (c *Collector) iterateTop50Repositories() {
+	// for _, row := range c.dao.Top50Repositories() {
+	// 	url := row["url"].(string)
+	// 	repo_count := row["repo_count"].(int64)
+	// 	c.metrics.Top50Repositories.WithLabelValues(
+	// 		url,
+	// 	).Set(float64(repo_count))
+	// }
+}
```
jlsherrill, line 63: cleanup :)

pkg/config/config.go
```diff
@@ -258,11 +302,24 @@ func ConfigureEcho() *echo.Echo {
...
+	e := echo.New()
+	configureEchoCommon(e)
+	return e
+}
+
+func ConfigureEchoService(metrics *instrumentation.Metrics) *echo.Echo {
+	e := echo.New()
+	e.Use(createMetricsMiddleware(metrics))
+	configureEchoCommon(e)
+	e.Use(WrapMiddlewareWithSkipper(identity.EnforceIdentity, SkipLiveness))
```
rverdile, line 323: Following up from my previous comment, I see you removed `metrics` from `configureEchoCommon`, which I think is a good simplification.

I still think you could make these functions even simpler, by replacing `configureEchoCommon`, `ConfigureEchoMetrics` and `ConfigureEchoService` with [these two functions](https://gist.github.com/rverdile/150de14374ee08d0d182daa16f8e5033).

If you don't like that, I think you should at least rename your current functions: `ConfigureEchoMetrics -> ConfigureEcho` and `ConfigureEchoService -> ConfigureEchoMetrics`. The current naming seems backwards.

pkg/config/config.go
```diff
@@ -248,8 +267,33 @@ func CustomHTTPErrorHandler(err error, c echo.Context) {
...
+		return true
+	case path == "/metrics" || path == "/metrics/":
+		return true
+	}
+	pathItems := handler_utils.NewPathWithString(path).RemovePrefixes()
+	return pathItems.StartWithResources(
+		[]string{"ping"},
+		[]string{"repositories", "validate"},
+		[]string{"repository_parameters", "validate"},
+	)
```
rverdile, line 284: What is `StartWithResources` used for?

pkg/dao/interfaces.go
```diff
@@ -34,3 +34,12 @@ type ExternalResourceDao interface {
 	FetchSignature(url string) (*string, int, error)
 	FetchRepoMd(url string) (*string, int, error)
 }
+
+type MetricsDao interface {
```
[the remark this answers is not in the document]
avisiedo: It is weird! yes, I can see all the metrics!

try the below:
- stop prometheus
- check the scraper configuration is pointing out to the hostname (no localhost)
- check the scrapper interval is a low value (the repo configuration has 3 seconds).
- start prometheus
- do some actions from the frontend (I use to add / delete popular repositories), this will generate information for the metrics.
- await 3 seconds (the scraper interval value).
- refresh prometheus-ui.
- do click on the metrics explorer (it requires to refresh the page); this icon is on the left of `execute` button. I expect here to see all the metrics.

If the above is still failing, let me know, I would need to dig more.

pkg/config/config.go
avisiedo: thanks! I like it! I will update with this! sorry if I missed it

pkg/instrumentation/custom/collector.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (c *Collector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- c.metrics.RepositoriesTotal.Desc()
+	ch <- c.metrics.RepositoryConfigsTotal.Desc()
+	ch <- c.metrics.PublicRepositoriesNotIntrospectedLast24HoursTotal.Desc()
+	ch <- c.metrics.PublicRepositoriesWithFailedIntrospectionTotal.Desc()
+	ch <- c.metrics.NonPublicRepositoriesNotIntrospectedLast24HoursTotal.Desc()
+}
+
+func (c *Collector) Collect(ch chan<- prometheus.Metric) {
```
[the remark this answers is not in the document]
avisiedo: yes and no! I wrongly added the Collect interface, but it is not being used, because the code that register this is commented;  so I thought I was using that (but not used) and I thought I was gathering the values using that interface (but it is not the case); I have removed as that could evoke wrong behaviors. By the way, I think I have implemented wrongly.

What I think about when it should be used that interface:
- When I am defining a new more complex new metric not covered by the basic types, so one method gather the description of the metric, and the other write all the values of the metric (it could be several values generated as an histogram, or several values with different labels, etc...); I name it value to distinguish from the metric that could have several values; a gauge with labels could print out several values too and I give the scope of metric to one description.

What I think the dead code would do (wrongly):
- It would mix all the descriptions of the different metrics.
- It would write all the values together as part of the last description.
- It would not match the metrics descriptions with the expected metric value.

I wrongly assumed it was working because I saw the metrics into the curl request! many thanks mate!

pkg/config/config.go
avisiedo: It is used for matching the starting items of the path once it is removed the prefix items ([/beta]/api/<service-name>/<version>).

Examples of matching:
* For "/api/content-sources/ping/something-else" would return true as it would match with `[]string{"ping"}`.
* For "/api/content-sources/repositories/validate/" would return true.
* For "/api/content-sources/anyotherthing" would return false.
* For "/api/repositories" would return false.

pkg/instrumentation/custom/collector.go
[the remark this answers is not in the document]
rverdile: Ah okay, thanks! I was just wondering what the methods were for, but I'm glad I also accidentally helped you find a bug :).

pkg/dao/interfaces.go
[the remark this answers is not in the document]
rverdile: Ah, it's working now! I forgot to change `prometheus.yaml` from localhost.

On a similar note, do we need the following? 

1.  add `configs/prometheus.yaml` to `.gitignore`  
2.  add `configs/prometheus.yaml.example`

pkg/config/config.go
rverdile: Thanks. A couple more questions/comments.

1. `/ping` is already being skipped earlier in the code, so does it still need to be skipped here?
2. There's no `repositories/validate` endpoint, so I don't think this line is needed.
3. Why skip `repository_parameters/validate`? If that gave a `5xx` error, wouldn't we want prometheus to count that?

avisiedo: 1. `ping` is skipped for enforceidentity wrapper middleware, but that skipper apply only to that middleware, is another middleware is before or after the chain, and the chain is not interrupted, the skip would not apply and it needs to check its own cases.
2. Thanks! I missed that.
3. At some point I recall it was mentioned to do not get metrics for the validation endpoints, but I don't recall the motivation for it; @rverdile @jlsherrill  if I am wrong, let me know and I remove the skipper for it.

rverdile: 1. To clarify, `ping` is also skipped in the first case of the switch statement in `metricsMiddlewareSkipper`, right before `StartWithResources` is called. With this in mind, did you still mean to pass ping to `StartWithResources`?

3. Ah okay, then I'm not sure whether or not validation should be skipped. It's a question for @jlsherrill :).

avisiedo: Sorry, I thought it was a concern about the other middleware skipper. It is correct the filter for `/ping` into the first switch; we have 2 different paths for ping `/ping` and `/api/content-sources/v1/ping`; the switch is skipping the first ping path, the StartWithResources is skipping the second ping path.

jlsherrill: 2)  The purpose of skipping that particular endpoint for metrics is that we don't want to measure things like latency on that endpoint.  The latency would be higher because its reaching out to  the internet to grab the repomd.xml file.

pkg/dao/metrics.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (d metricsDaoImpl) PublicRepositoriesNotIntrospectedLas24HoursCount() int {
+	// select COUNT(*) from repositories where last_introspection_update_time < NOW() - INTERVAL '24 hours' and public;
+	var output int64 = -1
+	d.db.
+		Model(&models.Repository{}).
+		// TODO Ask if we want to include the repositories which have not been introspected yet (Pending state)
+		Where(d.db.Where("last_introspection_update_time < NOW() - INTERVAL '24 hours'").
+			Or("last_introspection_update_time is NULL")).
+		Where("public").
```
jlsherrill, line 48: curious why you are limiting it to 'public' repositories?   Also, i don't think we actually want 'pending' repos, we want where last_introspection_time is not in the past 24 hours.

pkg/dao/metrics.go
```diff
@@ -0,0 +1,81 @@
...
+	// from repositories
+	// where status in ('Invalid','Unavailable')
+	//   and public;
+	var output int64 = -1
+	d.db.
+		Model(&models.Repository{}).
+		Where("status in (?, ?)", config.StatusInvalid, config.StatusUnavailable).
+		Where("public").
+		Count(&output)
+	return int(output)
```
jlsherrill, line 64: same question here about Public?

jlsherrill: oh i see you're logging them separately, ignore me :)

pkg/dao/metrics.go
```diff
@@ -0,0 +1,81 @@
...
+	// select COUNT(*)
+	//   from repositories
+	//  where last_introspection_update_time < NOW() - INTERVAL '24 hours'
+	//    and status in ('Invalid','Unavailable')
+	//    and not public;
+	var output int64 = -1
+	d.db.
+		Model(&models.Repository{}).
+		Where("status in (?, ?)", config.StatusInvalid, config.StatusUnavailable).
+		Where("last_introspection_update_time < NOW() - INTERVAL '24 hours'").
```
jlsherrill, line 77: same question here about last_intropsection_time.

pkg/dao/metrics.go
jlsherrill: And to add to this, i think the goal here is to detect if something has gone wrong, and the daily introspections are not happening for some repository.

pkg/config/config.go
rverdile: Do we also want to skip counting errors caused by reaching out to the internet? 

An alternative option is to filter out the endpoint when querying for the latency metrics i.e. `content_sources_http_status_histogram_bucket{le="0.01",path!="/api/content-sources/v1.0/repository_parameters/validate"}`.

jlsherrill: Ah yes, thats a better option! Lets do that :)  that may simplify some code as well.

avisiedo: I have removed `repository_parameters/validate` route for the skipper, and I missed to remove the unexisting route `repositories/validate`. Thanks for the comments

pkg/dao/metrics.go
avisiedo: thanks updated!

pkg/dao/metrics.go
avisiedo: :+1: thanks for the clarification! I have updated the queries!

avisiedo:
Test requirements:

```raw
Scenario 1: metrics endpoint is up and listening

Given the service is up and running
    and the metrics endpoint configuration read from clowder
  When the metrics endpoint is reached out (http://content-sources-backend:9000/metrics)
   Then a get a response with the current metrics value and description.

----

Scenario 2: metrics are generated for total number of repositories

Given the service up and running
    and one repository created into the database
    and two repository configurations into the database pointing to the above repo
    and another one public repository introspected beyond 24 hours above in Invalid or Failed
    and another one public repository with Failed state before the 24 hours boundary
    and another one non public repository with state Invalid or Failed beyond 24 hours
  When the metrics endpoint is scrapped
   Then the number of repositories count is 4
     and the number of repository config count is 2
     and the number of public repositories not introspected in the last 24 hours is 1
     and the number of non publis repositories not introspected in the last 24 hours is 1
     and the number of failed repositories is 1 (we have 1 repo bayond 24 hours and 1 before)
     
** I was not able to think a way to get repos in the above states without create the entry directly into the database.

----

Scenario 3: api metrics

Given the service up and running
    and two repositories created with the api or ui
   and one of the above repositories deleted
  When the metrics endpoint is scrapped
   Then we get the metrics count of the api status according to the previous operations

Example of content scrapped by `curl http://localhost:9000/metrics`:

# HELP content_sources_http_status_histogram Duration of HTTP requests
# TYPE content_sources_http_status_histogram histogram
content_sources_http_status_histogram_bucket{method="DELETE",path="/api/content-sources/v1/repositories/:uuid",status="2xx",le="0.005"} 0
content_sources_http_status_histogram_bucket{method="DELETE",path="/api/content-sources/v1/repositories/:uuid",status="2xx",le="0.01"} 0
content_sources_http_status_histogram_bucket{method="DELETE",path="/api/content-sources/v1/repositories/:uuid",status="2xx",le="0.025"} 1
content_sources_http_status_histogram_bucket{method="DELETE",path="/api/content-sources/v1/repositories/:uuid",status="2xx",le="0.05"} 1
content_sources_http_status_histogram_bucket{method="DELETE",path="/api/content-sources/v1/repositories/:uuid",status="2xx",le="0.1"} 1
content_sources_http_status_histogram_bucket{method="DELETE",path="/api/content-sources/v1/repositories/:uuid",status="2xx",le="0.25"} 1
content_sources_http_status_histogram_bucket{method="DELETE",path="/api/content-sources/v1/repositories/:uuid",status="2xx",le="0.5"} 1
content_sources_http_status_histogram_bucket{method="DELETE",path="/api/content-sources/v1/repositories/:uuid",status="2xx",le="1"} 1
content_sources_http_status_histogram_bucket{method="DELETE",path="/api/content-sources/v1/repositories/:uuid",status="2xx",le="2.5"} 1
content_sources_http_status_histogram_bucket{method="DELETE",path="/api/content-sources/v1/repositories/:uuid",status="2xx",le="5"} 1
content_sources_http_status_histogram_bucket{method="DELETE",path="/api/content-sources/v1/repositories/:uuid",status="2xx",le="10"} 1
content_sources_http_status_histogram_bucket{method="DELETE",path="/api/content-sources/v1/repositories/:uuid",status="2xx",le="+Inf"} 1
content_sources_http_status_histogram_sum{method="DELETE",path="/api/content-sources/v1/repositories/:uuid",status="2xx"} 0.011591979
content_sources_http_status_histogram_count{method="DELETE",path="/api/content-sources/v1/repositories/:uuid",status="2xx"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1.0/repositories/:uuid/rpms",status="2xx",le="0.005"} 0
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1.0/repositories/:uuid/rpms",status="2xx",le="0.01"} 0
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1.0/repositories/:uuid/rpms",status="2xx",le="0.025"} 0
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1.0/repositories/:uuid/rpms",status="2xx",le="0.05"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1.0/repositories/:uuid/rpms",status="2xx",le="0.1"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1.0/repositories/:uuid/rpms",status="2xx",le="0.25"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1.0/repositories/:uuid/rpms",status="2xx",le="0.5"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1.0/repositories/:uuid/rpms",status="2xx",le="1"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1.0/repositories/:uuid/rpms",status="2xx",le="2.5"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1.0/repositories/:uuid/rpms",status="2xx",le="5"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1.0/repositories/:uuid/rpms",status="2xx",le="10"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1.0/repositories/:uuid/rpms",status="2xx",le="+Inf"} 1
content_sources_http_status_histogram_sum{method="GET",path="/api/content-sources/v1.0/repositories/:uuid/rpms",status="2xx"} 0.03865349
content_sources_http_status_histogram_count{method="GET",path="/api/content-sources/v1.0/repositories/:uuid/rpms",status="2xx"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/popular_repositories/",status="2xx",le="0.005"} 3
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/popular_repositories/",status="2xx",le="0.01"} 4
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/popular_repositories/",status="2xx",le="0.025"} 4
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/popular_repositories/",status="2xx",le="0.05"} 4
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/popular_repositories/",status="2xx",le="0.1"} 4
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/popular_repositories/",status="2xx",le="0.25"} 4
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/popular_repositories/",status="2xx",le="0.5"} 4
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/popular_repositories/",status="2xx",le="1"} 4
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/popular_repositories/",status="2xx",le="2.5"} 4
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/popular_repositories/",status="2xx",le="5"} 4
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/popular_repositories/",status="2xx",le="10"} 4
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/popular_repositories/",status="2xx",le="+Inf"} 4
content_sources_http_status_histogram_sum{method="GET",path="/api/content-sources/v1/popular_repositories/",status="2xx"} 0.014736989999999998
content_sources_http_status_histogram_count{method="GET",path="/api/content-sources/v1/popular_repositories/",status="2xx"} 4
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repositories/",status="2xx",le="0.005"} 5
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repositories/",status="2xx",le="0.01"} 5
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repositories/",status="2xx",le="0.025"} 5
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repositories/",status="2xx",le="0.05"} 5
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repositories/",status="2xx",le="0.1"} 5
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repositories/",status="2xx",le="0.25"} 5
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repositories/",status="2xx",le="0.5"} 5
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repositories/",status="2xx",le="1"} 5
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repositories/",status="2xx",le="2.5"} 5
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repositories/",status="2xx",le="5"} 5
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repositories/",status="2xx",le="10"} 5
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repositories/",status="2xx",le="+Inf"} 5
content_sources_http_status_histogram_sum{method="GET",path="/api/content-sources/v1/repositories/",status="2xx"} 0.008500505
content_sources_http_status_histogram_count{method="GET",path="/api/content-sources/v1/repositories/",status="2xx"} 5
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repository_parameters/",status="2xx",le="0.005"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repository_parameters/",status="2xx",le="0.01"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repository_parameters/",status="2xx",le="0.025"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repository_parameters/",status="2xx",le="0.05"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repository_parameters/",status="2xx",le="0.1"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repository_parameters/",status="2xx",le="0.25"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repository_parameters/",status="2xx",le="0.5"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repository_parameters/",status="2xx",le="1"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repository_parameters/",status="2xx",le="2.5"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repository_parameters/",status="2xx",le="5"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repository_parameters/",status="2xx",le="10"} 1
content_sources_http_status_histogram_bucket{method="GET",path="/api/content-sources/v1/repository_parameters/",status="2xx",le="+Inf"} 1
content_sources_http_status_histogram_sum{method="GET",path="/api/content-sources/v1/repository_parameters/",status="2xx"} 0.000141457
content_sources_http_status_histogram_count{method="GET",path="/api/content-sources/v1/repository_parameters/",status="2xx"} 1
content_sources_http_status_histogram_bucket{method="PATCH",path="/api/content-sources/v1.0/repositories/:uuid",status="2xx",le="0.005"} 0
content_sources_http_status_histogram_bucket{method="PATCH",path="/api/content-sources/v1.0/repositories/:uuid",status="2xx",le="0.01"} 0
content_sources_http_status_histogram_bucket{method="PATCH",path="/api/content-sources/v1.0/repositories/:uuid",status="2xx",le="0.025"} 1
content_sources_http_status_histogram_bucket{method="PATCH",path="/api/content-sources/v1.0/repositories/:uuid",status="2xx",le="0.05"} 1
content_sources_http_status_histogram_bucket{method="PATCH",path="/api/content-sources/v1.0/repositories/:uuid",status="2xx",le="0.1"} 1
content_sources_http_status_histogram_bucket{method="PATCH",path="/api/content-sources/v1.0/repositories/:uuid",status="2xx",le="0.25"} 1
content_sources_http_status_histogram_bucket{method="PATCH",path="/api/content-sources/v1.0/repositories/:uuid",status="2xx",le="0.5"} 1
content_sources_http_status_histogram_bucket{method="PATCH",path="/api/content-sources/v1.0/repositories/:uuid",status="2xx",le="1"} 1
content_sources_http_status_histogram_bucket{method="PATCH",path="/api/content-sources/v1.0/repositories/:uuid",status="2xx",le="2.5"} 1
content_sources_http_status_histogram_bucket{method="PATCH",path="/api/content-sources/v1.0/repositories/:uuid",status="2xx",le="5"} 1
content_sources_http_status_histogram_bucket{method="PATCH",path="/api/content-sources/v1.0/repositories/:uuid",status="2xx",le="10"} 1
content_sources_http_status_histogram_bucket{method="PATCH",path="/api/content-sources/v1.0/repositories/:uuid",status="2xx",le="+Inf"} 1
content_sources_http_status_histogram_sum{method="PATCH",path="/api/content-sources/v1.0/repositories/:uuid",status="2xx"} 0.01137019
content_sources_http_status_histogram_count{method="PATCH",path="/api/content-sources/v1.0/repositories/:uuid",status="2xx"} 1
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repositories/bulk_create/",status="2xx",le="0.005"} 0
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repositories/bulk_create/",status="2xx",le="0.01"} 0
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repositories/bulk_create/",status="2xx",le="0.025"} 2
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repositories/bulk_create/",status="2xx",le="0.05"} 2
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repositories/bulk_create/",status="2xx",le="0.1"} 2
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repositories/bulk_create/",status="2xx",le="0.25"} 2
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repositories/bulk_create/",status="2xx",le="0.5"} 2
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repositories/bulk_create/",status="2xx",le="1"} 2
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repositories/bulk_create/",status="2xx",le="2.5"} 2
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repositories/bulk_create/",status="2xx",le="5"} 2
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repositories/bulk_create/",status="2xx",le="10"} 2
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repositories/bulk_create/",status="2xx",le="+Inf"} 2
content_sources_http_status_histogram_sum{method="POST",path="/api/content-sources/v1.0/repositories/bulk_create/",status="2xx"} 0.023565302
content_sources_http_status_histogram_count{method="POST",path="/api/content-sources/v1.0/repositories/bulk_create/",status="2xx"} 2
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repository_parameters/validate/",status="2xx",le="0.005"} 0
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repository_parameters/validate/",status="2xx",le="0.01"} 0
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repository_parameters/validate/",status="2xx",le="0.025"} 0
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repository_parameters/validate/",status="2xx",le="0.05"} 0
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repository_parameters/validate/",status="2xx",le="0.1"} 0
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repository_parameters/validate/",status="2xx",le="0.25"} 0
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repository_parameters/validate/",status="2xx",le="0.5"} 0
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repository_parameters/validate/",status="2xx",le="1"} 1
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repository_parameters/validate/",status="2xx",le="2.5"} 2
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repository_parameters/validate/",status="2xx",le="5"} 4
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repository_parameters/validate/",status="2xx",le="10"} 4
content_sources_http_status_histogram_bucket{method="POST",path="/api/content-sources/v1.0/repository_parameters/validate/",status="2xx",le="+Inf"} 4
content_sources_http_status_histogram_sum{method="POST",path="/api/content-sources/v1.0/repository_parameters/validate/",status="2xx"} 8.110344537
content_sources_http_status_histogram_count{method="POST",path="/api/content-sources/v1.0/repository_parameters/validate/",status="2xx"} 4
# HELP content_sources_non_public_repositories_not_introspected_last_24_hours_total Number of non public repositories not introspected in the last 24 hours
# TYPE content_sources_non_public_repositories_not_introspected_last_24_hours_total gauge
content_sources_non_public_repositories_not_introspected_last_24_hours_total 0
# HELP content_sources_public_repositories_not_introspected_last_24_hours_total Number of public repositories not introspected into the last 24 hours
# TYPE content_sources_public_repositories_not_introspected_last_24_hours_total gauge
content_sources_public_repositories_not_introspected_last_24_hours_total 0
# HELP content_sources_public_repositories_with_failed_introspection_total Number of repositories with failed introspection
# TYPE content_sources_public_repositories_with_failed_introspection_total gauge
content_sources_public_repositories_with_failed_introspection_total 0
# HELP content_sources_repositories_total Number of repositories
# TYPE content_sources_repositories_total gauge
content_sources_repositories_total 1
# HELP content_sources_repository_configs_total Number of repository configurations
# TYPE content_sources_repository_configs_total gauge
content_sources_repository_configs_total 1
# HELP go_build_info Build information about the main Go module.
# TYPE go_build_info gauge
go_build_info{checksum="",path="",version=""} 1
# HELP promhttp_metric_handler_errors_total Total number of internal errors encountered by the promhttp metric handler.
# TYPE promhttp_metric_handler_errors_total counter
promhttp_metric_handler_errors_total{cause="encoding"} 0
promhttp_metric_handler_errors_total{cause="gathering"} 0



```

from the above can be derived more specific scenarios as they are considered.

> by scrapping the metrics endpoint it is:
> for the workstation: `curl http://localhost:9000/metrics`
> for the ephemeral environment: `curl http://content-sources-backend-service:9000/metrics`
> where <service>=content-sources-backend

rverdile (approved):
looks good!!

pkg/dao/metrics.go
jlsherrill: doesn't look like this has updated?

pkg/dao/metrics.go
```diff
@@ -0,0 +1,86 @@
...
+	//  where not public
+	//    and status in ('Invalid','Unavailable')
+	//    and (last_introspection_update_time < NOW() - INTERVAL '24 hours' or last_introspection_update_time is NULL);
+	var output int64 = -1
+	d.db.
+		Model(&models.Repository{}).
+		Where("not public").
+		Where("status in (?, ?)", config.StatusInvalid, config.StatusUnavailable).
+		Where(d.db.Where("last_introspection_update_time is NULL").
+			Or("last_introspection_update_time < NOW() - INTERVAL '24 hours'")).
```
jlsherrill, line 83: same thing here, i think we should look at last_introspection_time

pkg/dao/metrics.go
avisiedo: it is weird! I don't know why this code block is not flaged as `outdated`; the changes are pushed to the repository into the branch; I can see them here: https://github.com/avisiedo/content-sources-backend/blob/hmscontent-267/pkg/dao/metrics.go#L40

jlsherrill (approved):

swadeley:
Hi

lgtm


(.iqe-env) [sjw@t480s iqe-content-sources-plugin]$ oc rsh -n $NAMESPACE $POD

sh-4.4$  curl http://content-sources-backend-service:9000/metrics
# HELP content_sources_non_public_repositories_not_introspected_last_24_hours_total Number of non public repositories not introspected in the last 24 hours
# TYPE content_sources_non_public_repositories_not_introspected_last_24_hours_total gauge
content_sources_non_public_repositories_not_introspected_last_24_hours_total 0
# HELP content_sources_public_repositories_not_introspected_last_24_hours_total Number of public repositories not introspected into the last 24 hours
# TYPE content_sources_public_repositories_not_introspected_last_24_hours_total gauge
content_sources_public_repositories_not_introspected_last_24_hours_total 0
# HELP content_sources_public_repositories_with_failed_introspection_total Number of repositories with failed introspection
# TYPE content_sources_public_repositories_with_failed_introspection_total gauge
content_sources_public_repositories_with_failed_introspection_total 0
# HELP content_sources_repositories_total Number of repositories
# TYPE content_sources_repositories_total gauge
content_sources_repositories_total 22
# HELP content_sources_repository_configs_total Number of repository configurations
# TYPE content_sources_repository_configs_total gauge
content_sources_repository_configs_total 0
# HELP go_build_info Build information about the main Go module.
# TYPE go_build_info gauge
go_build_info{checksum="",path="",version=""} 1
# HELP promhttp_metric_handler_errors_total Total number of internal errors encountered by the promhttp metric handler.
# TYPE promhttp_metric_handler_errors_total counter
promhttp_metric_handler_errors_total{cause="encoding"} 0
promhttp_metric_handler_errors_total{cause="gathering"} 0

PR #152, opened 2022-11-30
Fixes 286: 404s not logged by default

avisiedo:
- Add default log level value to info.
- Add LOGGING_LEVEL parameter to the deployment level (we could increase the level in deployment time for ephemeral environment if needed).
- Fix '/ping' and '/ping/' endpoint and update the tests.

TODO:

- [x] Pending to check last time in ephemeral environment (I have to move some commits to the branch I cherry pick that help me on this, which works with bonfire 4.11.0).

### DIFF
--- a/deployments/deployment.yaml
+++ b/deployments/deployment.yaml
@@ -73,6 +73,8 @@ objects:
                   secretKeyRef:
                     name: content-sources-certs
                     key: cdn.redhat.com
+              - name: LOGGING_LEVEL
+                value: ${{LOGGING_LEVEL}}
             resources:
               limits:
                 cpu: ${CPU_LIMIT}
@@ -144,6 +146,8 @@ objects:
                   secretKeyRef:
                     name: content-sources-certs
                     key: cdn.redhat.com
+              - name: LOGGING_LEVEL
+                value: ${{LOGGING_LEVEL}}
             resources:
               limits:
                 cpu: ${CPU_LIMIT}
@@ -174,6 +178,8 @@ objects:
                   secretKeyRef:
                     name: content-sources-certs
                     key: cdn.redhat.com
+              - name: LOGGING_LEVEL
+                value: ${{LOGGING_LEVEL}}
       database:
         name: content-sources
         version: 13
@@ -210,3 +216,6 @@ parameters:
     value: 1Gi
   - name: MEMORY_REQUESTS
     value: 100Mi
+  - name: LOGGING_LEVEL
+    value: "info"
+
--- a/deployments/deployment.yaml
+++ b/deployments/deployment.yaml
@@ -217,5 +217,5 @@ parameters:
   - name: MEMORY_REQUESTS
     value: 100Mi
   - name: LOGGING_LEVEL
-    value: "warn"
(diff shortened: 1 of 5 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-286

pkg/handler/api_test.go
```diff
@@ -30,24 +30,37 @@ func serveRouter(req *http.Request) (int, []byte, error) {
...
+		t.Log(path)
+		req, _ := http.NewRequest("GET", path, nil)
+		code, body, err := serveRouter(req)
+		require.NoError(t, err)
+		assert.Equal(t, http.StatusNotFound, code)
+
+		expected := "{\"errors\":[{\"status\":404,\"detail\":\"Not Found\"}]}\n"
+		assert.Equal(t, expected, string(body))
+		// expected := "{\"message\":\"Not Found\"}\n"
+		// assert.Equal(t, expected, string(body))
```
jlsherrill, line 62: commented code :)

jlsherrill:
I confirmed in ephemeral:
```
{"level":"error","error":"code=404, message=Not Found","remote_ip":"10.128.8.9","host":"env-ephemeral-yejy47-gcn8a3ez.apps.c-rh-c-eph.8p0c.p1.openshiftapps.com","method":"GET","uri":"/api/content-sources/v1/zombies/","user_agent":"curl/7.82.0","status":404,"referer":"","latency":0.013939,"latency_human":"13.939µs","bytes_in":"0","bytes_out":"49","time":"2022-11-30T13:38:36Z","time":"2022-11-30T13:38:36Z","caller":"/go/pkg/mod/github.com/ziflex/lecho/v3@v3.1.0/middleware.go:124"}
```

deployments/deployment.yaml
```diff
@@ -210,3 +216,6 @@ parameters:
     value: 1Gi
   - name: MEMORY_REQUESTS
     value: 100Mi
+  - name: LOGGING_LEVEL
+    value: "info"
```
jlsherrill, line 220: any reason to go with info instead of warn?   Testing locally, it still shows the 404s, and i'd think in a prod environment you'd just want warn, but IDK?

avisiedo: thanks for this!! no reason indeed!! updating the level to warn

pkg/handler/api_test.go
avisiedo: ups! :) thanks!

avisiedo:
There is another situation that I will address in a different ticket; the new situation is that I cannot set the log level to filter messages; I mean, if I set the log level to "error" and the message to print out is "info" level, I would expect the message is not printed out, however, I am getting that message printed out.

Debugging I have seen that at some point, the log level for the logger is "debug", so everything is going to be printed out. I have tried several changes, but nothing gave me the expected result. By the way, the current root cause for this pr is addressed, I will open a new card to address that other situation, as this pr is letting me to deploy into the ephemeral environment.

jlsherrill (approved):

jlsherrill:
feel free to merge!

PR #154, opened 2022-12-12
Fixes 280: switch to using repomd checksum

jlsherrill:
to determine if a repo was updated.

### DIFF
--- a/pkg/external_repos/introspect.go
+++ b/pkg/external_repos/introspect.go
@@ -1,8 +1,10 @@
 package external_repos
 
 import (
+	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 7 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-280

jlsherrill:
To test this, you can use this repository url: https://jlsherrill.fedorapeople.org/fake-repos/revision/current/

You should see one package.  Ask me to 'switch' it and add a 2nd package.  Then introspect it again, or wait ~8 hours.  you should now see 2 packages

Andrewgdewar (approved):
Tested for functionality. Works as intended. 
Code looks good but will defer to @rverdile's opinion in that regard.
Ack

pkg/external_repos/introspect.go
```diff
@@ -84,7 +86,13 @@ func Introspect(repo *dao.Repository, repoDao dao.RepositoryDao, rpm dao.RpmDao)
 		return 0, err
 	}
 
-	if repo.Revision != "" && repomd.Revision != "" && repomd.Revision == repo.Revision {
+	checksumStr := ""
+	if repomd.RepomdString != nil && *repomd.RepomdString != "" {
+		sum := sha256.Sum256([]byte(*repomd.RepomdString))
+		checksumStr = hex.EncodeToString(sum[:])
```
rverdile, line 92: Does this package from the standard library accomplish the same thing? https://pkg.go.dev/encoding/hex#EncodeToString

Also, just curious, why do we need to encode the checksum here?

jlsherrill: doh! thats what i meant to use.  IDE autocomplete for the lose. 

The checksum comes through as a byte array of numbers.  We could just store those.. but converting it to hex matches what you normally see when you compute a sha256 checksum

mshriver (approved):

PR #160, opened 2022-12-20
Refs 269: add more arch detail to docs

jlsherrill:


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-269

jlsherrill:

docs/db-model.puml
```diff
@@ -26,10 +32,12 @@ entity "**repository_configurations**" {
   *""name"": //character varying(255) //
   ""versions"": //character varying(255)[] //
   *""arch"": //character varying(255) //
```
Andrewgdewar, line 34: If the * is supposed to mean required, I don't think arch actually is.

docs/db-model.puml
```diff
@@ -26,10 +32,12 @@ entity "**repository_configurations**" {
...
-  *""account_id"": //character varying(255) //
+  ""account_id"": //character varying(255) //
   *""org_id"": //character varying(255) //
   *""org_id"": //character varying(255) //
   *""repository_uuid"": //uuid [FK]//
+  ""gpg_key"": //text //
+  *""metadata_verification"": //boolean //
```
Andrewgdewar, line 40: I'm unsure if metadata_verification can be undefined as well (if the * is as suggested above).

Andrewgdewar (approved):
Looks good.

docs/db-model.puml
jlsherrill: i believe its required in the db, but not required in the api.   In the DB its stored as empty string if not passed in  (this is auto generated puml)

docs/db-model.puml
jlsherrill: same as above, its required in the db schema, but defaults to false

PR #161, opened 2022-12-21
Fixes 299: add cloudwatch support

jlsherrill:
TODO:

* [x] Integrate with app-interface/clowder

Note, I tested this with cloudwatch directly with:

```
CLOUDWATCH_REGION=us-east-1  \
	CLOUDWATCH_GROUP=ContentSources \
        CLOUDWATCH_STREAM=test \
	CLOUDWATCH_KEY="somekey" \
	CLOUDWATCH_SECRET=somesecret \
	go run cmd/content-sources/main.go api consumer
```

you may need to create a cloudwatch group & stream, as well as a user that has access with the given key and secret.


jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-299

mk/lint.mk
```diff
@@ -1,5 +1,5 @@
 ADD_PYTHON_ENV := source .venv/bin/activate &&
-GOLANGCI_LINT_VERSION := v1.46.2
+GOLANGCI_LINT_VERSION := ""
```
jlsherrill, line 2: updates to install latest golangci-lint

rverdile (commented):
Looks like it works to me. I created a cloudwatch group and stream in my aws account, added the configuration, and then I could see the server requests being logged to the stream.

Anything else to think about here?

jlsherrill:
I don't think so, we'll have to confirm clowdet support is good once this
goes to stage

PR #163, opened 2023-01-05
Fixes 289: Package count accuracy fix

Andrewgdewar:
This fixes an issue seen when two packages were identical but have different names.

In such a case the total count would be incorrect, as the checksum data used for package uniqueness would prevent duplication of the package, but the original package count would still be used after this process was completed. 

This fix makes an actual count of the associated items in the database after the introspection is finished to accurately return the correct number for storage. 

- [x]  Needs tests

### DIFF
--- a/pkg/external_repos/introspect_mocks_test.go
+++ b/pkg/external_repos/introspect_mocks_test.go
@@ -28,6 +28,11 @@ type MockRepositoryDao struct {
 	mock.Mock
 }
 
+func (m *MockRepositoryDao) FetchRepositoryRPMCount(repoUUID string) (int, error) {
+	args := m.Called(repoUUID)
+	return args.Int(0), args.Error(1)
+}
+
 func (m *MockRepositoryDao) List() ([]dao.Repository, error) {
 	return []dao.Repository{}, nil
 }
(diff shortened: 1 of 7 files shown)


pkg/external_repos/introspect_mocks_test.go
```diff
@@ -28,6 +28,10 @@ type MockRepositoryDao struct {
 	mock.Mock
 }
 
+func (m *MockRepositoryDao) FetchRepositoryRPMCount(repoUUID string) (int, error) {
+	return 13, nil
```
Andrewgdewar, line 32: There is only one test currently making use of this. So the return value matches that. 
I likely would want to do something similar here if I had more insight.

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-289

pkg/external_repos/introspect_mocks_test.go
rverdile: I would do `return args.Int(0), args.Error(1)` and then values can be set in the test itself. For example, in [this test](https://github.com/content-services/content-sources-backend/blob/main/pkg/handler/repositories_test.go#L303) there's `.Return(repo, nil)`. In your case, you might do `.Return(13, nil)`.

Andrewgdewar: Cool.

rverdile (approved):
looks good. tested and working. nice!

swadeley:
OK, now it shows just one package. Thank you.

PR #167, opened 2023-01-06
Refs 299: set default log stream

jlsherrill:
to reproduce the error we were seeing in stage, run with: `CLOUDWATCH_STREAM=""`  on main branch.

With this change, if you run with `CLOUDWATCH_STREAM=""`  it will default to a 'default' stream.  Otherwise it will use the specified stream.  This could allow us to customize the stream with env variables in our clowdapp config


rverdile (approved):
ack!

PR #170, opened 2023-01-09
Fixes 326: Filter repositories by status

rverdile:
Adds filtering repository list by status e.g. `/api/content-sources/v1/repositories/?status=Pending,Valid`

To test, create repositories with different introspection statuses and try filtering.

### DIFF
--- a/pkg/dao/repository_configs.go
+++ b/pkg/dao/repository_configs.go
@@ -199,6 +199,11 @@ func (r repositoryConfigDaoImpl) List(
 		filteredDB = filteredDB.Where(orGroup)
 	}
 
+	if filterData.Status != "" {
+		statuses := strings.Split(filterData.Status, ",")
+		filteredDB = filteredDB.Where("status IN ?", statuses)
+	}
+
 	sortMap := map[string]string{
 		"name":                  "name",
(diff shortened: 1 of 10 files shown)


pkg/dao/repository_configs.go
```diff
@@ -199,6 +201,12 @@ func (r repositoryConfigDaoImpl) List(
 		filteredDB = filteredDB.Where(orGroup)
 	}
 
+	if filterData.Status != "" {
+		statusesString := cases.Title(language.Und).String(filterData.Status)
+		statuses := strings.Split(statusesString, ",")
```
rverdile, line 206: I made the filtering not case sensitive. I'm not sure if we want that, but it's easy to remove.

jlsherrill:
https://issues.redhat.com/browse/HMSCONTENT-326

pkg/dao/repository_configs.go
Andrewgdewar: Sounds like we likely don't unfortunately (as discussed during standup).

Andrewgdewar (approved):
Tested both via postman and front-end, working well.

PR #185, opened 2023-01-26
Fixes 1118: cleanup url before bulkcreate

rverdile:
## Summary
During BulkCreate, the repository URL was not being cleaned up (trailing slashes add/removed) before searching for an existing repository. This would cause an error when creating a repository without a trailing slash, if the repository already existed (with a trailing slash) because `FirstOrCreate` would not find the repository.

## Testing steps
1. Create a repository https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/
2. delete the repository
3. Create the repository again, but without a trailing slash: https://jlsherrill.fedorapeople.org/fake-repos/needed-errata

Without changes, this would say that it cannot create repository that already exists.

### DIFF
--- a/pkg/dao/repository_configs.go
+++ b/pkg/dao/repository_configs.go
@@ -119,7 +119,8 @@ func (r repositoryConfigDaoImpl) bulkCreate(tx *gorm.DB, newRepositories []api.R
 		}
 		ApiFieldsToModel(newRepositories[i], &newRepoConfigs[i], &newRepos[i])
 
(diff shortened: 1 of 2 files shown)


jlsherrill (approved):

mshriver:
I've got test coverage for this, and am retesting because there may just be a timeout issue. Addressing that timeout issue directly within the test, but may see green here earlier.

mshriver (approved):

mshriver:
coverage included in E2E case for backend and frontend contexts.

PR #187, opened 2023-01-30
Fixes 1138: use env variable for count grafana panels

rverdile:
## Summary
The grafana panels for the repository counts were using the wrong datasource. I changed the datasource to the environment variable that holds the value of the datasource selected for the whole dashboard (a dropdown in the outer UI).
## Testing steps
none


jlsherrill:
going ahead and merging, no qe needed.

PR #188, opened 2023-01-31
Fixes 548: delete orphaned repos

jlsherrill:
## Summary

as part of introspect-all.  This deletes any non-public repositories that were created more than a week ago and do not have any repo_configs defined

## Testing steps
Create a repo:
```
curl -X POST http://localhost:8000/api/content-sources/v1/repositories/bulk_create/  -d '[{"name":"foobarz", "url":"https://jlsherrill.fedorapeople.org/fake-repos/needed-errataz"}]'  -H "`./scripts/header.sh 3 1`"   -H "Content-Type: application/json"   
```
update the created_at time:
```
make db-cli-connect
```
```
# update repositories  set created_at = LOCALTIMESTAMP - INTERVAL '1 week' where url = 'https://jlsherrill.fedorapeople.org/fake-repos/needed-errataz/';
UPDATE 1
```

Delete the repo (using UUID from create output):
```
curl -X DELETE http://localhost:8000/api/content-sources/v1/repositories/f3430e8e-38b1-4563-84c1-4b93c3e63585  -H "`./scripts/header.sh 3 1`"   -H "Content-Type: application/json"
```

introspect-all
```
go run cmd/external-repos/main.go introspect-all
```

notice:

```
{"level":"debug","time":"2023-01-31T16:30:43-05:00","message":"Cleaned up 1 orphaned repositories"}
```

and in the db, it should be gone:
```
select url from repositories;
```

### DIFF
--- a/pkg/dao/repositories.go
+++ b/pkg/dao/repositories.go
@@ -110,7 +110,6 @@ func (r repositoryDaoImpl) OrphanCleanup() error {
 
 	// Delete orphans
 	tx := r.db.
-		Unscoped().
 		Where("repositories.uuid in (?)", query).
 		Delete(&models.Repository{})
 	if tx.Error != nil {
--- a/pkg/dao/repositories.go
+++ b/pkg/dao/repositories.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	"github.com/content-services/content-sources-backend/pkg/models"
+	"github.com/rs/zerolog/log"
 	"gorm.io/gorm"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/external_repos/introspect.go
```diff
@@ -153,6 +153,10 @@ func IntrospectAll(force bool) (int64, []error) {
 			errors = append(errors, err)
 		}
 	}
+	err = repoDao.OrphanCleanup()
```
rverdile, line 156: What about a small comment here to say that repository orphan cleanup is called here because it's a convenient place for it, but not because it's actually needed for introspection?

rverdile: actually, the more I think about it the more I think it fits here. maybe no explanation necessary!

rverdile (commented):
played around with this and looks good!

I left a couple of comments. I don't think they'll lead to any big changes.  but I guess I'll wait and see before acking.

pkg/external_repos/introspect.go
rverdile: the _more_ I think about this, I realize the same thing could be said for `rpm.OrphanCleanup()`. We just call these in `IntrospectAll()` because the cron job will trigger them, right? `IntrospectURL()` on the other hand does not call `OrphanCleanup()`.

Anyway, I think this is fine. I had just assumed `rpm.OrphanCleanup()` was more directly tied to introspection specifically.

pkg/dao/repositories.go
```diff
@@ -98,6 +99,28 @@ func (p repositoryDaoImpl) Update(repoIn RepositoryUpdate) error {
...
+	query := r.db.Model(&models.Repository{}).
+		Joins("left join repository_configurations on repositories.uuid = repository_configurations.repository_uuid").
+		Where("repository_configurations.uuid is NULL").
+		Where("repositories.public is false").
+		Where("repositories.created_at < (LOCALTIMESTAMP - INTERVAL '1 week' ) ").
+		Select("repositories.uuid")
+
+	// Delete orphans
+	tx := r.db.
+		Unscoped().
```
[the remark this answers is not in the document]
jlsherrill: Thats a good question, i had adapted the rpm orphan cleanup method and it has this.  I assume it was due to the soft delete: https://gorm.io/docs/delete.html#Delete-permanently  
but since we don't have a deleted_at field, i'm not sure why we need it.  I'll delete from both

pkg/external_repos/introspect.go
jlsherrill: sounds good.  Yes, i could see moving those two calls to a cleanup method might make it more clear, but maybe not necessary at this point

rverdile (approved):
lgtm!

PR #221, opened 2023-02-23
Refs 533: Add repository introspection endpoint

rverdile:
## Summary
This will also have a frontend PR, but I'll put this up as a draft in the meantime for judgement.

- Adds `/repository/{uuid}/introspect` endpoint to introspect a repository associated with the given repository configuration UUID.
- Adds `failed_introspection_count` to the `repositories` table, which represents the number of consecutive failed introspections and is used to skip introspection if a certain limit is reached.
- Adds a `reset_count` boolean param to this endpoint that will reset the `failed_introspections_count` column back to zero.
- Changed format for timestamps returned by API to be RFC3339.
## Testing steps
1. Migrate DB
2. Set the `OPTIONS_INTROSPECT_API_TIME_LIMIT_SEC` environment variable to 10. This should make it should you cannot introspect a repository more than once every 10 seconds. You should also test setting this  option in config.yaml, `options.introspect_api_time_limit_sec`.
3. Introspect an invalid repository with the endpoint, without passing the reset_count param. The failed introspections count should increase. You can verify this by fetching the repository from its GET endpoint.
4. Introspect an invalid repository and pass the reset_count param. The failed introspections count should become 1 after introspection fails.
6. Introspect a valid repository that was failing previously. The failed introspections count should reset to 0.
7. Introspect a repository over the FailedIntrospectionsLimit (20), using the CLI, without using the `--force` flag. It should be skipped. Doing that again, with the `--force` flag, should return an error.


jlsherrill:
https://issues.redhat.com/browse/HMS-533

jlsherrill:

jlsherrill:
we'll need to make sure that the metrics from here https://github.com/content-services/content-sources-backend/pull/220/files#diff-f171c549d2735774b148a59f7f39861d8e58abe0a6f81735ea7cf933749c72b0R53-R77  get updated depending on which pr goes in first.

pkg/external_repos/introspect.go
```diff
@@ -73,6 +74,10 @@ func Introspect(repo *dao.Repository, repoDao dao.RepositoryDao, rpm dao.RpmDao)
 	)
 	log.Debug().Msg("Introspecting " + repo.URL)
 
+	if repo.FailedIntrospectionsCount > FailedIntrospectionsLimit {
+		return 0, fmt.Errorf("introspection skipped because this repository has failed more than %v times in a row", FailedIntrospectionsLimit)
+	}
```
jlsherrill, line 79: we probably only want to do this if its not a 'public' repository, i think we'd want those to keep trying

pkg/test/mocks/repository_config_dao.go
```diff
@@ -84,3 +86,12 @@ func (r *RepositoryConfigDao) ValidateParameters(orgId string, req api.Repositor
 	r.Called(orgId, req)
 	return api.RepositoryValidationResponse{}, nil
 }
+
+func NewRepositoryConfigDao(t *testing.T) *RepositoryConfigDao {
+	m := &RepositoryConfigDao{}
+	m.Mock.Test(t)
+
+	t.Cleanup(func() { m.AssertExpectations(t) })
```
rverdile, line 94: This makes it so we don't have to call AssertExpectations in our handler tests, which in a couple of test cases we were forgetting to do. `Cleanup(func())` is run at the end of the test, but before it ends.

The mockery generator was already doing this. I just added it here for repository configs.

pkg/handler/repositories.go
```diff
@@ -318,3 +326,57 @@ func (rh *RepositoryHandler) deleteRepository(c echo.Context) error {
...
+	repo, err := rh.RepositoryDao.FetchForUrl(response.URL)
+	if err != nil {
+		return ce.NewErrorResponse(ce.HttpCodeForDaoError(err), "Error fetching repository uuid", err.Error())
+	}
+
+	limit := time.Second * time.Duration(config.Get().Options.IntrospectApiTimeLimitSec)
+	since := time.Since(*repo.LastIntrospectionTime)
+	if since < limit {
+		detail := fmt.Sprintf("This repository has been introspected recently. Try again in %v", (limit - since).Truncate(time.Second))
+		return ce.NewErrorResponse(http.StatusBadRequest, "Error introspecting repository", detail)
```
rverdile, line 364: I'm treating this time limit like an error, but maybe we just want it to quietly skip introspection?

jlsherrill: I think a 400 is fine here.  We'd probably want to prevent it in the UI, so a normal user would never see it.

db/migrations/20230214142349_add_count_to_repositories.up.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+alter table repositories add column failed_introspections_count int default 0 not null;
+
+COMMIT;
```
jlsherrill, line 5: apparently there's a migration dated after this one (2-15), and our db migrator won't actually migrate this one if that the later one is already applied.  I guess it doesn't support out-of-order migrations like rails does.  So locally this never got applied, but renaming it to 20230216...  fixed it.

pkg/external_repos/introspect.go
```diff
@@ -49,6 +50,10 @@ func Introspect(repo *dao.Repository, repoDao dao.RepositoryDao, rpm dao.RpmDao)
 	)
 	log.Debug().Msg("Introspecting " + repo.URL)
 
+	if repo.FailedIntrospectionsCount > FailedIntrospectionsLimit && !repo.Public {
+		return 0, fmt.Errorf("introspection skipped because this repository has failed more than %v times in a row", FailedIntrospectionsLimit)
+	}
```
jlsherrill, line 55: During the nightly introspection, I /think/ these shouldn't necessarily be treated as errors, but instead should just be ignored.   That said, if introspect() is explicitly called on it, we do probably want to error i think?  So my suggestion would be to update the query that fetches the list of repos for introspection to filter these out automatically.

rverdile:
I haven't addressed your comments yet. This latest push is for the Retrying status addition.

pkg/external_repos/introspect.go
rverdile: I think my idea here was I wanted the UI to show "this repository is failing too many times so introspection has been stopped". But I guess I can communicate that by checking the count first, if that makes sense?

jlsherrill:
i *think* you need to add an rbac entry here: https://github.com/content-services/content-sources-backend/blob/main/pkg/middleware/rbac_map.go#L22-L35

I filed this task to try to improve this:  https://issues.redhat.com/browse/HMS-1486

jlsherrill:
Oh and we also need to adjust https://github.com/content-services/content-sources-backend/blob/main/pkg/dao/metrics.go#L59-L89  to ignore repos that are beyond the 'failed count', as we don't want to include them in the stats

pkg/handler/repositories.go
```diff
@@ -318,3 +325,70 @@ func (rh *RepositoryHandler) deleteRepository(c echo.Context) error {
...
+// @summary 		introspect a repository
+// @ID				introspect
+// @Tags			repositories
+// @Param  			uuid            path    string  true  "Identifier of the Repository"
+// @Param			reset_count     query   bool    false "Reset the failed introspections count"
+// @Success			204 "Introspection was successfully queued"
+// @Failure      	400 {object} ce.ErrorResponse
+// @Failure      	404 {object} ce.ErrorResponse
+// @Failure      	500 {object} ce.ErrorResponse
+// @Router			/repositories/{uuid}/introspect [post]
```
jlsherrill, line 339: the router here shows /introspect, but the router above actually only accepts /introspect/

pkg/handler/repositories.go
```diff
@@ -318,3 +325,70 @@ func (rh *RepositoryHandler) deleteRepository(c echo.Context) error {
 	}
 	return c.NoContent(http.StatusNoContent)
 }
+
+// IntrospectRepository godoc
+// @summary 		introspect a repository
+// @ID				introspect
+// @Tags			repositories
+// @Param  			uuid            path    string  true  "Identifier of the Repository"
+// @Param			reset_count     query   bool    false "Reset the failed introspections count"
```
jlsherrill, line 334: any reason this is a query param instead of part of the body?

rverdile: I guess not. I'll make it part of the body,

pkg/dao/repositories.go
```diff
@@ -66,11 +71,17 @@ func (p repositoryDaoImpl) FetchForUrl(url string) (Repository, error) {
...
-func (p repositoryDaoImpl) List() ([]Repository, error) {
+func (p repositoryDaoImpl) List(ignoreFailed bool) ([]Repository, error) {
 	var dbRepos []models.Repository
 	var repos []Repository
 	var repo Repository
-	result := p.db.Find(&dbRepos)
+	var result *gorm.DB
+
+	if ignoreFailed {
+		result = p.db.Where("failed_introspections_count < ?", config.FailedIntrospectionsLimit+1).Find(&dbRepos)
```
rverdile, line 81: I made this  `config.FailedIntrospectionsLimit+1` so that it errors the first time the limit is hit, but after that it skips.

rverdile:
From [this discussion](https://github.com/content-services/content-sources-frontend/pull/104#discussion_r1154611700) on the front PR, I'm thinking it may have been unnecessary for me to add a 'Retrying' status.

I don't think I had another reason for adding it. I can't think of one at the moment, other than maybe it will be useful at some point to handle them separately. If it makes sense to remove it, I'll remove it.

pkg/dao/repository_configs.go
```diff
@@ -345,15 +347,16 @@ func ModelToApiFields(repoConfig models.RepositoryConfiguration, apiRepo *api.Re
...
-		apiRepo.LastIntrospectionTime = repoConfig.Repository.LastIntrospectionTime.String()
+		apiRepo.LastIntrospectionTime = repoConfig.Repository.LastIntrospectionTime.Format(time.RFC3339)
 	}
 	if repoConfig.Repository.LastIntrospectionSuccessTime != nil {
-		apiRepo.LastIntrospectionSuccessTime = repoConfig.Repository.LastIntrospectionSuccessTime.String()
+		apiRepo.LastIntrospectionSuccessTime = repoConfig.Repository.LastIntrospectionSuccessTime.Format(time.RFC3339)
 	}
 	if repoConfig.Repository.LastIntrospectionUpdateTime != nil {
-		apiRepo.LastIntrospectionUpdateTime = repoConfig.Repository.LastIntrospectionUpdateTime.String()
+		apiRepo.LastIntrospectionUpdateTime = repoConfig.Repository.LastIntrospectionUpdateTime.Format(time.RFC3339)
```
rverdile, line 359: Small change: changed the format of the timestamps in the API response to be the RFC3339 standard. Still okay to test PR.

swadeley:
Hi

I see the failed introspections count  (introspections, plural):

```
Out[3]: 
{'data': [{'account_id': '0369233',
           'distribution_arch': 'any',
           'distribution_versions': ['any'],
           'failed_introspections_count': 2,
           'gpg_key': '',
           'last_introspection_error': 'Cannot fetch '
                                       'https://fixtures.pulpproject.org/rpm-invalid-rpm/repodata/repomd.xml: '
                                       '404',
           'last_introspection_time': '2023-04-05 16:04:47.668374 +0000 UTC',
           'last_success_introspection_time': '',
           'last_update_introspection_time': '',
           'metadata_verification': False,
           'name': 'rpm-invalid-rpm',
           'org_id': '3340851',
           'package_count': 0,
           'status': 'Invalid',
           'url': 'https://fixtures.pulpproject.org/rpm-invalid-rpm/',
           'uuid': '1a0a8537-d286-4612-9209-5a852b1cf4c5'}],
```

swadeley:
Hi,
 the code says `"/repositories/{uuid}/introspect/"` not `/repository/{uuid}/introspect ` as per comment 0

Anyway,

I had :
`last_introspection_time': '2023-04-05 16:04:47.668374 +0000 UTC',`

and this:
`In [8]: app.content_sources.rest_client.repositories_api.introspect('1a0a8537-d286-4612-9209-5a852b1cf4c5')
2023-04-05 19:55:02.624 [    INFO] [iqe.base.rest_client] REST: METHOD=POST, request_id=TEObNNjI64UjTHSwJkJuhxNzQve6nOMk, params=[]`

Gave me:

` 'last_introspection_time': '2023-04-05 18:55:02.786693 +0000 UTC',`

PR #222, opened 2023-02-28
Fixes 1358: set rbac timeout properly

jlsherrill:
## Summary

I'm about 75% sure this is the problem in stage, as the config value wasn't properly being converted to 
seconds in the main usage (but was as part of metrics).

## Testing steps
Tests pass
Works against mock rbac service


jlsherrill:
this isn't actually needed, but i'm not sure what is

PR #227, opened 2023-03-08
Refs 1203: run instrumentation on consumers

jlsherrill:
If its not running, then we aren't listening on port 9000 for metrics


jlsherrill:
https://issues.redhat.com/browse/HMS-1203

rverdile (approved):
ack

PR #229, opened 2023-03-13
Fixes 1426: change introspection check to 23 hrs

jlsherrill:
so if there is slight variation, we will still introspect ever 24 hours


jlsherrill:
https://issues.redhat.com/browse/HMS-1426

Andrewgdewar (approved):
aack!

PR #230, opened 2023-03-14
Refs 1203: Rename kafka metrics

jlsherrill:
So that we can reuse them when we switch to a newer tasking system


jlsherrill:
https://issues.redhat.com/browse/HMS-1203

rverdile (approved):
ack

PR #234, opened 2023-03-22
Fixes 1440: update dependencies

rverdile:
## Summary
There is a change here: https://github.com/go-gorm/postgres (which was causing our tests to fail), so I had to pin it to the previous version (v1.4.8). There's an issue open for it, so hopefully it gets resolved.

It's causing a bug for us because gorm's error was no longer returning as type `pgconn.PgError`, but `errors.errorString` instead, which is a problem here: https://github.com/content-services/content-sources-backend/blob/main/pkg/dao/repository_configs.go#L31
## Testing steps
none


pkg/dao/repository_configs.go
```diff
@@ -11,7 +11,7 @@ import (
 	ce "github.com/content-services/content-sources-backend/pkg/errors"
 	"github.com/content-services/content-sources-backend/pkg/models"
 	"github.com/content-services/yummy/pkg/yum"
-	"github.com/jackc/pgconn"
+	"github.com/jackc/pgx/v5/pgconn"
```
rverdile, line 14: this I changed because the newer pgx version (v5) uses this package, instead of the independent repository used by pgx v4. It will cause an error in our tests if you use the previous import.

jlsherrill:
https://issues.redhat.com/browse/HMS-1440

jlsherrill (approved):

PR #237, opened 2023-03-24
Fixes 1507: use deprecated security proto config

jlsherrill:
## Summary
https://github.com/content-services/content-sources-backend/commit/3f26de3b5798caf97e2c20e7d4c7bf48d7562cf6 

moved us away from using a now deprecated clowder setting, but it turns out the non-deprecated setting isn't actually populated.  This now checks and tries to use either if they are populated.

## Testing steps
Deploy to stage.


rverdile (approved):
ack

PR #241, opened 2023-04-03
Fixes 1564: Update migration to prevent status ui bug

Andrewgdewar:
## Summary
This fixes an issue where in dev/ephemeral environments adding a "Popular Repository" would cause the API to return a Status of "" (empty string). 
This prevented the UI from updating the repository list the frontend looks at a repository's "Status" to determine whether or not to poll the API.

## Testing steps
Add a popular repository that needs to introspect.
One you haven't added before, or clean out your database.

After adding quickly switch to the repository list view you should see a blank status.

Result after fixing: 
It should say Pending, or any state, if introspection has finished.


jlsherrill:
https://issues.redhat.com/browse/HMS-1564

jlsherrill (approved):
ACK, normally we wouldn't update a migration, but in this case it actually makes sense.  The problem seen here would never actually be seen in an existing running environment, only for new environments

PR #248, opened 2023-04-10
Fixes 1468: add repo snapshot task

jlsherrill:
## Summary
This adds a task to snapshot a repository.  It is meant to recover from failures due to interruptions after:
* remote creation
* repo creation

but not due to interuptions during:
* repo sync
* publication creation
* distribution creation

This resiliency will have to be added during integration with the tasking system.

Other changes:
* Moves dao mocks out of test/mocks into the dao package.  This makes more sense from a dependency perspective
* introduces a Dao registry, to simplify creating and passing around these dao objects, especially in tests.
* Adds a snapshot dao with creation supported

## Testing steps
create a repository via ui or api:

```
POST http://{{host}}/api/content-sources/v1.0/repositories/
x-rh-identity: eyJpZGVudGl0eSI6eyJ0eXBlIjoiQXNzb2NpYXRlIiwiYWNjb3VudF9udW1iZXIiOiIxIiwiaW50ZXJuYWwiOnsib3JnX2lkIjoiMSJ9fX0K

{
  "name": "needed",
  "url": "https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/",
  "distribution_versions": [
    "8"
  ],
  "distribution_arch": "x86_64"
}
```

run the 'task' manually:
```
go run cmd/external-repos/main.go pulp-create https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/ 1
```

Confirm that a distribution was created:
```
curl http://localhost:8080/pulp/api/v3/distributions/rpm/rpm/  -u admin:password | jq
```

and you can fetch from it:

```
curl http://localhost:8080/pulp/content/a63e6d0a-9df9-4553-8eaa-e3c19180e15d/a4300e93-4b62-4704-aba5-13bfae95f593/repodata/repomd.xml
```

## TODO:
* [x] Integration tests
* [x] Down migration
* [ ]


jlsherrill:
https://issues.redhat.com/browse/HMS-1468

rverdile:
+1 to the dao registry change

pkg/dao/snapshot_model.go
```diff
@@ -0,0 +1,47 @@
...
+
+	var counts ContentCounts
+	err := json.Unmarshal(source, &counts)
+	if err != nil {
+		return err
+	}
+
+	*cc = counts
+	return nil
+}
```
jlsherrill, line 47: note, that i moved the 'model' here.  I think longer term it makes sense to move all models to the dao layer, as it could reduce some of the boilerplate code.  We are using models really just as simple holders of data, so having it in a separate package isn't really providing much value, and i think making these more complex.  I'd consider this somewhat of a trial as we implement more, as i'm not 100% sold on it.  Curious what others think

rverdile: I agree. It makes `repository dao` in particular a lot more complex. The only packages that import `models` are `seeds` and `dao`, so it isn't even that big of a change. 

Would you also want to combine `seeds` in as well? IIRC `dao` tests imports seeds (for repository configs), so there would have to be some changes there to prevent a circular dependency.

pkg/pulp_client/task.go
```diff
@@ -0,0 +1,128 @@
...
+	FAILED    string = "failed"
+)
+
+// GetTask Fetch a pulp task
+func (r pulpDaoImpl) GetTask(taskHref string) (zest.TaskResponse, error) {
+	task, _, err := r.client.TasksApi.TasksRead(r.ctx, taskHref).Execute()
+
+	if err != nil {
+		return zest.TaskResponse{}, err
+	}
```
rverdile, line 29: I haven't checked zest, but does this or anything similar block while waiting for a response? And if so would it timeout if it takes too long? Either here or on the zest side of things.

rverdile (commented):
I was trying to see what happens when things go wrong and I might be seeing an issue with creating a snapshot, i.e. `pulp-create`, if it fails to create previously. I don't know if it's an issue here or a pulp issue.

For example,
 1.  `pulp-create https://fixtures.pulpproject.org/rpm-unsigned/  1`, for example 
 2. `ctrl+c` exit the command mid-process 
 3. `pulp-create` again,  it tells me "successfully created". However, when I navigate to `http://localhost:8080/pulp/content` I don't actually see the content there. 

I think I saw similar behavior hard-coding a panic in `PollTask` or an error `SyncRpmRepository`, such that it returns an error.

jlsherrill:
Right now that is expected.  What we're gonna need to do is save the task ids on the task itself somehow (to be done during integration) and then as part of the task check that the sync task/publish task is completed, and if not retry it.   

Since this isn't being done today, what happens is the sync finishes, the next run happens and sees that nothing has changed so no new snapshot is created.  With the above changes, it'd basically pickup the previous pulp task and continue with it.

pkg/pulp_client/task.go
```diff
@@ -0,0 +1,115 @@
...
+
+	return *task, nil
+}
+
+// PollTask Poll a task and return the final task object
+func (r pulpDaoImpl) PollTask(taskHref string) (*zest.TaskResponse, error) {
+	var task zest.TaskResponse
+	var err error
+	syncInProgress := true
```
rverdile, line 38: Is this still representative of just "syncInProgress" or did it evolve into "taskInProgress"? I see publication and distribution also call `PollTask`, so I'm guessing the meaning has changed. If that's true I would rename it.

rverdile (commented):
This is looking good to me. 

Based on the TODO comments, it seems like the gaps are meant to be addressed when integrating with the tasking system. I think it's doing everything it's supposed to in its current state. I think the way it's organized will also make it easy to integrate with the tasking system, so that's a big plus.

I'll think about this a little longer, and I did leave the one small comment, but so far I haven't found anything missing that's not addressed by a TODO comment.

test/integration/snapshot_test.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func TestSnapshotSuite(t *testing.T) {
+	suite.Run(t, new(SnapshotSuite))
+}
+
+func (s *SnapshotSuite) TestSnapshot() {
+	log.Logger.Error().Msgf("USERNAME: %v", config.Get().Clients.Pulp.Username)
+	log.Logger.Error().Msgf("Password: %v", config.Get().Clients.Pulp.Password)
+	s.dao = dao.GetDaoRegistry(s.db)
```
rverdile, line 34:
```suggestion
	s.dao = dao.GetDaoRegistry(s.tx)
```
one more thing :)

jlsherrill: good catch!

rverdile (approved):

PR #253, opened 2023-04-19
Fixes 1447: add postgres tasking system

rverdile:
I'll update this with a bunch of information and a todo.
## Testing steps
1. Still working on a few things
2. If you add `new_tasking_system: True` to your config.yaml, it will use this for introspect instead of kafka.
3. Should work right now for all endpoints that trigger introspect (`create`, `update`, introspect`).
4. For now, if you want to test the tasking system more thoroughly I suggest modifying the introspect handler to do other things, then trigger it via the introspect endpoint.


jlsherrill:
https://issues.redhat.com/browse/HMS-1447

pkg/handler/repositories.go
```diff
@@ -201,13 +222,25 @@ func (rh *RepositoryHandler) bulkCreateRepositories(c echo.Context) error {
...
+			}
+		} else {
+			if msg, err = adapter.NewIntrospect().FromRepositoryResponse(&repo); err != nil {
+				log.Error().Msgf("bulkCreateRepositories could not map to IntrospectRequest message: %s", err.Error())
+				continue
+			}
+			if err = rh.IntrospectRequestProducer.Produce(c, msg); err != nil {
+				log.Error().Msgf("bulkCreateRepositories returned an error: %s", err.Error())
+				continue
+			}
```
jlsherrill, line 243: I think you could extract this big block into a helper function right? it looks quite a bit duplicated from the non-bulk call

jlsherrill: and update :)

pkg/tasks/introspect.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+const Introspect = "introspect"
+
+type IntrospectPayload struct {
+	Url string
+}
+
+// TODO possibly remove context arg
+func IntrospectHandler(ctx context.Context, task *queue.TaskInfo) error {
```
jlsherrill, line 21: we might need to discuss a bit about how we'll handle the request-id header to know if we can remove the context or not, something we haven't discussed before

pkg/handler/repositories.go
rverdile: good idea!

jlsherrill:
could you add a README in the docs directory detailing the arch of the tasking system and explaining its features?

pkg/tasks/worker/worker_pool.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
+)
+
+type TaskHandler func(ctx context.Context, task *queue.TaskInfo) error
+
+type TaskWorkerPool interface {
+	// StartWorkers Starts workers up to number numWorkers defined in config.
+	// Should be run as a go routine.
+	StartWorkers()
```
rverdile, line 19: `TaskWorkerPool` is the interface used to interface with the workers. It represents a group of workers, where the number of workers is configurable, and each worker processes one task at a time.

I separated `worker` from `WorkerPool` to potentially allow more control over the individual worker. I had an idea to restart the worker automatically if it panicked, but I could not get that to work as expected. My other concern with automatically restarting workers is that they'll just continue to panic. Instead, I tried to write the worker loop such that panics are caught and the worker doesn't exit.

pkg/tasks/worker/worker.go
```diff
@@ -0,0 +1,134 @@
...
+				continue
+			}
+			if taskInfo != nil {
+				w.metrics.RecordMessageLatency(taskInfo.Queued)
+				taskId = taskInfo.Id
+				taskToken = taskInfo.Token
+				go w.process(ctx, taskInfo)
+			}
+		case <-beat.C:
+			w.queue.RefreshHeartbeat(taskToken)
```
rverdile, line 87: You'll see here the worker is periodically refreshing the task's heartbeat. A drawback is this assumes the task handler never hangs. Because the task handler is run in a goroutine, if the task handler hangs, the worker will keep refreshing the heartbeat, but the task will never finish. My thought is, since we're writing the task handlers, we might be able to make this assumption.

If we don't want to assume that the task handler never hangs,  I think the alternative to this is having the task handler refresh the task's heartbeat. In this case, we might want two heartbeats, one for the task and one for the worker (so we also know the worker isn't hanging).

Not sure.

pkg/tasks/worker/worker.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+	}
+}
+
+func (w *worker) fetch(ctx context.Context) (*queue.TaskInfo, error) {
+	defer recoverOnPanic(w.logger)
+	return w.queue.Dequeue(ctx, w.taskTypes)
+}
```
rverdile, line 95: by wrapping queue.Dequeue like this and deferring `recoverOnPanic`, queue.Dequeue can panic without exiting the worker.

pkg/tasks/worker/worker.go
```diff
@@ -0,0 +1,134 @@
...
+
+	beat := time.NewTimer(w.heartbeatInterval / 3)
+	defer beat.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			if taskId != uuid.Nil {
+				err := w.requeue(taskId)
+				if err != nil {
```
rverdile, line 66: I made workers requeue in-progress tasks on exit. I think the other option was wait for tasks to finish (which does not work for syncing). This could also be an option passed to the worker (requeue or wait). My thought was requeue is good enough.

rverdile (commented):
Leaving some comments to give context to some details and draw attention to cases we might prefer to handle differently.

pkg/config/config.go
```diff
@@ -17,17 +17,18 @@ import (
...
+	Loaded           bool
+	Certs            Certs
+	Options          Options
+	Kafka            event.KafkaConfig
+	Cloudwatch       Cloudwatch
+	Metrics          Metrics
+	Clients          Clients `mapstructure:"clients"`
+	Mocks            Mocks   `mapstructure:"mocks"`
+	Sentry           Sentry  `mapstructure:"sentry"`
+	NewTaskingSystem bool    `mapstructure:"new_tasking_system"`
```
jlsherrill, line 31: you have to set a 'default' below for this or else it won't work as an env variable... its super annoying :(

rverdile:
I was trying to add integration/queue tests, but it's actually a lot to think about because the database connection works differently for `pgqueue.go`, and so decent change probably has to be made there to pass a transaction through to it.

I've pushed up everything else, which should be good to go.

pkg/api/repositories.go
```diff
@@ -18,6 +18,7 @@ type RepositoryResponse struct {
 	Status                       string   `json:"status"`                              // Status of repository introspection (Valid, Invalid, Unavailable, Pending)
 	GpgKey                       string   `json:"gpg_key"`                             // GPG key for repository
 	MetadataVerification         bool     `json:"metadata_verification"`               // Verify packages
+	RepositoryUUID               string   `json:"-" swaggerignore:"true"`              // UUID of the dao.Repository
```
jlsherrill, line 21: what was the reasoning behind this?

jlsherrill: i.e. why does it need to be exposed to the user?

pkg/tasks/introspect.go
jlsherrill: I'll resolve this for now, and open a task to cover it

pkg/tasks/worker/worker.go
jlsherrill: maybe we should chat about this in person at standup

pkg/external_repos/pulp_external_repos/pulp.go
```diff
@@ -1,4 +1,4 @@
-package external_repos
+package pulp_external_repos
```
rverdile, line 1: I moved this to a new package. I don't have a good solution to the cyclic import here. But I think this command is kinda temporary anyway?

pkg/tasks/queue/pgqueue.go
```diff
@@ -87,27 +89,37 @@ const (
...
+                VALUES ($1, $2, statement_timestamp())`
 	sqlQueryHeartbeats = `
                 SELECT token
                 FROM task_heartbeats
-                WHERE age(now(), heartbeat) > $1`
+				WHERE age(statement_timestamp(), heartbeat) > $1`
 	sqlRefreshHeartbeat = `
                 UPDATE task_heartbeats
-                SET heartbeat = now()
+                SET heartbeat = statement_timestamp()
```
rverdile, line 99: for a transaction, `now()` returns the start of the transaction, so I changed this to `statement_timestamp()` which I think would be close enough.

pkg/tasks/queue/pgqueue.go
```diff
@@ -231,11 +244,9 @@ func (q *PgQueue) waitAndNotify(ctx context.Context) error {
 }
 
 func (p *PgQueue) Enqueue(task *Task) (uuid.UUID, error) {
-	conn, err := p.pool.Acquire(context.Background())
-	if err != nil {
-		return uuid.Nil, fmt.Errorf("error connecting to database: %v", err)
-	}
-	defer conn.Release()
```
rverdile: I believe it is okay to remove these `pool.Acquire` because `pool` methods acquire a connection implicitly ([example](https://github.com/jackc/pgx/blob/v4.18.1/pgxpool/pool.go#L593)). Doing this enables a `pool` or a `tx` to be used interchangeably through the `Connection` interface.

pkg/external_repos/pulp_external_repos/pulp.go
jlsherrill: agreed, i think this is fine.  We'll likely delete this file as part of the integration work

pkg/tasks/queue/pgqueue.go
jlsherrill: make sense 👍🏻

jlsherrill:
Overall looks good, thoughts on incorporating these metrics into the tasking system?  https://github.com/content-services/content-sources-backend/blob/main/pkg/instrumentation/metrics.go#L105-L117

to capture latency (time from queued and then picked up) and tasking failures (failures with processing, not with external factors like failed repo syncs)

rverdile:
Those should be there already. Are they working?

Latency: https://github.com/content-services/content-sources-backend/blob/68e92b7975ae5518c5dc237b9b2097e75ce7c599/pkg/tasks/worker/worker.go#L81

Result: https://github.com/content-services/content-sources-backend/blob/68e92b7975ae5518c5dc237b9b2097e75ce7c599/pkg/tasks/worker/worker.go#L107-L110

jlsherrill:
ohhh!  nice work, i overlooked it, let me check!

jlsherrill:
yes, working great!!

jlsherrill (approved):

jlsherrill:
[test]

PR #255, opened 2023-04-21
Update e2e test template to pull testing image every time

tpapaioa:
## Summary
This PR updates the e2e test template to use an `imagePullPolicy` of `Always` for the testing container image. This is required because we always use the same `hms-integration` image tag to point to the latest build, and by default tags other than `latest` are not re-pulled.


swadeley (approved):
ACK

PR #256, opened 2023-04-21
Move imagePullPolicy to container level in e2e job template

tpapaioa:
## Summary
The previous PR added `imagePullPolicy` at the wrong level in the e2e job template.


swadeley (approved):
ACK

PR #261, opened 2023-04-28
Fixes 1672: support multiple gpg keys

rverdile:
## Summary
Previously, if there were multiple keys in a gpg key file, only the first key was read. Now multiple keys can be read.

## Testing steps
1. In UI, create this repo: https://dl.google.com/linux/chrome/rpm/stable/x86_64
2. With this gpg key: https://dl.google.com/linux/linux_signing_key.pub
3. With metadata verification set to true
4. Should not return a validation error. Previously, would return the error `Error validating signature: openpgp: signature made by unknown entity. Is this the correct GPG Key?`


jlsherrill:
https://issues.redhat.com/browse/HMS-1672

pkg/dao/repository_configs.go
```diff
@@ -529,8 +529,47 @@ func (r repositoryConfigDaoImpl) checkSignaturePresent(request *api.RepositoryVa
...
+		if val == -1 {
+			break
+		}
+		gpgKeys = append(gpgKeys, gpgKeyCopy[:val+len(EndGpgKey)])
+		gpgKeyCopy = gpgKeyCopy[val+len(EndGpgKey):]
+	}
+
+	if len(gpgKeys) == 0 {
+		return nil, fmt.Errorf("no gpg key was found")
+	}
```
rverdile, line 571: this error check needed to be added, or an invalid gpg key could be considered valid.

rverdile: Any key without `"-----END PGP PUBLIC KEY BLOCK-----"`. For example: "a"

mayurilahane:
Just deployed the pod with these changes in it.
And as mentioned:

_"Should not return a validation error. Previously, would return the error Error validating signature: openpgp: signature made by an unknown entity. Is this the correct GPG Key?"_

It is **not** giving any error for this url and gpg key with metadata verification true

mayurilahane:
Screenshot from 2023-04-28 17-37-05

pkg/dao/repository_configs.go
mayurilahane: facing this https://issues.redhat.com/browse/HMS-1700

if i tried to add a key  as "a" .

PR #264, opened 2023-05-08
Fixes 1735: add rbac caching w/ redis

jlsherrill:
## Summary
This adds caching support for rbac requests, with the goal to reduce overall latency from users using the application.

This uses a 1 minute cache expiration by default.  

## Testing steps
A couple of ways to test
1) spin up an ephemeral env with these changes (using custom deployment.yaml from custom bonfire config and pr built image)
2) curl the repositories api ```time curl https://env-ephemeral-pzuqlw-iiajlc06.apps.c-rh-c-eph.8p0c.p1.openshiftapps.com/api/content-sources/v1/repositories/  -u jdoe:CiUhwxHyzE2K87zz ```
3)  immediately curl again. You should see 100-200ms drop in request time
4) wait 60 seconds and curl again.  You should it revert back initially to the higher request time.

Second way (locally in dev):
1) add new config options from config.example.yaml
2) `make compose-up`  (again to get a new redis locally)
2) Turn on rbac:
```
clients:
  rbac_enabled: True
  rbac_base_url: http://localhost:8800/api/rbac/v1
```
3) add a sleep statement to your rbac client:  clients/rbac.go right above 
```
		acl, err = r.client.GetAccess(ctx, identity.GetIdentityHeader(ctx), "")
```
add:

```
time.sleep(5*time.Second)
```
4) ```curl http://localhost:8000/api/content-sources/v1.0/repositories/ -H "`./scripts/header.sh 1 foo`"```  should take 5 seconds, while immediately running it again should take much less time.

TODO: 
* [x] verify behavior for no access from rbac (error or empty list?)
* [x] Testing around this is currently minimal, and i think the ephemeral tests are probably the best tests rather than trying to mock behavior.  I did add some tests around the rbac client to ensure its respecting the configured cache


pkg/client/rbac.go
```diff
@@ -64,24 +68,45 @@ func NewRbac(baseUrl string, timeout time.Duration) Rbac {
...
+			log.Logger.Err(err).Msg("Cache error")
+		}
+	}
+	if !cacheHit {
+		ctx, cancel := context.WithTimeout(ctx, r.timeout)
+		defer cancel()
+
+		acl, err = r.client.GetAccess(ctx, identity.GetIdentityHeader(ctx), "")
+		if err != nil {
+			// TODO check for 400?
```
jlsherrill, line 102: i remember there being some weirdness around this where an err may mean no access versus having an empty list.  I'll investigate

deployments/deployment.yaml
```diff
@@ -213,6 +213,7 @@ objects:
       database:
         name: content-sources
         version: 13
+      inMemoryDb: true
```
Andrewgdewar, line 216: Does this give you a elasticache?

jlsherrill: in ephemeral its a redis instance i believe, but in stage/prod it'll be elasticache.  I believe there is an app-interface change i'll need to make after merge to make it actually available in stage

Andrewgdewar (approved):
Tested working. We should likely get the mentioned need changes in appSRE before merging.

jlsherrill:
The needed stage changes are now available in app-interface

PR #265, opened 2023-05-08
Fixes 1759: add 9.2 repos

jlsherrill:
## Summary
Image builder will support rhel 9.2 soon.  We need to add repos to support them.

## Testing steps
can't test now, as these repos aren't live


swadeley (approved):
lgtm

PR #269, opened 2023-05-11
Fixes 602: Send create notification

Andrewgdewar:
## Summary

Initial test of the notifications service integration. 

This current PR only implements the "BulkCreate" API.

A follow up PR will implement the rest of the notifications required after this pattern is proven.

## Testing steps

To test locally:  

- Navigate to "/pkg/notifications/client.go"

- comment out lines 19 - 25, and the closing brace on line 67

- Replace the "kafkaServers" on line 32 with the following:
Screen Shot 2023-05-11 at 10 24 36 AM

- Install/run [KafkaCat](https://docs.confluent.io/platform/current/clients/kafkacat-usage.html#consumer-mode) so that you can record the message when it is sent.
- Call bulkCreate (through the API or via UI)
- Notice that kafkaCat should output a json blob. 
- (Optional) For schema verification, you can copy the json output from KafkaCat and paste it [here](https://internal.console.stage.redhat.com/api/notifications/internal/#/utils/console-cloud-event-validator), to very it is a valid schema.


pkg/handler/repositories.go
```diff
@@ -193,6 +195,14 @@ func (rh *RepositoryHandler) bulkCreateRepositories(c echo.Context) error {
 		return ce.NewErrorResponseFromError("Error creating repository", errs...)
 	}
 
+	// Send notifications
+	mappedValues := []repositories.Repositories{}
+	for i := 0; i < len(responses); i++ {
+		mappedValues = append(mappedValues, notifications.MapRepositoryResponse(responses[i]))
+	}
+
+	notifications.SendNotification(orgID, notifications.RepositoryCreated, mappedValues)
```
jlsherrill, line 204: should we add the  non-bulk create notification as part of this, or in a followup?

Andrewgdewar: We can, as a follow up.

pkg/notifications/client.go
```diff
@@ -0,0 +1,86 @@
...
+			log.Error().Msgf("failed to create cloudevents client, %v", err)
+		}
+
+		ctx := cloudevents.WithEncodingStructured(context.Background())
+
+		// Send the event
+		if result := c.Send(ctx, e); cloudevents.IsUndelivered(result) {
+			log.Printf("failed to send: %v", result)
+		} else {
+			log.Printf("accepted: %t", cloudevents.IsACK(result))
```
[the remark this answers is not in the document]
Andrewgdewar: .. yes it likely should.

pkg/notifications/client.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+
+		ctx := cloudevents.WithEncodingStructured(context.Background())
+
+		// Send the event
+		if result := c.Send(ctx, e); cloudevents.IsUndelivered(result) {
+			log.Error().Msgf("failed to send: %v", result)
+		} else {
+			log.Debug().Msgf("accepted: %t", cloudevents.IsACK(result))
+		}
```
jlsherrill, line 66: Can you adjust these debug messages a bit to be more descriptive:  "Notification message accepted: %t"  for example, and  "Notification message failed to send", just to give a little more context:

i was running locally and completely overlooked this:
```
1:51PM DBG accepted: true

```

jlsherrill (approved):

pkg/handler/repositories.go
rverdile, line 204: how would you feel about passing `[]api.RepositoryRepsonse` to `SendNotifaction` and calling the mapping function in there instead?

rverdile: eh nevermind

Andrewgdewar: hmm.. We could do that, but the problem I see is that depending on the action, we might not have that specific struct readily available causing us to make a further dao call.

The repositories.Repositories{} struct also only **requires** a few of the variables, meaning that even though other structs may not map perfectly, they should be able to satisfy the required items in the  Repositories{} struct without additional dao calls.

Here are the actions that we will be sending notifications on, and thier related/most accessible structs.
* repository create
    * RepositoryResponse
* repository introspected
    * RepositoryUpdate?
* repository updated
    *  RepositoryUpdate
* repository introspected failed
    *  Repository
* repository deleted
    * We only send the UUID, we should likely call to get the repo info to send in the notification response.

jlsherrill:
This now needs a rebase

PR #286, opened 2023-05-25
Fixes 1841: correct param validate mock

jlsherrill:
## Summary
The param validator mock was improperly defined, and thus improperly used.  When replaced with a proper mock, it caused a hang, this due to the wait group within the handler not properly marking Done() if an unexpected error occured.  By using defer, we can mark it Done() no matter what happens within the go routine.

## Testing steps
1.  checkout pr
2. make test
3. no hang!


jlsherrill:
https://issues.redhat.com/browse/HMS-1841

Andrewgdewar (approved):
ACK!

mshriver (approved):

PR #288, opened 2023-05-25
Fixes 602: Add Logging

Andrewgdewar:

### DIFF
--- a/pkg/notifications/client.go
+++ b/pkg/notifications/client.go
@@ -14,7 +14,7 @@ import (
 
 func SendNotification(orgId string, eventName EventName, repos []repositories.Repositories) {
 	if config.Get().NotificationsClient != nil {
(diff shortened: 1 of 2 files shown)


pkg/notifications/client.go
```diff
@@ -14,6 +14,7 @@ import (
 
 func SendNotification(orgId string, eventName EventName, repos []repositories.Repositories) {
 	if config.Get().NotificationsClient != nil {
+		log.Warn().Msgf("Notification started: %t", config.Get().NotificationsClient)
```
jlsherrill, line 17: %t is actually just true or false, you might use %v

Andrewgdewar: Is there a list of these somewhere, couldn't find them in the docs for zerolog

jlsherrill:
https://issues.redhat.com/browse/HMS-602

jlsherrill (approved):

PR #289, opened 2023-05-25
Fixes 602: Add SSL configuration

Andrewgdewar:

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -367,6 +367,7 @@ func SetupNotifications(kafkaServers []string, kafkaCaPath string) cloudevents.C
 	saramaConfig.Consumer.Offsets.Initial = sarama.OffsetOldest
 
 	if kafkaCaPath != "" {
+		log.Warn().Msgf("kafkaCaPath is: %s", kafkaCaPath)
 		tlsConfig, err := tlsutils.NewTLSConfig(kafkaCaPath)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -260,7 +261,7 @@ func Load() {
 
 	if len(clowder.KafkaServers) > 0 {
 		log.Warn().Msgf("clowder.KafkaServers has length of %s", strconv.Itoa(len(clowder.KafkaServers)))
-		LoadedConfig.NotificationsClient = SetupNotifications(clowder.KafkaServers)
+		LoadedConfig.NotificationsClient = SetupNotifications(clowder.KafkaServers, LoadedConfig.Kafka.Capath)
```
jlsherrill, line 264: i don't think this is quite right, we don't load the clowder() kafka config into our own.  i think you need to pull it from clowder: 

``` ca, err := clowder.LoadedConfig.KafkaCa(clowder.LoadedConfig.Kafka.Brokers...)```

you'll need to wrap it in a clowder.IsClowderEnabled()  too

jlsherrill: actually, i'm wrong, but lets log the Capath while we're at it?

rverdile:
should this be Refs or Fixes?

Andrewgdewar:
> 

Well it's attempting to fix something broken? 😞

jlsherrill:
yeah normally, you use Refs for initial commits before the pr that fixes it, or you may use Refs for a small followup pr, but this maybe is a little different, not entirely optimal situation (where we have to test in stage)

PR #291, opened 2023-05-25
Fixes 602: Update error to better return result

Andrewgdewar:


jlsherrill:
https://issues.redhat.com/browse/HMS-602

PR #298, opened 2023-05-30
Refs 1470: turn on new tasking system in stage

rverdile:
## Summary
This is to turn on the new tasking system in stage.

Still trying to figure out how to completely test this in ephemeral. Can see the env var set to "true", but that's it so far. Logging doesn't seem to work and I need to somehow create a user to trigger introspect.

## Testing steps
Should be able to see this change in ephemeral.


jlsherrill:
https://issues.redhat.com/browse/HMS-1470

jlsherrill:
any idea why this causes automation test failures?

jlsherrill (approved):
Tested in ephemeral, works great!

PR #299, opened 2023-05-30
Fixes 1470: switch nightly introspect to tasking system

rverdile:
## Summary
- Renames the `introspect-all` command to `nightly-jobs`
- If new tasking system has been enabled, adds each repository as introspect job to queue, instead of directly invoking introspect for all repositories

## Testing steps
1. Set `NEW_TASKING_SYSTEM` env var or `config.yaml` to true.
2. `go run cmd/external_repos/main.go nightly-jobs` will show `[Enqueued Task]...` for each repository.
3. Main server will need to be running for tasks to process.
4. Setting NEW_TASKING_SYSTEM to false should _not_ use the new tasking system, and instead directly invoke introspection for all repositories.

### DIFF
--- a/cmd/external-repos/main.go
+++ b/cmd/external-repos/main.go
@@ -1,7 +1,7 @@
 package main
 
 import (
-	"flag"
+	"fmt"
 	"os"
 	"sort"
 
@@ -9,6 +9,9 @@ import (
 	"github.com/content-services/content-sources-backend/pkg/dao"
 	"github.com/content-services/content-sources-backend/pkg/db"
 	"github.com/content-services/content-sources-backend/pkg/external_repos"
+	"github.com/content-services/content-sources-backend/pkg/tasks"
+	"github.com/content-services/content-sources-backend/pkg/tasks/client"
+	"github.com/content-services/content-sources-backend/pkg/tasks/queue"
 	_ "github.com/golang-migrate/migrate/v4/source/file"
 	_ "github.com/lib/pq"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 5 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-1470

cmd/external-repos/main.go
```diff
@@ -130,3 +131,35 @@ func scanForExternalRepos(path string) {
...
+		if err != nil {
+			log.Err(err).Msgf("error enqueueing introspecting for repository %v", repo.URL)
+		}
+	}
+	err = repoDao.OrphanCleanup()
+	if err != nil {
+		log.Err(err).Msg("error during orphan cleanup")
+	}
+	return nil
+}
```
jlsherrill, line 165: I kinda feel like this method should go in external_repos somewhere, what do you think?

pkg/handler/repositories.go
```diff
@@ -390,7 +390,7 @@ func (rh *RepositoryHandler) enqueueIntrospectEvent(c echo.Context, response api
 	var msg *message.IntrospectRequestMessage
 	var err error
 	if config.Get().NewTaskingSystem {
-		task := queue.Task{Typename: tasks.Introspect, Payload: tasks.IntrospectPayload{Url: response.URL}, OrgId: orgID, RepositoryUUID: response.RepositoryUUID}
+		task := queue.Task{Typename: tasks.Introspect, Payload: tasks.IntrospectPayload{Url: response.URL, Force: true}, OrgId: orgID, RepositoryUUID: response.RepositoryUUID}
```
jlsherrill, line 393: why force it here?

rverdile: This was already being forced. This is what resets the failed count, which is what we wanted to do with API since that's used for manually triggering introspection.

cmd/external-repos/main.go
rverdile: I thought I had a reason to put it here, but I now I can't remember. Maybe not. I'm gonna try moving it and see if I don't hit some kind of dependency error.

jlsherrill (approved):

PR #300, opened 2023-05-31
Fixes 1644: Enforce JSON Content Type

dpang314:
## Summary

Adds middleware `EnforceJSONContentType` that returns an error with status code 415 "Unsupported Media Type" if the Content-Type header is missing, invalid, or not `application/json`. If the Content-Type has optional parameters (e.g. `application/json; parameter=value`), it is still accepted.

The middleware is added at root level and applied to any request that has a body

It affects the following endpoints: 

1. /rpms/names [post] - searchRpmByName
2. /repository_parameters/external_gpg_key/ [post] - fetchGpgKey
3. /repository_parameters/validate/ [post] - validate
4. /repositories [post] - createRepository
5. /repositories/bulk_create/ [post] - bulkCreateRepositories
6. /repositories/{uuid} [put] - fullUpdate
7. /repositories/{uuid} [patch] - partialUpdate

/repositories/{uuid}/introspect/ [post] accepts an optional JSON body, so the middleware is only applied when the body is passed in.

## Testing steps

Make a curl request to any of the 7 endpoints without setting the Content-Type header or with a Content-Type header that isn't `application/json`

 `curl -H "$( ./scripts/header.sh 9999 1111 )" "http://localhost:8000/api/content-sources/v1.0/rpms/names" -X POST -d '{ "urls": [ "https://omaciel.fedorapeople.org/fakerepo01/", "https://dl.fedoraproject.org/pub/epel/8/Everything/x86_64/" ],   "search": "3" } '`

Curl seems to set the content type to `application/x-www-form-urlencoded` by default so if no content type is specified, the error title returned will be "Incorrect content type" 

To get "Error parsing content type" make a request with an empty Content-Type or a value that doesn't follow the MIME type format

 `curl -H "$( ./scripts/header.sh 9999 1111 )" "http://localhost:8000/api/content-sources/v1.0/rpms/names" -X POST -d '{ "urls": [ "https://omaciel.fedorapeople.org/fakerepo01/", "https://dl.fedoraproject.org/pub/epel/8/Everything/x86_64/" ],   "search": "3" } ' -H "Content-Type:"`

 `curl -H "$( ./scripts/header.sh 9999 1111 )" "http://localhost:8000/api/content-sources/v1.0/rpms/names" -X POST -d '{ "urls": [ "https://omaciel.fedorapeople.org/fakerepo01/", "https://dl.fedoraproject.org/pub/epel/8/Everything/x86_64/" ],   "search": "3" } ' -H "Content-Type: not a valid type"`

### DIFF
--- a/api/docs.go
+++ b/api/docs.go
@@ -238,6 +238,12 @@ const docTemplate = `{
                             "$ref": "#/definitions/errors.ErrorResponse"
                         }
                     },
+                    "415": {
+                        "description": "Unsupported Media Type",
+                        "schema": {
+                            "$ref": "#/definitions/errors.ErrorResponse"
+                        }
+                    },
                     "500": {
                         "description": "Internal Server Error",
                         "schema": {
@@ -305,6 +311,12 @@ const docTemplate = `{
                     },
                     "404": {
                         "description": "Not Found",
+                        "schema": {
+                            "$ref": "#/definitions/errors.ErrorResponse"
+                        }
+                    },
+                    "415": {
+                        "description": "Unsupported Media Type",
                         "schema": {
                             "$ref": "#/definitions/errors.ErrorResponse"
                         }
@@ -430,6 +442,12 @@ const docTemplate = `{
                             "$ref": "#/definitions/errors.ErrorResponse"
                         }
                     },
+                    "415": {
+                        "description": "Unsupported Media Type",
+                        "schema": {
+                            "$ref": "#/definitions/errors.ErrorResponse"
+                        }
+                    },
                     "500": {
                         "description": "Internal Server Error",
                         "schema": {
@@ -535,6 +553,12 @@ const docTemplate = `{
                     },
                     "404": {
                         "description": "Not Found",
+                        "schema": {
+                            "$ref": "#/definitions/errors.ErrorResponse"
+                        }
+                    },
+                    "415": {
+                        "description": "Unsupported Media Type",
                         "schema": {
                             "$ref": "#/definitions/errors.ErrorResponse"
                         }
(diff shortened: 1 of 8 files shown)


jlsherrill:
Would it be possible to implement a [middleware](https://echo.labstack.com/middleware/) to check for this?  I think you could assume that any PUT/POST/PATCH request should expect this content-type, and you could just ignore the check for GET/DELETE/anything else.

dpang314:
Middleware would definitely be a much cleaner solution. `POST /repositories/{uuid}/introspect/` currently has an optional body, but adding the middleware would require setting the content type to `application/json` even if you didn't want to specify a body. Is this fine?

jlsherrill:
https://issues.redhat.com/browse/HMS-1644

jlsherrill:
I think that's okay.  At least i'm okay with it.  Curious if @rverdile has any thoughts.  we could also ignore certain paths, which can be annoying but also would assure we dont' forget to add a check

rverdile:
Yeah if the middleware can blanket apply to all the PUT/POST/PATCH endpoints I think that makes sense. I think it also makes to explicitly ignore endpoints where the body is not required. I'm assuming that won't happen too often.

jlsherrill:
oktotest

jlsherrill:
[test]

jlsherrill (approved):

jlsherrill:
nice work @dpang314 !!   I've approved this and now it waits for QE to verify

swadeley:
Hi

`application/json; parameter=value` still works

if you omit the content type:
`{"errors":[{"status":415,"title":"Incorrect content type","detail":"Content-Type must be application/json"}]}`

If you pass `-H "Content-Type: Not a valid type"`:
`{"errors":[{"status":415,"title":"Error parsing content type","detail":"mime: expected slash after first token"}]}`

If you pass `-H "Content-Type:"`:
`{"errors":[{"status":415,"title":"Error parsing content type","detail":"mime: no media type"}]}`

PR #302, opened 2023-06-01
Refs 1871: Deployments: Introspect rhel-8.8 in ephemeral

jrusz:
We started building rhel-8.8 images and so we now need to also update the repository introspection in ephemeral environments.

Related to https://github.com/content-services/content-sources-backend/pull/301


mshriver (commented):
I don't think this is strictly necessary (there are a lot more repos to introspect) but we're using 8.8 in tests that end up failing because its not introspected yet.

mshriver (approved):
I don't think this is strictly necessary (there are a lot more repos to introspect) but we're using 8.8 in tests that end up failing because its not introspected yet.

PR #304, opened 2023-06-05
Fixes 1486: set rbac perms at route definition

jlsherrill:
this moves the permission declaration for routes to when the routes are defined so that we dont keep forgetting to define them.  This caused me to refactor a lot of the rbac files
* pkg/clients/ was renamed to pkg/rbac/
* rbac.go was renamed to client.go
* anything rbac related that wasn't middleware but was in /pkg/middlware was moved to pkg/rbac/
* a new type rbac.Resource was added, similar to rbac.Verb

In addition:
* parameter validate and external gpg key calls now require write permission, it didn't make sense for them to be read

## Testing steps
Enable rbac clients in  your configuration:
```
clients:
  rbac_enabled: true
  rbac_base_url: http://localhost:8800/api/rbac/v1
```
and configure mock service:
```
mocks:
  rbac:
    # Update with your account number for admin
    user_read_write: ["readwrite"]
    # Update with yout account number for viewer
    user_read: ["read"]
    user_no_permissions: ["noperms"]
```

then try to do various read/write tasks with the users above and very it works properly


pkg/handler/utils.go
```diff
@@ -24,3 +25,8 @@ func removeEndSuffix(source string, suffix string) string {
 
 	return output
 }
+
+func addRoute(e *echo.Group, method string, path string, h echo.HandlerFunc, verb rbac.Verb, m ...echo.MiddlewareFunc) {
+	e.Add(method, path, h, m...)
+	rbac.ServicePermissions.Add(method, path, rbac.ResourceRepositories, verb)
```
jlsherrill, line 31: note that this assumes a resource of 'Repositories'.  we can refactor if we ever have a 2nd resource type.

pkg/rbac/client_wrapper.go
```diff
@@ -33,43 +33,25 @@ import (
...
-	RbacVerbUnlink    RbacVerb = "unlink"
-	RbacVerbOrder     RbacVerb = "order"
-	RbacVerbExecute   RbacVerb = "execute"
-	RbacVerbUndefined RbacVerb = ""
-)
-
-type Rbac interface {
-	Allowed(ctx context.Context, resource string, verb RbacVerb) (bool, error)
+type ClientWrapper interface {
+	Allowed(ctx context.Context, resource Resource, verb Verb) (bool, error)
```
jlsherrill, line 37: Initially i wanted to name this client, but it already has an 'rbac.client' import, so ClientWrapper seemed more apt

rverdile (approved):
lgtm!

PR #306, opened 2023-06-06
Fixes 25: Remove selenium container from testing.yaml

tpapaioa:
## Summary
The test deployment in testing.yaml only runs api tests, so the `ui` section is unnecessary and creates an unused selenium container for each deployment. This PR removes the `ui` section so that the container isn't created at all.

## Testing steps
N/A


jlsherrill:
https://issues.redhat.com/browse/HMS-25

mshriver (approved):

PR #311, opened 2023-06-19
Fixes 1948: Add introspection/edit/delete notifications

Andrewgdewar:
## Testing steps

1) configure your config for kafka if not already:
```
kafka:
  auto:
    offset:
      reset: latest
    commit:
      interval:
        ms: 5000
  bootstrap:
    servers: localhost:9092
  group:
    id: content-sources
  message:
    send:
      max:
        retries: 15
  request:
    timeout:
      ms: 30000
    required:
      acks: -1
  retry:
    backoff:
      ms: 100
  timeout: 10000
  topics:
    - platform.content-sources.introspect
    - platform.notifications.ingress
```
2) run server `make run`
3) run listener ` KAFKA_TOPIC=platform.notifications.ingress make  kafka-topic-consume`
4) create a bulk repo, edit it, delete it, and see the associated notifications

### DIFF
--- a/pkg/dao/repository_configs.go
+++ b/pkg/dao/repository_configs.go
@@ -8,11 +8,14 @@ import (
 	"time"
 
 	"github.com/ProtonMail/go-crypto/openpgp"
+	"github.com/RedHatInsights/event-schemas-go/apps/repositories/v1"
 	"github.com/content-services/content-sources-backend/pkg/api"
 	ce "github.com/content-services/content-sources-backend/pkg/errors"
 	"github.com/content-services/content-sources-backend/pkg/models"
+	"github.com/content-services/content-sources-backend/pkg/notifications"
 	"github.com/content-services/yummy/pkg/yum"
 	"github.com/jackc/pgx/v5/pgconn"
+	"github.com/rs/zerolog/log"
 	"gorm.io/gorm"
 	"gorm.io/gorm/clause"
 )
@@ -80,8 +83,16 @@ func (r repositoryConfigDaoImpl) Create(newRepoReq api.RepositoryRequest) (api.R
 
 	var created api.RepositoryResponse
 	ModelToApiFields(newRepoConfig, &created)
+
 	created.URL = newRepo.URL
 	created.Status = newRepo.Status
+
+	//Send update notification
+	notifications.SendNotification(
+		newRepoConfig.OrgID,
+		notifications.RepositoryCreated,
+		[]repositories.Repositories{notifications.MapRepositoryResponse(created)},
+	)
 
 	return created, nil
 }
@@ -98,6 +109,13 @@ func (r repositoryConfigDaoImpl) BulkCreate(newRepositories []api.RepositoryRequ
 		}
 		return err
 	})
+
+	// Send notifications
+	mappedValues := []repositories.Repositories{}
+	for i := 0; i < len(responses); i++ {
+		mappedValues = append(mappedValues, notifications.MapRepositoryResponse(responses[i]))
+	}
+	notifications.SendNotification(*newRepositories[0].OrgID, notifications.RepositoryCreated, mappedValues)
 
 	return responses, errs
 }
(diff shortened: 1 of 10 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-1948

pkg/notifications/client.go
```diff
@@ -71,6 +69,28 @@ func MapRepositoryResponse(importedRepo api.RepositoryResponse) repositories.Rep
...
+		DistributionVersions:         importedRepo.Versions,
+		FailedIntrospectionsCount:    &failedIntrospectionsCount,         // Optional but defaults to 0
+		PackageCount:                 &packageCount,                      // Optional but defaults to 0
+		MetadataVerification:         &importedRepo.MetadataVerification, // Optional but defaults to false
+		DistributionArch:             SetEmptyToNil(importedRepo.Arch),   // Below are all optional, we need to set them as nil if empty for the cloud schema
+		GPGKey:                       SetEmptyToNil(importedRepo.GpgKey),
+		LastIntrospectionError:       SetEmptyToNil(*importedRepo.Repository.LastIntrospectionError),
+		LastIntrospectionTime:        SetEmptyToNil(importedRepo.Repository.LastIntrospectionTime.String()),
+		LastSuccessIntrospectionTime: SetEmptyToNil(importedRepo.Repository.LastIntrospectionSuccessTime.String()),
+		LastUpdateIntrospectionTime:  SetEmptyToNil(importedRepo.Repository.LastIntrospectionUpdateTime.String()),
```
jlsherrill, line 89: These 3 lines are assuming that the Time is not nil, but in our test cases they are actually nil, and nil.String() blows up with a panic, this is your test failure

Andrewgdewar: OH! nice catch. Thank you.

pkg/external_repos/introspect.go
```diff
@@ -164,9 +174,62 @@ func IntrospectAll(urls *[]string, force bool) (int64, []error, []error) {
...
+				wg.Add(1)
+				count = count + 1
+				go func(index int) {
+					notifications.SendNotification(
+						repos[index].OrgID,
+						notifications.RepositoryIntrospected,
+						[]repositories.Repositories{notifications.MapRepositoryResponse(repos[index])},
+					)
+					count = count - 1
+					wg.Done()
```
rverdile, line 205: Goroutines can't share variables from outside their scope like you're doing with `count = count - 1`. Since each parallel goroutine is simultaneously updating that variable, the result is undefined.

You might do instead is reset the counter under the `if count > 100` block. And with that, I think you would also want to call `wg.Wait()` after the for loop, to wait for whatever's left.

Andrewgdewar: Thanks for that, I also caught an error in the notificationFailure loop.
Please take a look at the logic in the follow-up commit and let me know what you think @rverdile. 

Cheers

api/swagger.json
rverdile, line 1: what are these swagger.json and swagger.yaml files for? 

Our `make openapi` command makes these and then removes them. Are these here on purpose?

pkg/dao/repository_configs.go
```diff
@@ -332,6 +361,15 @@ func (r repositoryConfigDaoImpl) Delete(orgID string, uuid string) error {
...
+		ModelToApiFields(repoConfig, &repositoryResponse)
+		//Send delete notification
+		notifications.SendNotification(
+			orgID,
+			notifications.RepositoryDeleted,
+			[]repositories.Repositories{notifications.MapRepositoryResponse(repositoryResponse)},
+		)
 	}
 	return nil
 }
```
jlsherrill, line 375: The previous notification pr sent this in the handler on bulkcreate, i think we should probably send them either in the handler OR in the dao, but not both.  I.e. it should either be at the outer layers or the inner.  Thoughts?

jlsherrill:
I think we need a notification for non-bulk create?

api/swagger.json
Andrewgdewar: I don't believe so, I did "make openapi" and they stuck around. It does create a misleading change log.

rverdile: Them sticking around seems like a bug. They don't stick around for me.

pkg/dao/repository_configs.go
Andrewgdewar: Hmm I suppose that makes sense. 👍

Andrewgdewar: Thinking on this, it wouldn't make sense to have everything on the handler layer. 
Many dao methods are called in different places. So I think the dao is the logical place to put this, if we want to follow that thought process.

api/swagger.json
Andrewgdewar: I removed them!

pkg/dao/repository_configs.go
```diff
@@ -80,9 +83,17 @@ func (r repositoryConfigDaoImpl) Create(newRepoReq api.RepositoryRequest) (api.R
 
 	var created api.RepositoryResponse
 	ModelToApiFields(newRepoConfig, &created)
+
 	created.URL = newRepo.URL
 	created.Status = newRepo.Status
 
+	//Send update notification
```
rverdile, line 90:
```suggestion
	//Send create notification
```
also okay with removing the comment because the code is named well enough to explain itself

rverdile (commented):
Not sure if @jlsherrill wants to take another look, but I tested all the endpoints notifications were added to and it's working.

jlsherrill:
needs a rebase, but ack from a code perspective.  I trust @rverdile's testing

PR #323, opened 2023-06-29
Refs 1947: Update IQE CJI template name and parameter defaults

tpapaioa:
## Summary
Now that testing.yaml is used for both stage and prod post-deploy tests, I've made a couple updates:
- Remove references to "stage" in the template and CJI names
- Remove the default IQE test marker and filter values, and marked the parameters as required, so that the values are passed in from the saas files in app-interface.


jlsherrill:
https://issues.redhat.com/browse/HMS-1947

jlsherrill (approved):

PR #324, opened 2023-06-29
Fixes 2058: Empty url on introspect notification

Andrewgdewar:
## Summary
Fixed a bug that would return an empty URL for the introspect notification, causing the schema to blow up. 
This was a simple DB call order mistake.


jlsherrill:
https://issues.redhat.com/browse/HMS-2058

rverdile (approved):

PR #325, opened 2023-06-30
Fixes 326: Add repository bulk delete API

dpang314:
## Summary

Adds an endpoint at `POST /repositories/bulk_delete/` to bulk delete repository configs by passing a list of UUIDs in the request body:

```
{
    uuids: ["uuid-1", "uuid-2"]
}
```

If an error occurs in deleting any of the repositories, the entire call will be canceled and return a list of errors. The max number of repositories you can delete in a single call is 100

## Testing steps

The org_id must match the org_id of the repository_configurations you are trying to delete.

```
curl "http://localhost:8000/api/content-sources/v1/repositories/bulk_delete/"  -H "`./scripts/header.sh {org_id} user`" -H "Content-Type: application/json" -d '{"uuids": [ "uuid-1" ]} '
```


jlsherrill:
https://issues.redhat.com/browse/HMS-326

pkg/handler/repositories.go
```diff
@@ -347,6 +349,80 @@ func (rh *RepositoryHandler) deleteRepository(c echo.Context) error {
...
 
+// BulkDeleteRepositories godoc
+// @Summary      Bulk delete repositories
+// @ID           bulkDeleteRepositories
+// @Description  bulk delete repositories
+// @Tags         repositories
+// @Accept       json
+// @Produce      json
+// @Param        body  body     api.UUIDListRequest  true  "Identifiers of the repositories"
+// @Success			 204 "Repository was successfully deleted"
```
rverdile, line 360:
```suggestion
// @Success			 204 "Repositories were successfully deleted"
```

pkg/dao/repository_configs.go
```diff
@@ -387,6 +387,66 @@ func (r repositoryConfigDaoImpl) Delete(orgID string, uuid string) error {
...
+		}
+
+		if err = tx.Delete(&repoConfig).Error; err != nil {
+			dbErr = DBErrorToApi(err)
+			errors[i] = dbErr
+			tx.RollbackTo(save)
+			continue
+		}
+
+		if dbErr != nil {
```
rverdile, line 438:
```suggestion
		if dbErr == nil {
```
This is saying "only create responses if there is any error during this process". But what we want to do is the opposite, "only create responses if there is no error during this process". Does that make sense?

pkg/dao/repository_configs.go
```diff
@@ -387,6 +387,66 @@ func (r repositoryConfigDaoImpl) Delete(orgID string, uuid string) error {
...
+			err = errors.New("rollback bulk delete")
+		}
+		return err
+	})
+
+	mappedValues := make([]repositories.Repositories, len(uuids))
+	for i := 0; i < len(responses); i++ {
+		mappedValues = append(mappedValues, notifications.MapRepositoryResponse(responses[i]))
+	}
+	notifications.SendNotification(orgID, notifications.RepositoryDeleted, mappedValues)
```
rverdile, line 407:
```suggestion
	mappedValues := make([]repositories.Repositories, len(uuids))
	for i := 0; i < len(responses); i++ {
		mappedValues[i] = notifications.MapRepositoryResponse(responses[i])
	}
	if len(mappedValues) > 0 {
	        notifications.SendNotification(orgID, notifications.RepositoryDeleted, mappedValues)
	}
```
A couple of things here. And you don't have to change it exactly this way, but I think showing the difference is the best way to explain it.

1. I changed the list append syntax to the indexing because the `make` function allocates a list of empty values. By appending, we'd be appending to that empty list, instead of replacing, causing blank notifications to go out.
2. We want to check that mappedValues has values to send out, otherwise we'd be sending out an empty notification even when there is an error.

rverdile:
everything else is working great!!

rverdile (commented):
Notifications are working now!! I found one more thing and then this should be good to go.

Right now if I make a request with no body, or with an empty list of UUIDs, the response returns 204 but nothing actually happens. I think we should catch that somewhere and ultimately return a 400.

rverdile (approved):
nice job!

swadeley:
Hi

I created two repos. I could delete them like so:
`app.content_sources.rest_client.repositories_api.bulk_delete_repositories({"uuids":['35d5f4ea-3a50-47bf-91d4-b92b27c5cfb6','5b9be49c-3cf7-4566-8a50-364609deda81']})`

I created two new repos. I tried to delete one plus the previously deleted  two repos:
`HTTP response body: {"errors":[{"status":404,"title":"Error deleting repositories","detail":"Could not find repository with UUID 35d5f4ea-3a50-47bf-91d4-b92b27c5cfb6"},{"status":404,"title":"Error deleting repositories","detail":"Could not find repository with UUID 5b9be49c-3cf7-4566-8a50-364609deda81"},{}]}`
and the repos are still there, as expected.

I confirm empty list returns 400:
```
In [16]: app.content_sources.rest_client.repositories_api.bulk_delete_repositories({"uuids":[]})
HTTPError: 400 Client Error: Bad Request for url: https://<snip>
```

Thank you

PR #326, opened 2023-06-30
Fixes 2065: conditional Introspect & fixed update 

Andrewgdewar:
## Summary

This fixes the empty struct values for the update notification that were causing it to fail to be parsed by the schems. 

This adds conditional sending of the introspect notification so that it now will only send a notification when the introspection would result in an update.

## Testing steps

1) configure your config for kafka if not already:
```
kafka:
  auto:
    offset:
      reset: latest
    commit:
      interval:
        ms: 5000
  bootstrap:
    servers: localhost:9092
  group:
    id: content-sources
  message:
    send:
      max:
        retries: 15
  request:
    timeout:
      ms: 30000
    required:
      acks: -1
  retry:
    backoff:
      ms: 100
  timeout: 10000
  topics:
    - platform.content-sources.introspect
    - platform.notifications.ingress
```
2) run server `make run`
3) run listener ` KAFKA_TOPIC=platform.notifications.ingress make  kafka-topic-consume`
4) create a bulk repo, edit it, delete it, and see the associated notifications. 
5) check the json for the "package_count": and "url": values (these were appearing as empty string "", when editing/updating a repo).

### DIFF
--- a/pkg/dao/repository_configs.go
+++ b/pkg/dao/repository_configs.go
@@ -331,11 +331,6 @@ func (r repositoryConfigDaoImpl) Update(orgID string, uuid string, repoParams ap
 		repoConfig.RepositoryUUID = repo.UUID
 	}
 
-	repoConfig.Repository = models.Repository{}
-	if err := r.db.Model(&repoConfig).Updates(repoConfig.MapForUpdate()).Error; err != nil {
-		return DBErrorToApi(err)
-	}
-
 	repositoryResponse := api.RepositoryResponse{}
 	ModelToApiFields(repoConfig, &repositoryResponse)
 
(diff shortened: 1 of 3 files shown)


rverdile (commented):
I noticed an issue with BulkCreate notifications while reviewing the BulkDelete PR.

Right now in BulkCreate, if there is an error and no repositories are created, a notification will be sent out because mappedValues here could be empty: https://github.com/content-services/content-sources-backend/blob/main/pkg/dao/repository_configs.go#L112-L116

Do you want to fix that here? You'd just check if it's empty or not before sending a notification.

swadeley (approved):
lgtm

PR #328, opened 2023-07-03
Debug pulp authentication

jlsherrill:
## Summary

Currently debugging authentication errors in ephemeral environments 

Current thought is that the backend app is up but not properly configured for pulp, although there is some evidence this isn't the case.  This change means that our app will not be 'up' until pulp is fully up and running.  

TODO:  make this work without a pulp server!
DO NOT MERGE


swadeley:
Hi

Where would you expect to see this message:
message = "Pulp DB Connection not available."

PR #333, opened 2023-07-13
Fixes 1846: log request ID

rverdile:
## Summary
- Adds support for the request ID, `x-rh-insights-request-id`, sent in the header and propagates that request ID to logging in the API, tasking, and redis
- Also propagates TaskID/Type logging throughout tasking system and pulp. 

The goal here is to more easily track which events correspond to which request. For example, if I get an error during repository snapshot, it is nice to know which request or task triggered that failed snapshot.

TODO
- [x] I need to double check redis logging. Will add testing step for that as well.

## Testing steps
1. Make an API request, for example, one to create a repository. Add the `x-rh-insights-request-id` header.
```
POST http://localhost:8000/api/content-sources/v1.0/repositories/
x-rh-identity: zEzcvAe....
x-Rh-Insights-Request-Id: 12345
```
3. You should see the request ID propagated through the to the logs for the API, tasking, and introspection. Set your log level to  `debug` to make sure you see all the logs.
4. Try variations of snapshotting, introspecting, CTRL+C exiting the server while it's in the middle of a task, and make sure the logs are all showing the request ID (or task ID if applicable).
5. Provide feedback if you think something should or shouldn't be logged.
6. A quick test for RBAC/Redis logging is to hard-code an error in `client_wrapper.go` and see the error get logged.

### DIFF
--- a/pkg/config/logging.go
+++ b/pkg/config/logging.go
@@ -17,6 +17,9 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+const HeaderRequestId = "x-rh-insights-request-id" // the header that contains the request ID
+const RequestIdLoggingKey = "request_id"           // the key that represents the request ID when logged
+
 func ConfigureLogging() {
 	var writers []io.Writer
 
--- a/pkg/dao/interfaces.go
+++ b/pkg/dao/interfaces.go
@@ -2,6 +2,7 @@ package dao
 
 import (
 	"context"
+
 	"github.com/content-services/content-sources-backend/pkg/api"
 	"github.com/content-services/content-sources-backend/pkg/models"
 	"github.com/content-services/content-sources-backend/pkg/pulp_client"
--- a/pkg/dao/interfaces.go
+++ b/pkg/dao/interfaces.go
@@ -1,6 +1,7 @@
 package dao
 
 import (
+	"context"
 	"github.com/content-services/content-sources-backend/pkg/api"
 	"github.com/content-services/content-sources-backend/pkg/models"
 	"github.com/content-services/content-sources-backend/pkg/pulp_client"
@@ -29,7 +30,7 @@ func GetDaoRegistry(db *gorm.DB) *DaoRegistry {
 		Metrics:    metricsDaoImpl{db: db},
 		Snapshot:   snapshotDaoImpl{db: db},
 		TaskInfo:   taskInfoDaoImpl{db: db},
-		AdminTask:  adminTaskInfoDaoImpl{db: db, pulpClient: pulp_client.GetPulpClient()},
+		AdminTask:  adminTaskInfoDaoImpl{db: db, pulpClient: pulp_client.GetPulpClient(context.Background())},
 	}
 	return &reg
 }
--- a/pkg/middleware/enforce_identity.go
+++ b/pkg/middleware/enforce_identity.go
@@ -23,6 +23,10 @@ func WrapMiddlewareWithSkipper(m func(http.Handler) http.Handler, skip echo_midd
 				if identityHeader != "" {
 					c.Response().Header().Set("X-Rh-Identity", identityHeader)
 				}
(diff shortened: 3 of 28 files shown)


pkg/middleware/enforce_identity.go
```diff
@@ -23,6 +23,10 @@ func WrapMiddlewareWithSkipper(m func(http.Handler) http.Handler, skip echo_midd
 				if identityHeader != "" {
 					c.Response().Header().Set("X-Rh-Identity", identityHeader)
 				}
+				requestIdHeader := c.Request().Header.Get(config.HeaderRequestId)
```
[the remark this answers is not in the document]
rverdile: good catch! Out of curiosity, I also tried commenting out `identityHeader` here and I didn't see anything break either. I'm not sure what this was for. I'm gonna remove it for `requestId` for now.

pkg/config/logging.go
```diff
@@ -17,6 +17,9 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+const HeaderRequestId = "x-rh-insights-request-id" // the header that contains the request ID
```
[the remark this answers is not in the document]
rverdile: That was my intention, but it looks like I missed one! I'll update that.

pkg/dao/interfaces.go
```diff
@@ -29,7 +30,7 @@ func GetDaoRegistry(db *gorm.DB) *DaoRegistry {
 		Metrics:    metricsDaoImpl{db: db},
 		Snapshot:   snapshotDaoImpl{db: db},
 		TaskInfo:   taskInfoDaoImpl{db: db},
-		AdminTask:  adminTaskInfoDaoImpl{db: db, pulpClient: pulp_client.GetPulpClient()},
+		AdminTask:  adminTaskInfoDaoImpl{db: db, pulpClient: pulp_client.GetPulpClient(context.Background())},
```
rverdile, line 33: Still thinking about the right way to update the pulp client in the request such that it's context has the request ID.

rverdile: But at the moment this would not log anything anyway

rverdile:
This is ready for another review. 

Propagating context through admin_tasks handler to pulp_client is a bigger change than I anticipated because of how it affects tests. Since is has no impact on the logging (since nothing is logged there at the moment), it might make sense to leave it out from here.

pkg/middleware/rbac.go
```diff
@@ -35,6 +35,7 @@ func NewRbac(config Rbac) echo.MiddlewareFunc {
 	}
 	return func(next echo.HandlerFunc) echo.HandlerFunc {
 		return func(c echo.Context) error {
+			logger := zerolog.Ctx(c.Request().Context())
```
dpang314, line 38: Logging works with the request id for me. However, when making a request without the `x-Rh-Insights-Request-Id` header, RBAC errors are logged with the random request id generated by `echo_middleware.RequestIDWithConfig`. This is in contrast to logging in other locations, such as when a task is enqueued/dequeued/finished. In those log messages, the request id is empty when not provided in the request. What is the intended behavior for requests missing the `x-Rh-Insights-Request-Id` header?

rverdile: I missed this, thanks. I've updated it so now the request ID, in any place, should use the generated one if a request ID is not specified.

dpang314 (approved):
Looks good to me!

pkg/dao/interfaces.go
jlsherrill: This might be a similar 'problem', but right now if i turn up db logging (set logging_level to trace), i don't see a request_id on those logs.  This would similarly change as part of each request

jlsherrill: And this could be helpful, if a certain query takes too long it is logged automatically.  Having the request_id associated with that would be helpful to trace it back to the original request.

PR #337, opened 2023-07-20
Fixes 2183: fix deadlock in concurrent task enqueue

jlsherrill:
## Summary

This resolves a couple of issues:
* Enqueue() was not properly releasing its acquired db pool connection, this resulted in deadlocks when more than ~20 create reqs came in at once
* Sets a limit on the postgres connections for gorm, as if you perform ~50 create reqs at the same time a postgres server with a default limit of connections will start to max out and start issuing errors
* Reworks pgqueue to use a wrapper around pgxpool.Pool, this allows for testing while giving pgqueue the ability to start/end transactions and acquire/release Pool connections

## Testing steps

Checkout this pr and 'make run'
create a new file `./cmd/test/main.go`   with the contents here: https://gist.github.com/jlsherrill/d07efa70bed152d8428c758b7332f95e

then run it:  `go run ./cmd/test/main.go`

You should only get back 201s, and all tasks should be processed without error (other than the 404)

### DIFF
--- a/pkg/tasks/queue/pgqueue.go
+++ b/pkg/tasks/queue/pgqueue.go
@@ -465,11 +465,10 @@ func (p *PgQueue) Status(taskId uuid.UUID) (*models.TaskInfo, error) {
 	// Use double pointers for timestamps because they might be NULL, which would result in *time.Time == nil
 	var info models.TaskInfo
 	conn, err := p.Pool.Acquire(context.Background())
-
-	defer conn.Release()
 	if err != nil {
 		return nil, err
 	}
+	defer conn.Release()
 	err = conn.QueryRow(context.Background(), sqlQueryTaskStatus, taskId).Scan(
 		&info.Id, &info.Typename, &info.Payload, &info.Queued, &info.Started, &info.Finished, &info.Status,
 		&info.Error, &info.OrgId, &info.RepositoryUUID, &info.Token,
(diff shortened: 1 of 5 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-2183

pkg/tasks/queue/pgqueue.go
```diff
@@ -434,12 +460,16 @@ func (p *PgQueue) taskDependents(ctx context.Context, conn Connection, id uuid.U
...
-	conn = p.Conn
 
 	// Use double pointers for timestamps because they might be NULL, which would result in *time.Time == nil
 	var info models.TaskInfo
+	conn, err := p.Pool.Acquire(context.Background())
+
+	defer conn.Release()
+	if err != nil {
+		return nil, err
+	}
```
rverdile, line 472: I think it would be better to check for an error before deferring release. I'm not sure what happens if there's an error and you release the connection.

pkg/tasks/queue/pgqueue.go
```diff
@@ -667,10 +691,11 @@ func (p *PgQueue) RefreshHeartbeat(token uuid.UUID) {
 }
 
 func (p *PgQueue) IdFromToken(token uuid.UUID) (id uuid.UUID, isRunning bool, err error) {
-	conn := p.Conn
-
 	var status string
-	err = conn.QueryRow(context.Background(), sqlQueryIdFromToken, token).Scan(&id, &status)
+	conn, err := p.Pool.Acquire(context.Background())
+	defer conn.Release()
```
rverdile, line 696: check for error here

rverdile:
everything else looks good to me

pkg/tasks/queue/pgqueue.go
jlsherrill: agreed! good point

rverdile (approved):
nice!

swadeley:
Tested API OK.

PR #338, opened 2023-07-21
Fixes 2083: log when a task fails to complete

rverdile:
## Summary
PR #333 should merge first. This PR is built on that PR.

To make it clearer that the task failed, this changes logging to log the error of the failed task.

This PR in combination with this [commit to zest](https://github.com/content-services/zest/commit/ee55f31f4234ca312ca13fc860b8cd0ca923b2cb) should fully address the JIRA issue. The change to zest will not be used until we update our app to a newer version of zest. 

## Testing steps
1. Create a repo with bad url (i.e. `http://example.com`), with snapshot enabled. 
2. You should see the error `json: cannot unmarshal string into Go struct field TaskResponse.error of type map[string]interface {}"` for the sync task specifically. 
3. To test the change to zest, update your go.mod to use the latest zest version. There will be some syntax errors in the pulp_client package that you will have to manually fix to use the latest version. The functions that end in `Api` have to be changed to `API`. 
4. Now you should see a different error for the sync task.

### DIFF
--- a/pkg/tasks/worker/worker.go
+++ b/pkg/tasks/worker/worker.go
@@ -155,21 +155,27 @@ func (w *worker) process(ctx context.Context, taskInfo *models.TaskInfo) {
 
 	if handler, ok := w.handlers[taskInfo.Typename]; ok {
 		var finishStr string
-		err := handler(ctx, taskInfo, &w.queue)
-		if err != nil {
-			finishStr = fmt.Sprintf("task failed with error: %v", err)
-			logger.Warn().Msgf("error during task processing: %v", err)
+
+		handlerErr := handler(ctx, taskInfo, &w.queue)
+		if handlerErr != nil {
+			finishStr = fmt.Sprintf("task failed with error: %v", handlerErr)
 			w.metrics.RecordMessageResult(false)
 		} else {
 			finishStr = "task completed"
 			w.metrics.RecordMessageResult(true)
 		}
 
-		err = w.queue.Finish(taskInfo.Id, err)
+		err := w.queue.Finish(taskInfo.Id, handlerErr)
 		if err != nil {
 			logger.Error().Msgf("error finishing task: %v", err)
 		}
-		logger.Info().Msgf("[Finished Task] %v", finishStr)
+
+		if handlerErr != nil {
+			logger.Warn().Msgf("[Finished Task] %v", finishStr)
+		} else {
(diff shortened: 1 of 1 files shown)


rverdile:
Do not merge before PR #333

jlsherrill:
https://issues.redhat.com/browse/HMS-2083

rverdile:
needs rebase

pkg/tasks/worker/worker.go
```diff
@@ -153,18 +154,22 @@ func (w *worker) process(ctx context.Context, taskInfo *models.TaskInfo) {
 	defer recoverOnPanic(*logger)
 
 	if handler, ok := w.handlers[taskInfo.Typename]; ok {
+		var finishStr string
 		err := handler(ctx, taskInfo, &w.queue)
 		if err != nil {
+			finishStr = fmt.Sprintf("task failed with error: %v", err)
+			logger.Warn().Msgf("error during task processing: %v", err)
```
jlsherrill, line 161: this actually results in the error getting logged twice for me:
```
4:47PM WRN error during task processing: traceback:   File "/usr/local/lib/python3.8/site-packages/pulpcore/tasking/tasks.py", line 65, in _execute_task

```

```
4:47PM INF [Finished Task] task failed with error: traceback:   File "/usr/local/lib/python3.8/site-packages/pulpcore/tasking/tasks.py", line 65, in _execute_task
```

jlsherrill: one is info and one is warn.  Could we just do it once (severity could change based on if its failed or not)

rverdile: fixed

jlsherrill (approved):

jlsherrill:
I still see 
```
json: cannot unmarshal string into Go str
uct field TaskResponse.error of type map[string]interface {}
```
in the task error field.  That should be fixed now right?

rverdile:
> I still see
> 
> ```
> json: cannot unmarshal string into Go struct field TaskResponse.error of type map[string]interface {}
> ```
> 
> in the task error field. That should be fixed now right?

It's still fixed for me. Are you on a recent zest version?

jlsherrill:
ah, i must have had an older version.  Looking good.  I am seeing the full traceback in the error, but we probably don't want to expose that to the user.  I'll file a bug!

PR #339, opened 2023-07-27
Fixes 2224: admin task fetch 404

dpang314:
## Summary

Improves the "not found" error message when fetching an admin task using a uuid with invalid syntax.

Previously, when making a GET request to `/admin/tasks/:uuid` with a uuid that had invalid syntax, e.g. "invalid-uuid", the error message would be `(ERROR: invalid input syntax for type uuid: "invalid-uuid" (SQLSTATE 22P02))`

The PR changes the error message to `Could not find task with UUID invalid-uuid` which is consistent with other fetch APIs

## Testing steps

```curl localhost:8000/api/content-sources/v1.0/admin/tasks/bad-uuid  -H "`./scripts/header.sh 1 adminUser`"```
The "admin_tasks" config feature must be enabled. Either both "admin_tasks.accounts" and "admin_tasks.users" must  be nil to allow any user to access it, or "adminUser" must be added to "admin_tasks.users" to access the endpoint.

This can also be tested through the frontend in [this PR](https://github.com/content-services/content-sources-frontend/pull/128) by navigating to `/beta/insights/content/admin-tasks/bad-uuid` You should be redirected to `/beta/insights/content/admin-tasks` and the error will pop up as a notification.

### DIFF
--- a/pkg/dao/admin_tasks.go
+++ b/pkg/dao/admin_tasks.go
@@ -30,7 +30,7 @@ func GetAdminTaskDao(db *gorm.DB, pulpClient pulp_client.PulpClient) AdminTaskDa
 
 func (a adminTaskInfoDaoImpl) Fetch(id string) (api.AdminTaskInfoResponse, error) {
 	taskInfo := models.TaskInfo{}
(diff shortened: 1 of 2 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-2224

jlsherrill (approved):
Nice work!

PR #350, opened 2023-08-03
Fixes 2270: prevent out of order migrations

jlsherrill:
## Summary

1.  Creates a new file ./db/migrations.latest which holds the timestamp of the latest migration
2. on db-migrate-up, the command errors out if the value in   ./db/migrations.latest doesn't match the latest in migration in ./db/migrations
3.  `go run ./cmd/dbmigrate/go.mod  new my_migration'  will auto-populate migrations.latest with the newest timestamp
4. The result is that prs should create conflicts on merge if they both contain migrations, ensuring that we never commit any that are out of order.

## Testing steps

Try generating a new migration without updating migrations.latest.  Try to migrate up (you should get an error).


jlsherrill:
https://issues.redhat.com/browse/HMS-2270

pkg/db/db.go
```diff
@@ -174,3 +176,41 @@ func getPreviousMigrationVersion(m *migrate.Migrate) (int, error) {
...
+	last := migrationFileNames[len(migrationFileNames)-1]
+	nameArr := strings.Split(last, "_")
+	expectedLatest, err := os.ReadFile(LatestMigrationFile)
+	if err != nil {
+		return err
+	}
+	datetime := nameArr[0]
+	trimmed := strings.TrimSpace(string(expectedLatest))
+	if datetime != trimmed {
+		return fmt.Errorf("Latests migration from %v (%v) does not match found latest file (%v)", LatestMigrationFile, trimmed, datetime)
```
rverdile, line 196:
```suggestion
		return fmt.Errorf("Latest migration from %v (%v) does not match found latest file (%v)", LatestMigrationFile, trimmed, datetime)
```

rverdile (approved):
found 1 typo but otherwise lgtm!

PR #353, opened 2023-08-09
Fixes 2310: fix 404 err on delete task

jlsherrill:
## Summary
if no domain was ever created

## Testing steps

In an org that has never created any repos (or at least none that were snapshotted).  Create a repo:

```
POST http://localhost:8000/api/content-sources/v1.0/repositories/
x-rh-identity: eyJpZGVudGl0eSI6eyJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJmb28ifSwiYWNjb3VudF9udW1iZXIiOiIzMCIsImludGVybmFsIjp7Im9yZ19pZCI6IjMwIn19fQo=
x-Rh-Insights-Request-Id: 9876


{"name":"pulp 3.16","url":"http://yum.theforeman.org/pulpcore/3.16/el8/x86_64/","snapshot": false}
```

get the UUID of the created repo and delete it:

```

DELETE http://localhost:8000/api/content-sources/v1.0/repositories/14c3a83d-7574-4654-9a2d-8d4c92254b39
x-rh-identity: eyJpZGVudGl0eSI6eyJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJmb28ifSwiYWNjb3VudF9udW1iZXIiOiIzMCIsImludGVybmFsIjp7Im9yZ19pZCI6IjMwIn19fQo=

```

Now check the tasks:

```
GET http://localhost:8000/api/content-sources/v1.0/tasks/
x-rh-identity: eyJpZGVudGl0eSI6eyJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJmb28ifSwiYWNjb3VudF9udW1iZXIiOiIzMCIsImludGVybmFsIjp7Im9yZ19pZCI6IjMwIn19fQo=
```

the delete task should not be in error state

### DIFF
--- a/pkg/tasks/delete_repository_snapshots_test.go
+++ b/pkg/tasks/delete_repository_snapshots_test.go
@@ -1,6 +1,7 @@
 package tasks
 
 import (
+	"context"
 	"fmt"
 	"testing"
 
@@ -28,6 +29,10 @@ func TestDeleteSnapshotSuite(t *testing.T) {
 	suite.Run(t, new(DeleteRepositorySnapshotsSuite))
 }
 
+func (s *DeleteRepositorySnapshotsSuite) pulpClient() pulp_client.PulpClient {
+	return &s.MockPulpClient
+}
+
 func (s *DeleteRepositorySnapshotsSuite) SetupTest() {
 	s.mockDaoRegistry = dao.GetMockDaoRegistry(s.T())
 	s.MockPulpClient = *pulp_client.NewMockPulpClient(s.T())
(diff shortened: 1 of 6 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-2310

test/integration/delete_test.go
```diff
@@ -0,0 +1,114 @@
...
+	wrk.HeartbeatListener()
+
+	wkrCtx := context.Background()
+	go (wrk).StartWorkers(wkrCtx)
+	go func() {
+		<-wkrCtx.Done()
+		wrk.Stop()
+	}()
+	// Force local storage for integration tests
+	config.Get().Clients.Pulp.StorageType = "local"
```
rverdile, line 51: I'm pretty sure that this will get overwritten by whatever is in `config.yaml` if `config.Get()` gets called again in one of the tests.

Assuming this can also be set by env variable, you could do something like `t.SetEnv(key, val)` instead.

test/integration/delete_test.go
```diff
@@ -0,0 +1,114 @@
...
+	taskUuid, err := taskClient.Enqueue(queue.Task{
+		Typename:       config.DeleteRepositorySnapshotsTask,
+		Payload:        tasks.DeleteRepositorySnapshotsPayload{RepoConfigUUID: repo.UUID},
+		OrgId:          repo.OrgID,
+		RepositoryUUID: repoUuid.String(),
+	})
+	assert.NoError(s.T(), err)
+	s.WaitOnTask(taskUuid)
+
+	// List(orgID string, paginationData api.PaginationData, filterData api.FilterData) (api.RepositoryCollectionResponse, int64, error)
```
rverdile, line 87: possibly left this here by mistake

pkg/tasks/delete_repository_snapshots_test.go
```diff
@@ -28,14 +29,73 @@ func TestDeleteSnapshotSuite(t *testing.T) {
 	suite.Run(t, new(DeleteRepositorySnapshotsSuite))
 }
 
+func (s *DeleteRepositorySnapshotsSuite) pulpClient() pulp_client.PulpClient {
+	return &s.MockPulpClient
+}
```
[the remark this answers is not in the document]
jlsherrill: I couldn't quite figure out how to make that work.  I think its this is a quirk around the way interface conversions work.  If you can get something like this work simply, let me know!
```
	var pulpClient *pulp_client.PulpClient
	a := pulp_client.NewMockPulpClient(t)
	pulpClient = a
```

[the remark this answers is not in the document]
rverdile: ah, I see what you mean!

test/integration/delete_test.go
rverdile: just tested it, you're right!

PR #356, opened 2023-08-10
Set value for IQE_MARKER_EXPRESSION for test collection

aadhikar:


mshriver:
@aadhikar the PR Validation fails because the title isn't matching expected format for this repo, need to prefix with `Refs HMS-####`.

Please put in draft until the iqe-hms-integration MR introducing the marker is merged, otherwise this change would result in no tests getting collected for the pipeline.

mshriver (changes requested):
Need to set parameter default, not envvar default

mshriver:
Going to cover this in #358

PR #364, opened 2023-08-16
Refs 2344: Add a job name stub for integration job

mshriver:
the IQE_PLUGINS parameter has invalid chars for a job name

## Summary

Auto promotion is failing because my updated job name doesn't meet job name specs with the plugin names.
https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/77500

### DIFF
--- a/deployments/testing-integration.yaml
+++ b/deployments/testing-integration.yaml
@@ -7,14 +7,15 @@ objects:
 - apiVersion: batch/v1
   kind: Job
   metadata:
-    name: content-sources-${IQE_PLUGINS}-${UID}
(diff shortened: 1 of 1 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-2344

jlsherrill (approved):

mshriver:
Bonfire and Openshift are happy when I run this job deployment locally, although the ephemeral cluster doesn't like our memory limit and request ratio.

Not changing that ratio as the intended target for this job is the stage and prod clusters.

PR #342, opened 2023-08-01
Fixes 2188: Add more detail in snapshot listing

Andrewgdewar:
## Summary
Adds more detail to Snapshots and Repository response: 

- Adds added_counts & removed_counts to the snapshots table/SnapshotResponse

- Adds last_snapshot_uuid and last_snapshot to the repository_configurations table/RepositoryResponse

## Testing steps
- Add a repository with snapshotting enabled. 
- Fetch/list/edit said repository after it has finished snapshotting, one should see the above new fields on the response.
- Call the snapshot list api, one should see the associated fields above on the response.

PR #371, opened 2023-08-23
Fixes 1968: Update RepositoryResponse with snapshot taskID

Andrewgdewar:
## Summary

Add a new parameter "last_snapshot_task_uuid" to the RepositoryResponse (repositoryList/Fetch/Edit responses)

When a repository enques a task to snapshot, the api will now return the above parameter.
Screen Shot 2023-08-23 at 2 11 47 PM

This UUID can be confirmed by checking the taskInfo in the Admin tasks view under the ViewDetails Modal  > UUID (for the associated task of course).

Screen Shot 2023-08-23 at 2 12 07 PM

## Suggested Testing steps
- Create or Edit a repository with snapshotting enabled, check the response in the network for the "last_snapshot_task_uuid". 
- Copy the ID from the network tab above and check it against the  in the tasksInfo modal in the associated task.


jlsherrill:
https://issues.redhat.com/browse/HMS-1968

compose_files/pulp/pulp-oci-images
rverdile, line 1: you can delete this

pkg/api/repositories.go
```diff
@@ -20,6 +20,7 @@ type RepositoryResponse struct {
 	MetadataVerification         bool     `json:"metadata_verification"`               // Verify packages
 	RepositoryUUID               string   `json:"-" swaggerignore:"true"`              // UUID of the dao.Repository
 	Snapshot                     bool     `json:"snapshot"`                            // Enable snapshotting and hosting of this repository
+	LastSnapshotTaskUUID         string   `json:"last_snapshot_task_uuid,omitempty"`   // UUID of the last snapshot task
```
rverdile, line 23: We don't use omitempty on the other fields. Do you prefer it be used for snapshots specifically?

pkg/handler/repositories.go
```diff
@@ -530,6 +530,11 @@ func (rh *RepositoryHandler) enqueueSnapshotEvent(c echo.Context, repositoryUUID
 			logger := tasks.LogForTask(taskID.String(), task.Typename, task.RequestID)
 			logger.Error().Msg("error enqueuing task")
 		}
+		if err == nil {
+			if err := rh.DaoRegistry.RepositoryConfig.UpdateLastSnapshotTask(taskID.String(), orgID, repositoryUUID); err != nil {
+				log.Warn().Msgf("Unable to update snapshotTask, this is likely a DB migration error: %v", err)
```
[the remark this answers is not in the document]
Andrewgdewar: Perhaps I could just call the error in a warn. 
Above confirms that the taskID should have been stored correctly, if I can't find it immediately after having it be set, I could only see a migration error causing it.

pkg/api/repositories.go
Andrewgdewar: For snapshot related things, I suppose it could be an empty string. I've done something similar [here](https://github.com/content-services/content-sources-backend/pull/342/files#diff-837db424b3bce5a001d8266958327c228eb1a97df5f7c23cf24cf55f52c1b6d5R26)

compose_files/pulp/pulp-oci-images
Andrewgdewar: UGh.. that keeps slipping in! 😞

pkg/api/repositories.go
rverdile: I was mostly just curious about it. I could see omitting fields making sense because some repositories will not have snapshotting enabled.

pkg/handler/repositories.go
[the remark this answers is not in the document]
rverdile: Hmm. I'm not sure I understand. An error here is caused whenever this column fails to update. I imagine for most users they'd be using this well after the database has been migrated, and so a migration error wouldn't be the cause.

rverdile (commented):
Which part of the UI is for? Specifically, where are we going to use the last snapshot task ID?

pkg/handler/repositories.go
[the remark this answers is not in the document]
Andrewgdewar: I guess I'll remove the comment too, the error could be caused by other things I suppose.

swadeley:
Hi @Andrewgdewar 

Please rebase

thank you

compose_files/pulp/pulp-oci-images
rverdile, line 1: delete this

pkg/dao/repository_configs.go
```diff
@@ -375,6 +375,22 @@ func (r repositoryConfigDaoImpl) Update(orgID, uuid string, repoParams api.Repos
...
 
+// UpdateLastSnapshotTask updates the RepositoryConfig with the latest SnapshotTask
+func (r repositoryConfigDaoImpl) UpdateLastSnapshotTask(taskUUID string, orgID string, repoUUID string) error {
+	result := r.db.Exec(`
+			UPDATE repository_configurations 
+			SET last_snapshot_task_uuid = ? 
+			WHERE repository_configurations.repository_uuid = ?`,
+		taskUUID,
+		repoUUID,
+	)
```
rverdile, line 386: you want to also filter by orgID here, right? It looks like you included it in the method params, but forgot to use it.

pkg/dao/repository_configs_test.go
```diff
@@ -1143,6 +1143,31 @@ func (suite *RepositoryConfigSuite) TestBulkDelete() {
...
+
+	err := seeds.SeedRepositoryConfigurations(suite.tx, repoConfigCount, seeds.SeedOptions{OrgID: orgID})
+	assert.Nil(t, err)
+
+	var uuids []string
+	err = suite.tx.Model(models.RepositoryConfiguration{}).Where("org_id = ?", orgID).Select("repository_uuid").Find(&uuids).Error
+	assert.NoError(t, err)
+	assert.Len(t, uuids, repoConfigCount)
+
+	taskUUID := "00000000-0000-1337-0000-000000000000"
```
rverdile, line 1160: you could also do `uuid.NewString()`

pkg/dao/repository_configs.go
Andrewgdewar: ⚠️  Good call out!! ugh.. so used to typescript screaming for unused variables.

rverdile (commented):
When I create a repository, I don't see the `last_snapshot_task_uuid`. And when I update a repository, I see the 2nd to last uuid, not the last uuid.

I suspect the response in the handler needs to be updated.

Andrewgdewar:
> 

I'll walk through this with you tomorrow, good catch.

Andrewgdewar:
Pushed the fix!

swadeley (approved):
lgtm now

PR #373, opened 2023-08-24
Refs 645: use only runAsUser

jlsherrill:
## Summary
I think that runAsNonroot is not needed (and may conflict)  if we specify runAsUser


jlsherrill:
closing in favor of https://github.com/content-services/content-sources-backend/pull/374

PR #377, opened 2023-08-29
Fixes 2419: ignore rbac check for org admins

jlsherrill:
## Summary
if rbac_org_admin_skip is set to true, then don't check rbac if a user identity header tells us the user is an org_admin

## Testing steps
1.  turn on rbac_enabled in config.yaml
2. run the app with skip turned on and the mock rbac service
```
RBAC_ORG_ADMIN_SKIP=true go run cmd/content-sources/main.go api mock_rbac
```
3. use this header script to generate org_admin headers:  https://gist.github.com/jlsherrill/880bb18c288be3916c8cda94db0efacd

4.  List/create repos via the api.  Any request generated with the new script should work.  Using the older script, only those users listed in 'mocks' -> 'rbac' section of config.yaml would be allowed, all others would be rejected

### DIFF
--- a/deployments/deployment.yaml
+++ b/deployments/deployment.yaml
@@ -211,6 +211,8 @@ objects:
                     name: cs-pulp-admin-password
                     key: password
                     optional: true
+              - name: RBAC_ORG_ADMIN_SKIP
+                value: ${{RBAC_ORG_ADMIN_SKIP}}
               - name: LOGGING_LEVEL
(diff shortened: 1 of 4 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-2419

rverdile (approved):
tested and working for read and write permissions

PR #385, opened 2023-09-08
Fixes 2543: use a null uuid for bad uuids

jlsherrill:
## Summary

instead of converting to text in queries, as
that has a performance penalty

## Testing steps

Tests pass

Here's the explain plan comparing two queries, the one with text(uuid) has a MUCH higher 'cost':
```
content=# explain SELECT * FROM repository_configurations WHERE (text(UUID) = '59a54071-06fb-45b6-b721-09bd0fb767cf' AND ORG_ID = '16791776') AND repository_configurations.deleted_at IS NULL ORDER BY repository_configurations.uuid LIMIT 1
;
                                                        QUERY PLAN                                                         
---------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=3529.23..3529.23 rows=1 width=172)
   ->  Sort  (cost=3529.23..3529.23 rows=1 width=172)
         Sort Key: uuid
         ->  Bitmap Heap Scan on repository_configurations  (cost=3197.68..3529.22 rows=1 width=172)
               Recheck Cond: (((org_id)::text = '16791776'::text) AND (deleted_at IS NULL))
               Filter: ((uuid)::text = '59a54071-06fb-45b6-b721-09bd0fb767cf'::text)
               ->  Bitmap Index Scan on repo_config_repo_org_id_deleted_null_unique  (cost=0.00..3197.68 rows=100 width=0)
                     Index Cond: ((org_id)::text = '16791776'::text)
(8 rows)
content=# explain SELECT * FROM repository_configurations WHERE (UUID = '59a54071-06fb-45b6-b721-09bd0fb767cf' AND ORG_ID = '16791776') AND repository_configurations.deleted_at IS NULL ORDER BY repository_configurations.uuid LIMIT 1
;
                                                       QUERY PLAN                                                       
------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.42..8.44 rows=1 width=172)
   ->  Index Scan using repository_configurations_pkey on repository_configurations  (cost=0.42..8.44 rows=1 width=172)
         Index Cond: (uuid = '59a54071-06fb-45b6-b721-09bd0fb767cf'::uuid)
         Filter: ((deleted_at IS NULL) AND ((org_id)::text = '16791776'::text))
(4 rows)
```

### DIFF
--- a/pkg/dao/task_info.go
+++ b/pkg/dao/task_info.go
@@ -42,9 +42,8 @@ func (t taskInfoDaoImpl) Fetch(orgID string, id string) (api.TaskInfoResponse, e
 
 	result := t.db.Table(taskInfo.TableName()+" AS t ").
 		Select(JoinSelectQuery).
-		Joins("LEFT JOIN repositories r on t.repository_uuid = r.uuid").
-		Joins("LEFT JOIN repository_configurations rc on r.uuid = rc.repository_uuid").
-		Where("text(t.id) = ? AND t.org_id = ?", id, orgID).First(&taskInfo)
+		Joins("LEFT JOIN repository_configurations rc on t.repository_uuid = rc.repository_uuid").
+		Where("t.id = ? AND t.org_id = ?", UuidifyString(id), orgID).First(&taskInfo)
 
 	if result.Error != nil {
 		if result.Error == gorm.ErrRecordNotFound {
@@ -69,8 +68,7 @@ func (t taskInfoDaoImpl) List(
 
 	filteredDB := t.db.Table(taskInfo.TableName()+" AS t ").
 		Select(JoinSelectQuery).
-		Joins("LEFT JOIN repositories r on t.repository_uuid = r.uuid").
-		Joins("LEFT JOIN repository_configurations rc on r.uuid = rc.repository_uuid").
+		Joins("LEFT JOIN repository_configurations rc on t.repository_uuid = rc.repository_uuid").
 		Where("t.org_id = ?", orgID)
 
 	if filterData.Status != "" {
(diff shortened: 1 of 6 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-2543

pkg/dao/task_info.go
```diff
@@ -42,9 +42,8 @@ func (t taskInfoDaoImpl) Fetch(orgID string, id string) (api.TaskInfoResponse, e
 
 	result := t.db.Table(taskInfo.TableName()+" AS t ").
 		Select(JoinSelectQuery).
-		Joins("LEFT JOIN repositories r on t.repository_uuid = r.uuid").
-		Joins("LEFT JOIN repository_configurations rc on r.uuid = rc.repository_uuid").
-		Where("text(t.id) = ? AND t.org_id = ?", id, orgID).First(&taskInfo)
+		Joins("LEFT JOIN repository_configurations rc on t.repository_uuid = rc.repository_uuid").
+		Where("t.id = ? AND t.org_id = ?", UuidifyString(id), orgID).First(&taskInfo)
```
Andrewgdewar, line 46: So this makes sense but what would happen if the the id was "not-a-proper-uuid"? 
I take it the UUIDifyString fucntion would return a uuid2.Nil. 
Would we want that in our database?
Would it be prudent to do an error check prior to the sql call?

jlsherrill: this is just for querying, we'd never insert a uuid.nil value.

jlsherrill: we could add an error check for each query to validate that its a proper UUID and return an empty set, but thats a bit more obtuse i think.  The bigger issue is that its easy to write a query that doesn't handle this well.  Maybe the longer term solution is to accept UUIDs at the api level?   I'll check if thats even possible but thats probably a bigger longer term solution and this is a good short term performance fix

Andrewgdewar (approved):
ackd

PR #387, opened 2023-09-11
Fixes 1963: nightly sync 

jlsherrill:
## Summary

Adds repository snapshotting to nightly tasks. When the snapshot feature is enabled in the config, nightly tasks will create a snapshot of all repositories with snapshot = true that haven't been snapshotted in the last 24 hours.

## Testing steps

Enable the snapshot feature in the config and create a repository with snapshotting enabled.
go run cmd/external-repos/main.go nightly-jobs triggers the sync task.
If there is a repository with snapshotting enabled that hasn't been snapshotted in the past 24 hours (including one that has no snapshots), then a snapshot task for that repository will be enqueued. Running the nightly-job will log that the snapshot task was enqueued.


jlsherrill:
https://issues.redhat.com/browse/HMS-1963

pkg/dao/repository_configs.go
```diff
@@ -173,6 +174,21 @@ func (r repositoryConfigDaoImpl) bulkCreate(tx *gorm.DB, newRepositories []api.R
 	}
 }
 
+func (p repositoryConfigDaoImpl) InternalOnly_ListReposToSnapshot() ([]models.RepositoryConfiguration, error) {
+	var dbRepos []models.RepositoryConfiguration
+	interval := fmt.Sprintf("%v hours", config.SnapshotInterval)
+	result := p.db.Where("snapshot IS TRUE").Joins("LEFT JOIN tasks on last_snapshot_task_uuid = tasks.id").
+		Where(p.db.Where("tasks.queued_at <= (current_date - cast(? as interval))", interval).
```
rverdile, line 181: is this wrapped `where(where(...))` intentional?

rverdile (commented):
If the snapshot fails, it should try again in 8 hours, right? I'm seeing that only snapshots older than 20 hours are queued.

pkg/dao/repository_configs.go
jlsherrill: yes, it means its selecting:

Snapshot is true AND  ( "last task is old" OR "status is not complete, pending, or running" OR "Never had a snapshot task" ) 

if it wasn't wrapped, it ends up like:   Snapshot is true AND  "last task is old" OR "status is not complete, pending, or running" OR "Never had a snapshot task"   which when grouped via precedence means:

( "Snapshot is true" AND  "last task is old")  OR "status is not complete, pending, or running" OR "Never had a snapshot task"

jlsherrill:
It should, i'll dig into why not

jlsherrill:
rebased and added a new parameter: OPTIONS_ALWAYS_RUN_CRON_TASKS.  When set to true: `OPTIONS_ALWAYS_RUN_CRON_TASKS=true`  this will cause all repos to be introspected and snapshotted anytime the cron job runs, ignoring any time constraint normally in place

swadeley:
Thank you.

PR #389, opened 2023-09-13
Refs 1963: Bump mem request to 200Mi

swadeley:
## Summary

The kafka consumer init container is crashing while trying to introspect repos (On startup, all repos in the database are introspected.). This problem only occurs in ephemeral env.

As per @jrusz 's suggestion, this issue could be due to memory taking to long to be increased, so bumping the request size to see if that helps to get the memory increased before the pod crashes. 

## Testing steps

Check for content-sources-backend-kafka-consumer restarting multiple times.

### DIFF
--- a/deployments/deployment.yaml
+++ b/deployments/deployment.yaml
@@ -358,7 +358,7 @@ parameters:
   - name: MEMORY_LIMIT
     value: 1Gi
   - name: MEMORY_REQUESTS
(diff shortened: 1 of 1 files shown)


jlsherrill (approved):

jlsherrill:
https://issues.redhat.com/browse/HMS-1963

PR #395, opened 2023-09-20
Fixes 2631: handle bucket hostname with proto

jlsherrill:
## Summary
Sometimes the objectstore  hostname has a protocol/scheme, sometimes it doesn't.  Handle both cases (with a test!)

## Testing steps
snapshotting works in ephemeral, we'll test stage after merge

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -5,6 +5,7 @@ import (
 	"crypto/x509"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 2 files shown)


Andrewgdewar (approved):
Confirmed working in ephemeral, all tests/checks passing.

jlsherrill:
merged with blessing from @swadeley

PR #397, opened 2023-09-21
Fixes 2381: specify if name or URL is duplicated

rverdile:
## Summary
When creating a repository, if a name or URL is already being used, the error message should say which field is duplicated. In a previous change, we renamed constraints that were being used to determine which field is duplicated. I've updated the function with the new constraint names.

## Testing steps
1. Create a repository
2. Create a repository again with the same name.
3. Check the error message. It should tell you "name" is the duplicated field.
4. Create a repository again with the a different name, but same URL.
5. Check the error message. It should tell you "URL" is the duplicated field.

### DIFF
--- a/pkg/dao/repository_configs.go
+++ b/pkg/dao/repository_configs.go
@@ -43,11 +43,11 @@ func DBErrorToApi(e error) *ce.DaoError {
 	if ok {
 		if pgError.Code == "23505" {
 			switch pgError.ConstraintName {
-			case "repo_and_org_id_unique":
+			case "repo_config_repo_org_id_deleted_null_unique":
 				dupKeyName = "URL"
 			case "repositories_unique_url":
 				dupKeyName = "URL"
-			case "name_and_org_id_unique":
+			case "repo_config_name_deleted_org_id_unique":
 				dupKeyName = "name"
 			}
 			return &ce.DaoError{BadValidation: true, Message: "Repository with this " + dupKeyName + " already belongs to organization"}
--- a/pkg/dao/repository_configs_test.go
+++ b/pkg/dao/repository_configs_test.go
@@ -16,7 +16,6 @@ import (
 	mockExt "github.com/content-services/content-sources-backend/pkg/test/mocks/mock_external"
 	"github.com/content-services/yummy/pkg/yum"
 	"github.com/google/uuid"
-	"github.com/jackc/pgx/v5/pgconn"
 	"github.com/lib/pq"
 	"github.com/openlyinc/pointy"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


pkg/dao/repository_configs.go
```diff
@@ -43,11 +43,11 @@ func DBErrorToApi(e error) *ce.DaoError {
 	if ok {
 		if pgError.Code == "23505" {
 			switch pgError.ConstraintName {
-			case "repo_and_org_id_unique":
+			case "repo_config_repo_org_id_deleted_null_unique":
 				dupKeyName = "URL"
 			case "repositories_unique_url":
 				dupKeyName = "URL"
-			case "name_and_org_id_unique":
+			case "repo_config_name_deleted_org_id_unique":
```
rverdile, line 50: I was contemplating removing this test because it only really tests the implementation of the `DBErrorToAPI` function. It didn't actually catch this bug. Thoughts?

jlsherrill:
https://issues.redhat.com/browse/HMS-2381

jlsherrill:
This needs a rebase still

pkg/dao/repository_configs.go
jlsherrill: did you mean to point to specific test?

rverdile: yeah whoops, I meant here: https://github.com/content-services/content-sources-backend/blob/af65a0279e05604169e35b3fe08760fd01b460f8/pkg/dao/repository_configs_test.go#L1559

swadeley:
Hi

tests OK

`HTTP response body: {"errors":[{"status":400,"title":"Error creating repository","detail":"Repository with this URL already belongs to organization"}]}`


`HTTP response body: {"errors":[{"status":400,"title":"Error creating repository","detail":"Repository with this name already belongs to organization"}]}
`

lgtm

pkg/dao/repository_configs.go
jlsherrill: IDK, i feel like it still has some value?

jlsherrill: oh, i see what you're saying, yeah, i think i we have tests that test the behavior and the error response, that is a lot more valuable.

jlsherrill: I'd say lets beef up TestRepositoryCreateAlreadyExists  to check the message, and make sure a test for name & url is covered, and then remove TestDBErrorToApi

rverdile: I already added a test to TestRepositoryCreateAlreadyExists. Are you saying I should add something else to it?

jlsherrill: yes, i meant testing the actual message, similar to what TestDBErrorToApi is doing.  This would help tell the difference if its name vs url instead of just checking for daoError.BadValidation)

jlsherrill: and the check could be something like    message.includes?("name")  instead of doing a literal string comparison.

jlsherrill: sorry, maybe i should have drank more coffee :)   Feel free to delete that test, otherwise ACK!

rverdile: ha, no worries :)

jlsherrill (approved):

PR #406, opened 2023-09-26
Fixes 2668: use newer yummy

jlsherrill:
## Summary
to limit the size of an xml file we can process

## Testing steps

Try to introspect a repo: https://raw.githubusercontent.com/lpardoRH/custom-repo/zip_bomb2/
You will likely run out of memory in ephemeral (maybe locally).  With this fix it will simply return zero packages (as its not really an xml file).   Right now this does not trigger an error which is a result of underlying libraries.  Since this edge case is an attempted attack and not a legitimate case, this is okay.  A partial (or corrupted) XML file would error as expected

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.20
 
 require (
 	github.com/ProtonMail/go-crypto v0.0.0-20230923063757-afb1ddc0824c
-	github.com/content-services/yummy v1.0.5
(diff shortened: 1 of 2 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-2668

rverdile (approved):
tested and working!

swadeley:
lgtm

PR #407, opened 2023-09-29
Fixes 2245: add red hat repo for nightly snapshotting

jlsherrill:
## Summary

Adds an intial set of Red hat repositories that is loaded from a json file.  
Adds a single gpg key loaded for all red hat repos
These repos get created into the repository_configs table with an org_id of -1

## Testing steps

load in ephemeral, or run `make repos-import`

1. curl  http://localhost:8000/api/content-sources/v1.0/repositories/  for some new org
  By default, nothing should be returned
2.  curl http://localhost:8000/api/content-sources/v1.0/repositories/?origin=red_hat
  you should get all the red hat repos
3.  create some repo
4. curl http://localhost:8000/api/content-sources/v1.0/repositories/?origin=red_hat,external
  you  should get both custom and redhat repos


db/migrations/20230922154941_add_config_constraint.down.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+alter table repository_configurations drop constraint IF EXISTS repo_uuid_and_org_id_del_at_unique;
```
jlsherrill, line 3: needed to add this constraint in order to do the insert or update

db/migrations/20230922154941_add_config_constraint.up.sql
```diff
@@ -0,0 +1,17 @@
...
+
+BEGIN
+alter table repository_configurations ADD CONSTRAINT repo_uuid_and_org_id_del_at_unique UNIQUE (repository_uuid, org_id, deleted_at);
+EXCEPTION
+    WHEN duplicate_table THEN  -- postgres raises duplicate_table at surprising times. Ex.: for UNIQUE constraints.
+    WHEN duplicate_object THEN
+      RAISE NOTICE 'Table constraint repo_uuid_and_org_id_del_at_unique already exists';
+END;
+
+END $$;
```
jlsherrill, line 14: This was the only way i could see to check if a constraint already existed to make this idempotent

jlsherrill:
https://issues.redhat.com/browse/HMS-2245

db/migrations/20230922154941_add_config_constraint.up.sql
```diff
@@ -0,0 +1,17 @@
+BEGIN;
+
+DO $$
+BEGIN
+
+BEGIN
+alter table repository_configurations ADD CONSTRAINT repo_uuid_and_org_id_del_at_unique UNIQUE (repository_uuid, org_id, deleted_at);
```
jlsherrill, line 7: I had to add this constraint in order for the create or update query to work.

db/migrations/20230922154941_add_config_constraint.up.sql
rverdile: what about dropping the constraint before adding it?

jlsherrill: This constraint may actually not be needed, the test failure is related to the unique index for the same fields.  anyway i'm playing around with it

jlsherrill: yep, it wasn't needed at all.  removed!

db/migrations/20230922154941_add_config_constraint.up.sql
jlsherrill: nevermind

jlsherrill:
Realized i need to use the red hat cdn cert for syncing, will work on adding that

jlsherrill:
this seems to be failing with: 

{"level":"panic","error":"ERROR: invalid reference to FROM-clause entry for table \"repository_configurations\" (SQLSTATE 42P01)","time":"2023-10-03T15:05:35Z","message":"Failed to save repositories"}
panic: Failed to save repositories

will investigate

rverdile (commented):
> curl http://localhost:8000/api/content-sources/v1.0/repositories/ for some org
By default, nothing should be returned

Here it says nothing should be returned when I curl this. If I curl this, I get both red hat repos and whatever repos have been created for the org. I would expect to get both (or only custom), but which is the behavior you intended?

jlsherrill:
nothing should be returned for a 'new org'.  do you have the NewRepositoryFiltering feature turned on? because that will affect the behavior.  With it turned on, you'll get both.  WIth it turned off (default), you'll get only get external (which for a new org would have none).

jlsherrill:
updated the testing steps to try to make that more clear

pkg/external_repos/redhat_repos.go
```diff
@@ -0,0 +1,95 @@
...
+		return err
+	}
+	gpgKey, err := redHatGpgKey()
+	if err != nil {
+		return err
+	}
+	for _, r := range repos {
+		r.GpgKey = gpgKey
+		_, err = rhr.daoReg.RepositoryConfig.InternalOnly_RefreshRedHatRepo(r.ToRepositoryRequest())
+		if err != nil {
```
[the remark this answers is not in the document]
jlsherrill: It gets called anytime we do a new deployment/code push:  https://github.com/content-services/content-sources-backend/blob/main/deployments/deployment.yaml#L44-L48 .   I guess i could rename it CreateOrUpdateRedHatRepos() if thats more clear.  Let me know what you think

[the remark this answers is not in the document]
rverdile: The name is good as is. I just wanted to make sure I understood the change.

pkg/external_repos/redhat_repos.json
```diff
@@ -0,0 +1,37 @@
+[
+    {
+        "name": "Red Hat Ansible Engine 2 for RHEL 8 x86_64 (RPMs)",
+        "url": "https://cdn.redhat.com/content/dist/layered/rhel8/x86_64/ansible/2/os",
```
rverdile, line 4: If I update a URL here, it will create the repository in the database, but returns an error creating the repository configuration.

`"ERROR: duplicate key value violates unique constraint \"repo_config_name_deleted_org_id_unique\" (SQLSTATE 23505)`

Do we anticipate needing to handle the case where a to URL to one of these repositories changes, but it's really the same repository?

jlsherrill: I wrote this thinking we'd want to change everything BUT the url, so this behavior is intentional.  That said, I dont' know what we'll need to do in the future.  So for now i'd say lets keep this as is, and support url changes in the future if needed.

pkg/dao/repository_configs.go
```diff
@@ -648,6 +649,46 @@ func isTimeout(err error) bool {
...
+
+	newRepoConfig.OrgID = config.RedHatOrg
+	newRepo.Origin = config.OriginRedHat
+
+	result := r.db.Clauses(clause.OnConflict{
+		Columns:   []clause.Column{{Name: "url"}},
+		DoUpdates: clause.AssignmentColumns([]string{"origin"})}).Create(&newRepo)
+	if result.Error != nil {
+		return nil, result.Error
+	}
```
rverdile, line 667: Does this query mean "if a repository with this URL already exists, rather than error, update its origin column to `red_hat`"?

jlsherrill: yep!

pkg/tasks/repository_snapshot.go
```diff
@@ -257,18 +262,33 @@ func (sr *SnapshotRepository) findOrCreatePulpRepo(repoConfigUUID string, remote
...
+		caCert = pointy.Pointer(string(ca))
+	}
+
 	remoteResp, err := sr.pulpClient.GetRpmRemoteByName(repoConfig.UUID)
 	if err != nil {
 		return "", err
 	}
 	if remoteResp == nil {
-		remoteResp, err = sr.pulpClient.CreateRpmRemote(repoConfig.UUID, repoConfig.URL)
+		remoteResp, err = sr.pulpClient.CreateRpmRemote(repoConfig.UUID, repoConfig.URL, clientCertPair, clientCertPair, caCert)
```
rverdile, line 286: I tested syncing and it's working, so I think there's no issue here. Just a question. You pass `clientCertPair` for both `clientCert` and `clientKey`. Is the "pair" a combination of the two? And pulp can parse it either way?

jlsherrill: yep, the pair is both combined.  Pulp (and most http clients) will pull out the right one.  This is how we're doing it with introspection so i just kept it the same.

rverdile (approved):
tested and looks good!

swadeley:
Hii

I cant get this to deploy:

content-sources-backend-kafka-consumer-fc68c66b7-tfqct           0/1     Init:ImagePullBackOff
content-sources-backend-service-897844f57-jl4zg                  0/2     Init:ImagePullBackOff

swadeley:
Hi

deployed new images

```
In [5]: app.content_sources.rest_client.repositories_api.list_repositories(origin='red_hat')
2023-10-10 14:48:00.411 [    INFO] [iqe.base.rest_client] REST: METHOD=GET, request_id=shDW6AGqfvP9rRlB0lTYr8SNkVyK1aTf, params=[('origin', 'red_hat')]
Out[5]: 
{'data': [{'account_id': '',
           'content_type': 'rpm',
           'distribution_arch': 'x86_64',
           'distribution_versions': ['8'],
           'failed_introspections_count': 0,
           'gpg_key': '-----BEGIN PGP PUBLIC KEY BLOCK-----\n'
                      '\n'
                      'mQINBErgSTsBEACh2A4b0O9t+vzC9VrVtL1AKvUWi9OPCjkvR7Xd8DtJxeeMZ5eF\n
<big-snip>
           'name': 'Red Hat Enterprise Linux 9 for x86_64 - AppStream (RPMs)',
           'org_id': '-1',
           'origin': 'red_hat','


```

`In [6]: app.content_sources.rest_client.repositories_api.list_repositories(origin='external')
2023-10-10 14:52:22.096 [    INFO] [iqe.base.rest_client] REST: METHOD=GET, request_id=9Yc3132QeURHAUyhKdnlIQXZYqAZvf1h, params=[('origin', 'external')]
Out[6]: 
{'data': [{'account_id': '0369233',
           'content_type': 'rpm',
           'distribution_arch': 'any',
           'distribution_versions': ['any'],
           'failed_introspections_count': 0,
           'gpg_key': '',`

and this gets all repos:
`In [9]: app.content_sources.rest_client.repositories_api.list_repositories(origin='external,red_hat')
`

Thank you

PR #415, opened 2023-10-04
Refs 1581: debug register listener

rverdile:
## Summary
Adding debugs logs to see if register listener is hanging

### DIFF
--- a/pkg/tasks/queue/pgqueue.go
+++ b/pkg/tasks/queue/pgqueue.go
@@ -721,11 +721,14 @@ func (p *PgQueue) ListenForCancel(ctx context.Context, taskID uuid.UUID, cancelF
 
 	// Register a channel for the task where a notification can be sent to cancel the task
 	channelName := getCancelChannelName(taskID)
+	// TODO remove debug logs. checking if channel register is hanging
+	logger.Debug().Msg("ListenForCancel: preparing register channel")
 	_, err = conn.Conn().Exec(ctx, "listen "+channelName)
 	if err != nil {
 		logger.Error().Err(err).Msg("ListenForCancel: error registering channel")
 		return
 	}
+	logger.Debug().Msg("ListenForCancel: finished register channel")
 
 	// When the function returns, unregister the channel
 	defer func(conn *pgx.Conn) {


pkg/tasks/queue/pgqueue.go
```diff
@@ -710,22 +710,31 @@ func (p *PgQueue) RemoveAllTasks() error {
...
 	_, err = conn.Conn().Exec(ctx, "listen "+channelName)
 	if err != nil {
+		//if !errors.Is(ErrNotRunning, context.Cause(ctx)) {
+		//	logger.Error().Err(err).Msg("ListenForCancel: error registering channel")
+		//}
 		logger.Error().Err(err).Msg("ListenForCancel: error registering channel")
+
 		return
 	}
+	logger.Debug().Msg("ListenForCancel: listening on channel")
```
rverdile, line 737: We're seeing an issue where the listener fails to register because the context is canceled. This could happen if the task finishes very quickly. This could also happen if this Exec call is hanging and then unblocking when the task is finished (because the context was canceled). 

I left that error in with some debug logs, the idea being to make sure this call the database isn't hanging.

rverdile: removed this for now

pkg/tasks/queue/pgqueue.go
```diff
@@ -721,11 +721,14 @@ func (p *PgQueue) ListenForCancel(ctx context.Context, taskID uuid.UUID, cancelF
...
 	// Register a channel for the task where a notification can be sent to cancel the task
 	channelName := getCancelChannelName(taskID)
+	// TODO remove debug logs. checking if channel register is hanging
+	logger.Debug().Msg("ListenForCancel: preparing register channel")
 	_, err = conn.Conn().Exec(ctx, "listen "+channelName)
 	if err != nil {
 		logger.Error().Err(err).Msg("ListenForCancel: error registering channel")
 		return
 	}
+	logger.Debug().Msg("ListenForCancel: finished register channel")
```
jlsherrill, line 731: would debug even show up in kilbana?  wouldn't it need to be warn?

jlsherrill: in stage i mean

rverdile: I see all the log levels when I  search through kibana

jlsherrill: oh interesting!  ACK

jlsherrill (approved):

PR #417, opened 2023-10-05
Refs 1581: cancel task with cause

rverdile:
## Summary
We're seeing an issue where ListenForCancel is erroring because context is canceled. Likely happening because the task is finishing before the listener is ready to wait for a notification.

You can see the timestamps are very close.
```
Oct 5, 2023 @ 08:01:11.935 --> time right before registering listener

Oct 5, 2023 @ 08:01:12.006 --> time failed because context cancelled
```

There are many cases where this does not happen and the listener has time to finish registering, which should rule out an issue with the register command hanging.

This adds `context.CancelWithCause` so that we can identify when the context is canceled because the task is finished. If the task is finished, we simply ignore these errors.

## Testing steps
You might try to reproduce this locally, but I was not able to.

### DIFF
--- a/pkg/tasks/queue/pgqueue.go
+++ b/pkg/tasks/queue/pgqueue.go
@@ -714,6 +714,7 @@ func (p *PgQueue) ListenForCancel(ctx context.Context, taskID uuid.UUID, cancelF
 	logger := zerolog.Ctx(ctx)
 	conn, err := p.Pool.Acquire(ctx)
 	if err != nil {
+		// If the task is finished before listen is initiated, a context canceled error is expected
 		if !errors.Is(ErrNotRunning, context.Cause(ctx)) {
 			logger.Error().Err(err).Msg("ListenForCancel: error acquiring connection")
 		}
--- a/pkg/tasks/queue/pgqueue.go
+++ b/pkg/tasks/queue/pgqueue.go
@@ -710,11 +710,13 @@ func (p *PgQueue) RemoveAllTasks() error {
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/tasks/queue/pgqueue.go
```diff
@@ -710,11 +710,13 @@ func (p *PgQueue) RemoveAllTasks() error {
 	return nil
 }
 
-func (p *PgQueue) ListenForCancel(ctx context.Context, taskID uuid.UUID, cancelFunc context.CancelFunc) {
+func (p *PgQueue) ListenForCancel(ctx context.Context, taskID uuid.UUID, cancelFunc context.CancelCauseFunc) {
 	logger := zerolog.Ctx(ctx)
 	conn, err := p.Pool.Acquire(ctx)
 	if err != nil {
-		logger.Error().Err(err).Msg("ListenForCancel: error acquiring connection")
+		if !errors.Is(ErrNotRunning, context.Cause(ctx)) {
```
jlsherrill, line 717: You might add a comment to explain "sometimes a task can finish before the Listen is initiated" for future self (or me)

jlsherrill:
I was able to reproduce locally by adding ```time.Sleep(100 * time.Millisecond)``` right before the listen for cancel.  With this change the issues is fixed!    just one small comment and then i think this is good

pkg/tasks/queue/pgqueue.go
rverdile: added a comment

jlsherrill (approved):

PR #418, opened 2023-10-05
Fixes 2588: Existing popular repositories not found error

Andrewgdewar:
## Summary

Fixes popular repositories not having a UUID added to their list due to a gorm default limit change. 
Previously, when limit was set to 0  (or not set, which defaults to 0) the limit was effectively infinite, the update changes that so that the limit is now actually 0. 
This made it so that on query of the repositoryConfig.List method (with the default Limit of 0), UUID updates for each item on the popular repositories list were not being applied. As "nothing" was found in the Data, despite a count of 1 being returned.

## Testing steps

- Add a popular repository
- Check that the UUID is now returned in the response
- This will also be evident in the UI by the "Remove" button being visible on the popular repositories view.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ cmd/test/
 # Prometheus
 prometheus/data
 configs/prometheus.yaml
+
+# Pulp Oci Images
+compose_files/pulp/pulp-oci-images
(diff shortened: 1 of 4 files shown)


compose_files/pulp/pulp-oci-images
rverdile, line 1: Remove this. If you haven't, you can also delete it locally. It's not used anymore.

jlsherrill (approved):

swadeley:
Hi

Remove buttons work again. Thank you.

PR #427, opened 2023-10-12
Fixes 2777: removed duplicate key from config example

xbhouse:
## Summary

Removed duplicate pulp key in config.yaml.example that was causing issues when running the app for the first time.


rverdile (approved):
nice!

rverdile:
merging, QE wouldn't need to look at this

PR #428, opened 2023-10-12
Refs 2776: capture pool stats

jlsherrill:


jlsherrill:
https://issues.redhat.com/browse/HMS-2776

jlsherrill:

rverdile (commented):
Since you said you were trying to think of a different way to do this, what if you added some fields to the logger? Kinda like here: https://github.com/content-services/content-sources-backend/blob/main/pkg/tasks/introspect.go#L50-L56

Maybe it could be done conditionally based on some config.

jlsherrill:
I'm thinking i could also stick the logging in the pgx_pool_wrapper instead of here

PR #433, opened 2023-10-17
Fixes 2738: support SCRAM-SHA-512 sasl mech for kafka

jlsherrill:
## Summary

based on  https://github.com/RedHatInsights/insights-results-aggregator/pull/1864/files

## Testing steps

We may need to just merge to stage to verify this works.

### DIFF
--- a/pkg/config/event.go
+++ b/pkg/config/event.go
@@ -1,11 +1,17 @@
 package config
 
 import (
+	"crypto/sha512"
 	"os"
 	"strings"
 
+	"github.com/RedHatInsights/insights-operator-utils/tls"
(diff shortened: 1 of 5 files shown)


pkg/config/event.go
```diff
@@ -67,3 +73,65 @@ func readEnv(key string, def string) string {
 	}
 	return value
 }
+
+func SetupNotifications() {
```
jlsherrill, line 77: i thought this method belonged here a bit more

jlsherrill:
https://issues.redhat.com/browse/HMS-2738

rverdile (approved):
I guess I can't test this, but it looks fine

PR #414, opened 2023-10-03
Fixes 2615: add config.repo endpoint

rverdile:
## Summary
- Adds endpoint, `repositories/:repo_uuid/snapshots/:snapshot_uuid`/config.repo that returns config.repo file for a particular snapshot.
- Adds snapshot UUID and content URL to snapshot response.
- Adds pulp content path base to cache after first query to pulp.

## Testing steps
1. Create a repository with a snapshot
2. Once the snapshot is created, list the snapshots for that repo to get the UUID of the snapshot.
3. GET the new endpoint.
4. It should return plain text containing the config.repo as described in the JIRA card.

PR #434, opened 2023-10-19
Refs 2615: don't propagate canceled context

rverdile:
## Summary
In PR #414, I mistakenly set context on a shared struct. This means that unless the context is manually reset by using `InitializePulpClient` before calling the `repoConfig.List` or `repoConfig.Fetch` dao methods, the request will fail with

`{"errors":[{"status":500,"title":"Error fetching repository","detail":"Get \"http://cs-pulp-api-svc.content-sources-stage.svc.cluster.local:24817/pulp/api/v3/status/\": context canceled"}]}`

This changes makes it so `context.TODO()` is used instead of the propagated context. This is only a quick and temporary fix to unblock things for now.

## Testing steps
1. Fetch a repository
2. Introspect the same repository
3. Without this change, the introspection will fail with the 500 error.

### DIFF
--- a/pkg/dao/repository_configs.go
+++ b/pkg/dao/repository_configs.go
@@ -82,7 +82,7 @@ func (r *repositoryConfigDaoImpl) InitializePulpClient(ctx context.Context, orgI
 		return err
 	}
 
-	pulpClient := pulp_client.GetPulpClientWithDomain(ctx, domainName)
+	pulpClient := pulp_client.GetPulpClientWithDomain(context.TODO(), domainName)
 	r.pulpClient = pulpClient
 	return nil
 }
--- a/pkg/dao/snapshots.go
+++ b/pkg/dao/snapshots.go
@@ -161,7 +161,7 @@ func (sDao *snapshotDaoImpl) InitializePulpClient(ctx context.Context, orgID str
 		return err
 	}
 
-	pulpClient := pulp_client.GetPulpClientWithDomain(ctx, domainName)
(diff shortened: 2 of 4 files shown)


swadeley:
Hello


```
Type
    introspect
Status
    completed
Queued At
    19 Oct 2023 20:51 UTC+01:00
Started At
    19 Oct 2023 20:51 UTC+01:00
Finished At
    19 Oct 2023 20:51 UTC+01:00
Error
    None
```

lgtm

pkg/dao/repository_configs.go
```diff
@@ -82,7 +82,7 @@ func (r *repositoryConfigDaoImpl) InitializePulpClient(ctx context.Context, orgI
 		return err
 	}
 
-	pulpClient := pulp_client.GetPulpClientWithDomain(ctx, domainName)
+	pulpClient := pulp_client.GetPulpClientWithDomain(context.Background(), domainName)
```
jlsherrill, line 85: I wonder if context.Todo() would make more sense here?  I think long term we may want to refactor such that the dao is created per request, so that we can get better request logging with a request id, and thus the context would need to be passed per request to the dao layer.

rverdile: I can change it to context.Todo(). Either way it was only meant to be an immediate short term fix because this is breaking things. My intention was to refactor.

swadeley:
Works, UI and API crud, thank you

swadeley (approved):
lgtm

PR #435, opened 2023-10-19
Fixes 2702,2887: Add orgID to snapshot list endpoint

Andrewgdewar:
## Summary

HMS-2702
- Adds orgID as a requirement to snapshot list

HMS 2887
- Allows the package list and snapshotList endpoint to return red hat repository (orgID of "-1") related data.

Additional changes: 
- Adds the new variable "any" to the "fetchRepoConfig" method.
- Updated dao methods using the above function as needed, create/delete/update are now prevented from taking effect on red_hat repositories.

## Testing steps

- Easiest testing method would be to checkout the accompanied front-end PR [here](https://github.com/content-services/content-sources-frontend/pull/156), and follow steps there-in.


jlsherrill:
https://issues.redhat.com/browse/HMS-2702

pkg/dao/snapshots.go
```diff
@@ -71,12 +71,13 @@ func (sDao *snapshotDaoImpl) List(repoConfigUuid string, paginationData api.Pagi
 	order := convertSortByToSQL(paginationData.SortBy, sortMap, "created_at asc")
 
 	filteredDB := sDao.db.
-		Where("snapshots.repository_configuration_uuid = ?", UuidifyString(repoConfigUuid))
+		Model(&models.Snapshot{}).
+		Joins("JOIN repository_configurations ON repository_configuration_uuid = repository_configurations.uuid").
+		Where("repository_configurations.org_id = ?", orgID).
+		Where("repository_configuration_uuid = ?", UuidifyString(repoConfigUuid))
```
rverdile, line 77: I think we can leave it if there's not a performance difference, but I wonder, does it make sense to just filter by orgID [here](https://github.com/content-services/content-sources-backend/blob/main/pkg/dao/snapshots.go#L56-L58) instead of joining the two tables?

Andrewgdewar: My reasoning was there was a purpose-built error check to see if the repoUUID existed or no, so I still wanted that error check to occur. 
If you pass in the orgId above you would get an error when the desired outcome (in my mind) would simply be an empty list.

rverdile: I agree that if there are no matching snapshots, but the repo config is found, there should be an empty list. I think there's a couple ways you could do that. For example, you could query all of the ones associated to a particular repo config UUID, if there aren't any, return an empty list.

It probably makes sense to filter by org ID [here](https://github.com/content-services/content-sources-backend/blob/main/pkg/dao/snapshots.go#L56-L58) either way though, because the repo config should belong to the requester. If it doesn't, there's no need to do anything else.

Andrewgdewar: oh hmm, I am working on Red Hat Repositories at the moment, I'll be making some changes to this PR to support that and will see about adding the orgId.

Andrewgdewar: Okay, I added this, along with some functionality around ensuring redhat repositories will be returned.

jlsherrill:
https://issues.redhat.com/browse/HMS-Fixes

pkg/dao/repository_configs.go
```diff
@@ -90,6 +90,11 @@ func (r *repositoryConfigDaoImpl) InitializePulpClient(ctx context.Context, orgI
 func (r repositoryConfigDaoImpl) Create(newRepoReq api.RepositoryRequest) (api.RepositoryResponse, error) {
 	var newRepo models.Repository
 	var newRepoConfig models.RepositoryConfiguration
+
+	if newRepoConfig.OrgID == config.RedHatOrg {
+		return api.RepositoryResponse{}, errors.New("Creating of Red Hat repositories is not permitted")
+	}
+
```
rverdile, line 97: This should be checking `newRepoReq.OrgID` instead of `newRepoConfig.OrgID` (which hasn't been set yet).

rverdile: Although I am wondering if this should instead be checked at the handler level and throw an error before getting to here

rverdile: Also you would have to do same thing for bulk create

pkg/dao/repository_configs.go
```diff
@@ -383,19 +388,26 @@ func (r repositoryConfigDaoImpl) Fetch(orgID string, uuid string) (api.Repositor
 	return repo, nil
 }
 
-func (r repositoryConfigDaoImpl) fetchRepoConfig(orgID string, uuid string) (models.RepositoryConfiguration, error) {
+// fetchRepConfig: "any" allows the fetching of red_hat repositories
+func (r repositoryConfigDaoImpl) fetchRepoConfig(orgID string, uuid string, any bool) (models.RepositoryConfiguration, error) {
```
rverdile, line 392: Instead of calling this boolean "any", how about "includeRedHatRepos" or something like that? Since that's specifically what it's deciding.

Also, "any" in Go is a keyword and alias for the type interface{}

pkg/dao/snapshots_test.go
```diff
@@ -215,7 +289,7 @@ func (s *SnapshotsSuite) TestListNotFound() {
 
 	s.createSnapshot(rConfig)
 
-	collection, total, err := sDao.List("bad-uuid", pageData, filterData)
+	collection, total, err := sDao.List(rConfig.OrgID, "bad-uuid", pageData, filterData)
```
rverdile, line 292: Could you add an additional test to "TestListNotFound" that creates has a bad org_id, but has a good repository uuid? This way we have to test to check that the query is filtered by org_id.

Unless you have a test elsewhere for that and I missed it.

pkg/dao/repository_configs.go
Andrewgdewar: Yeah, I probably should have expected that. will do.

pkg/dao/repository_configs.go
Andrewgdewar: I'll do the same for bulk delete, I think the idea was to prevent the creation of redHat repositories at the DAO layer explicitly, as we use a separate way to create those repositories today.

Andrewgdewar: Added a check to bulk delete. 

> This should be checking `newRepoReq.OrgID` instead of `newRepoConfig.OrgID` (which hasn't been set yet).
Good catch thank you! 
I'll add a test for this.

pkg/dao/snapshots_test.go
Andrewgdewar: Added TestListNotFoundBadOrgId! 
Thanks for the call out! 🍻

rverdile:
I'm not sure why, but it looks like most of the test didn't run. Do you see that too?

swadeley:
> I'm not sure why, but it looks like most of the test didn't run. Do you see that too?

Hi, `test_validate_repo_parameters`  filed because the same repo was used in the previous filter test and not cleaned up. I have added another repo with invalid metadata to the filter test to avoid this conflict.

rverdile (commented):
It says in your summary that create/update/delete do not have an affect for red hat repositories. Right now, if my org ID is -1, I can delete/update but not create. Obviously, that's not a real use case, but I feel the behavior should probably be consistent unless there's a reason for it not to be.

Do we expect to want to delete/update red hat repositories from the API? Or should that be blocked the way create is blocked?

pkg/dao/repository_configs.go
```diff
@@ -587,7 +606,7 @@ func (r repositoryConfigDaoImpl) bulkDelete(tx *gorm.DB, orgID string, uuids []s
 		var err error
 		var repoConfig models.RepositoryConfiguration
 
-		if repoConfig, err = r.fetchRepoConfig(orgID, uuids[i]); err != nil {
+		if repoConfig, err = r.fetchRepoConfig(orgID, uuids[i], false); err != nil {
```
rverdile, line 609: If I try to bulk delete a red hat repo, I correctly get the "not found" error. BUT it's returning a 500 response code instead of a 404

jlsherrill:
@swadeley we've added a command to run when deploying to ephemeral that should snapshot a small rhel repo (Red Hat Ansible Engine 2 for RHEL 8 x86_64 (RPMs))   Hopefully there's enough memory for it to snapshot.  Note that it doesn't 'block' deployment on the snapshot, it happens in the background.  

I'll spin it up to verify as well once the image is built

jlsherrill:
I've updated the init container to first wait until pulp is up before triggering the task.  I haven't been able to test this in ephemeral, as i can't seem to get the deployment.yaml changes to take effect when pointing to it with my local bonfire config.  I'm not 100% sure why.

jlsherrill:
So i got it working, about half the time :/   the other half the time, there is some timing issue and the snapshot task is not scheduled.   Pushing up some debugging

swadeley:
> @swadeley we've added a command to run when deploying to ephemeral that should snapshot a small rhel repo (Red Hat Ansible Engine 2 for RHEL 8 x86_64 (RPMs)) Hopefully there's enough memory for it to snapshot. Note that it doesn't 'block' deployment on the snapshot, it happens in the background.
> 
> I'll spin it up to verify as well once the image is built

Hi, I have had this PR and the frontend [1] deployed for over 24 hours. I saw the Ansible repo was snapshotted right away, but still no RHEL repos due to "Killed by signal 9." as mentioned before [2].

Let me know when I should redeploy to test a new image.

Thank you

[1] https://github.com/content-services/content-sources-frontend/pull/156
[2] https://github.com/content-services/content-sources-frontend/pull/156#issuecomment-1797120004

jlsherrill:
@swadeley so this should be good to merge if you are okay with it.  I have removed my debugging output

pkg/dao/repository_configs.go
Andrewgdewar: I was able to fix this issue, I think we largely just weren't hitting it before, and it was pre-existing.

Andrewgdewar:
> It says in your summary that create/update/delete do not have an effect for red hat repositories. Right now, if my org ID is -1, I can delete/update but not create. Obviously, that's not a real use case, but I feel the behavior should probably be consistent unless there's a reason for it not to be.
> 
> Do we expect to want to delete/update red hat repositories from the API? Or should that be blocked the way create is blocked?

If you can highlight somewhere that the dao layer would allow this, please do. 
I think I covered all cases of CrUD not being possible.

rverdile (approved):
nice!!

PR #438, opened 2023-10-23
Fixes 2896: refactor pulp client in dao

rverdile:
## Summary
The method I introduced [here](https://github.com/content-services/content-sources-backend/pull/414/files#diff-c90c06b1d5bb21209cf1d0b2bd26ae86ec3db770764e6653d5dc72580a10c16fR78-R89) to set the domain scoped pulp client is bug-prone.

This PR demonstrates a new way of setting the pulp client by adding two chained methods. Now, we can initialize the pulp client with `context.Background()` during dependency injection. Then, replace the context later, during the request, so that the pulp client is using the domain.

For example, the pulp client method `GetContentPath()` is called with `pulpClient.WithContext(ctx).WithDomain(domainName).GetContentPath()`

I've commented out most of the failing tests for now, as I was just trying out this approach.


pkg/dao/snapshots_test.go
```diff
@@ -86,9 +87,14 @@ func (s *SnapshotsSuite) TestCreateAndList() {
 
 	mockPulpClient := pulp_client.NewMockPulpClient(t)
 	sDao := snapshotDaoImpl{db: tx, pulpClient: mockPulpClient}
-	mockPulpClient.On("GetContentPath").Return(testContentPath, nil)
+	sDao.WithContext(context.Background())
 
-	repoDao := repositoryConfigDaoImpl{db: tx, yumRepo: &mockExt.YumRepositoryMock{}}
+	mockPulpClient.On("WithContext", context.Background()).Return(mockPulpClient).Twice()
+	mockPulpClient.On("GetContentPath").Return(testContentPath, nil).Twice()
+	mockPulpClient.On("WithDomain", "").Return(mockPulpClient)
```
rverdile, line 94: I think this PR is a step in the right direction, although it feels like it adds a lot more boiler plate (like this) to the tests which is not ideal.

jlsherrill:
https://issues.redhat.com/browse/HMS-2896

pkg/dao/snapshots_test.go
jlsherrill: here's an idea (if you think it makes sense) .   We could add some methods to the mock that do this for us.  I imagine something like:

```
mockPulpClient.WithContext().WithDomain().
     On("GetContentPath").Return(testContentPath, nil).Twice()

```


and under the hood, this would mock WithContext & WithDomain  MinTimes(1)   so we'd lose the count on those particular mock calls, but I'm not sure we actually care for that to be precise.  I would imagine these methods would have to live in some new file like "pulp_client_mock_helpers.go"  or something?

jlsherrill: actually i realized MinTimes() is in go mock and not testify.  It looks like if you just don't specify Times()  it doesn't care the number

jlsherrill:
this needs a rebase now too

pkg/dao/snapshots_test.go
rverdile: I like this idea. I'll make that change.

pkg/dao/repository_configs.go
```diff
@@ -75,20 +69,9 @@ func DBErrorToApi(e error) *ce.DaoError {
...
-	if err != nil {
-		return err
-	}
-
-	pulpClient := pulp_client.GetPulpClientWithDomain(context.TODO(), domainName)
-	r.pulpClient = pulpClient
-	return nil
+func (r *repositoryConfigDaoImpl) WithContext(ctx context.Context) RepositoryConfigDao {
+	r.ctx = ctx
+	return r
```
jlsherrill, line 74: Should this return a copy of the DoaImpl instead of the same DaoImpl? 

I'm imagining two requests coming in at the same time, being run at the same time in the same process.  They are sharing a DaoImpl, so if they each call .WithContext, they could stomp over each other?  

Its very possible i'm not understanding something corretly thougH :)

rverdile: I think you're right. I'll see if I can get that to return a copy such that mocking still works in the tests.

rverdile:
this is updated now

jlsherrill (approved):
ACK pending tests passing

PR #439, opened 2023-10-23
Build: use ubi-micro

jlsherrill:
## Summary

Switches to using the ubi-micro container, which appears to have everything we need other than a ca-trust, which we can copy from ubi-minimal

This reduces our security footprint and potential attack vectors

## Testing steps

tests pass

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -13,9 +13,14 @@ USER 0
 
 RUN make get-deps build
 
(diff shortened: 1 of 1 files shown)


rverdile (approved):
tests pass!

mshriver (approved):
@jlsherrill leaving this for you to merge

PR #440, opened 2023-10-24
Fixes 2890: don't query pulp if not enabled

rverdile:
## Summary
Seeing error where if pulp isn't enabled, certain API endpoints are broken because there are dao functions that query pulp. This checks if snapshots are enabled in all of those places before querying pulp.

## Testing steps
Turn off pulp and see if the errors are still present when listing/fetching repositories.

### DIFF
--- a/pkg/dao/repository_configs.go
+++ b/pkg/dao/repository_configs.go
@@ -76,6 +76,10 @@ func DBErrorToApi(e error) *ce.DaoError {
 }
 
 func (r *repositoryConfigDaoImpl) InitializePulpClient(ctx context.Context, orgID string) error {
+	if !config.Get().Features.Snapshots.Enabled {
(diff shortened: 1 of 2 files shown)


jlsherrill:
code looks good, i will test

jlsherrill (approved):

jlsherrill:
You should be able to test this in ephemeral with:

--set-parameter content-sources-backend/FEATURES_SNAPSHOTS_ENABLED=false   --set-parameter content-sources-backend/CLIENTS_PULP_SERVER=""

swadeley:
Hi, as per Ryans suggestion I had to use the bonfire conf to get `CLIENTS_PULP_SERVER=""` setting to work.

All API tests pass.

Thank you

PR #444, opened 2023-10-25
Fixes 2450: remove org admin skip setting

rverdile:
## Summary
Removes the setting that enables/disables "org admin rbac skip", but keeps the functionality i.e. it is permanently enabled. This allows a user with the org admin field set to true to skip the rbac check.

## Testing steps
1. In your `config.yaml` set `clients: rbac_enabled` to true.
2. When your start the server, start it with the `mock_rbac` argument. i.e. `go run cmd/content-sources/main.go api consumer instrumentation mock_rbac`
1. GET repositories with a user that does not have read permissions and the org admin field is _false_. You should get a 401. For example `./scripts/header.sh <org_id> usernoperms` will generate the identity.
3. GET repositories with a user that does not have read permissions and the org admin field is _true_. You should get a 200 response. For example `./scripts/header_org_admin.sh <org_id> usernoperms` will generate the identity.

## Checklist

### DIFF
--- a/deployments/deployment.yaml
+++ b/deployments/deployment.yaml
@@ -215,8 +215,6 @@ objects:
                     name: cs-pulp-admin-password
                     key: password
                     optional: true
-              - name: RBAC_ORG_ADMIN_SKIP
-                value: ${RBAC_ORG_ADMIN_SKIP}
               - name: LOGGING_LEVEL
                 value: ${{LOGGING_LEVEL}}
               - name: CLIENTS_RBAC_BASE_URL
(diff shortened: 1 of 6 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-2450

xbhouse (approved):
LGTM! tested and works as expected. someone else should probably double check the code though :)

jlsherrill (approved):
ACK from a code perspective, and don't forget to remove the RBAC_ORG_ADMIN_SKIP setttings from our app-interface deploy.yaml

PR #445, opened 2023-10-25
Fixes 2282: reduce number of introspect tasks

jlsherrill:
## Summary

This restructures the introspection code to move all decisions about whether a repo should be introspected into the main query to lookup repos for introspection.  the result is that we do not spawn tasks for repos that won't actually be introspected.  It should reduce the tasks we spawn by about 2/3rds.  

Some considerations:
* Now in pkg/external_repos  any introspect function will introspect without question
* The notifications are being sent one at a time for introspection, but they were before as well.  We likely want to revisit how those are generated.

## Testing steps

1.  Create one or more repos for introspection (but turn off snapshotting for simplicity)
2. run ```go run cmd/external-repos/main.go nightly-jobs```
3. notice that no tasks get run
4.  wait ~24 hours or adjust config.IntrospectTimeInterval in value_constraints.go to `time.Minute * 1`  and wait ~ 1 minute
5. run ```go run cmd/external-repos/main.go nightly-jobs```
6. notice taht tasks now get spawned 

## Checklist


cmd/external-repos/main.go
```diff
@@ -64,15 +60,16 @@ func main() {
 				forceIntrospect = true
 			}
```
jlsherrill, line 61: right now force isn't doing anything, i will fix that.

jlsherrill:
https://issues.redhat.com/browse/HMS-2282

pkg/external_repos/introspect.go
```diff
@@ -107,76 +129,6 @@ func Introspect(ctx context.Context, repo *dao.Repository, dao *dao.DaoRegistry)
...
-		if err != nil {
-			errors = append(errors, err)
-		}
-	}
-
-	// Logic to handle notifications
-	sendIntrospectionNotifications(introspectSuccessUuids, introspectFailedUuids, dao)
-
-	return total, introspectionErrors, errors
-}
```
jlsherrill, line 178: There isn't really any use for a 'All' function here.   Its always done on a per repo basis via  task, so i removed it.

xbhouse:
tested and getting expected results. 

no tasks spawned when running that command (`go run cmd/external-repos/main.go nightly-jobs`) right after adding a repo. changed the value_constraints and waited a minute to run it again and see that tasks now get spawned

Screenshot 2023-11-07 at 9 55 21 AM

xbhouse:
i'm going to ack but maybe someone else will have more feedback / see something i didn't

xbhouse (approved):

swadeley:
Hi, lgtm, waiting for one more ack.

rverdile (commented):
If introspection fails, e.g. URL http://example.com, I'm not seeing the task marked as failed. It still says "task completed". There's also no error saved on the task.

pkg/external_repos/introspect.go
```diff
@@ -57,10 +83,6 @@ func Introspect(ctx context.Context, repo *dao.Repository, dao *dao.DaoRegistry)
 
 	logger.Debug().Msg("Introspecting " + repo.URL)
 
-	if repo.FailedIntrospectionsCount >= config.FailedIntrospectionsLimit && !repo.Public {
-		return 0, fmt.Errorf("introspection skipped because this repository has failed more than %v times in a row", config.FailedIntrospectionsLimit), false
-	}
-
```
rverdile, line 63: The feature to block introspection after a certain number of failed attempts no longer works. I'm not sure if this was meant to be removed then placed somewhere else.

rverdile: Tested by calling `repositories/:uuid/introspect/` on a failing repo 20+ times.

rverdile (commented):
I'm also still seeing errors for public repos being logged at the warn level. The issue says to log them at the error level.

Tested by either running `make repos-import` then `go run cmd/external-repos/main.go nightly-jobs` (without a cdn cert so that red hat ones fail). Or, manually updating the database to set a failing repository to public.

pkg/external_repos/introspect.go
jlsherrill: ah, it was meant to be covered as part of the query, i'll check i may have missed that

jlsherrill: oh i see, the nightly introspection would honor this, but the api isn't.  Will fix!

pkg/dao/repositories.go
```diff
@@ -72,17 +72,27 @@ func (p repositoryDaoImpl) FetchForUrl(url string) (Repository, error) {
...
-		result = p.db.Find(&dbRepos)
+	db := p.db
+	if !force && !config.Get().Options.AlwaysRunCronTasks {
+		introspectThreshold := time.Now().Add(config.IntrospectTimeInterval * -1) // Add a negative duration
+		db = db.Where(
+			db.Where("status != ?", config.StatusValid).
+				Or("last_introspection_time is NULL").                   // It was never introspected
+				Or("last_introspection_time < ?", introspectThreshold)). // It was introspected more than the threshold ago)
+			Where(
+				db.Where("failed_introspections_count < ?", config.FailedIntrospectionsLimit).
```
rverdile, line 88: I think this should be `config.FailedIntrospectionsLimit+1`, as it was before. I did that so the error on the repository would be "failed because it's over the limit", before introspection started skipping. This way the repository being skipped is communicated in the UI.

rverdile: Otherwise I think this would just silently start skipping introspection for repository over the limit

swadeley:
@jlsherrill rebase please

rverdile (approved):
working!

swadeley:
lgtm, thank you.

PR #467, opened 2023-11-13
Fixes 2768: add canceling state to PollTask

rverdile:
## Summary
Adds "canceling" as one of the possible states while polling a pulp task. Previously it was jumping to the default case of "Pulp task with unexpected state" and logging an error.

If you search for "canceling" [here in the pulp docs](https://docs.pulpproject.org/pulpcore/restapi.html#tag/Tasks/operation/tasks_list), you can see it is a valid state.

## Testing steps
I can't find a reliable way to reproduce this, but it's a very small and simple change.

You can try

1. Creating a larger repo like epel 9, with snapshot enabled.
2. Canceling that task using the cancel endpoint (`/tasks/:uuid/cancel/`), and the `last_snapshot_task_uuid` from the create response.
3. You may or may not see the new state logged.

## Checklist

### DIFF
--- a/pkg/pulp_client/task.go
+++ b/pkg/pulp_client/task.go
@@ -19,6 +19,7 @@ const (
 	RUNNING   string = "running"
 	SKIPPED   string = "skipped"
 	CANCELED  string = "canceled"
+	CANCELING string = "canceling"
 	FAILED    string = "failed"
 )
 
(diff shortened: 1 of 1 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-2768

jlsherrill (approved):

PR #477, opened 2023-11-20
Fixes 3081: always intropsect public repos

jlsherrill:
## Summary
even if failed counts has been reached

As public repos should never fail  in the real world, some action may need to be taken, and we can retry.

We add RHEL repos before they are actually on the cdn, which means it can fail introspection for some time before its officially there.  Once it is, we want to introspect it properly.

## Testing steps

I wrote a test to test this, i think test passing is sufficient. 

## Checklist


Andrewgdewar (approved):
LGTM!

jlsherrill:
since iqe tests passed, and the only test failure is some ci issue, i'm merging

PR #483, opened 2023-11-21
Refs 3050: delete all snapshots a second time

jlsherrill:
## Summary
Delete all snapshots a 2nd time

## Testing steps
none
## Checklist


jlsherrill:
actually the snapshots i saw were like the iqe tests that run after deployment

PR #497, opened 2023-12-06
Fixes 3250: provide filter for excluding loaded RH repos

jlsherrill:
## Summary

Provide an env variable that can be used to exclude some or all of the red hat repos from being loaded.  

## Testing steps

In dev,
In between each test, a ```make compose-clean compose-up``` is needed to clear out what is there.
With backend and frontend running:
```
go run ./cmd/external-repos/main.go import
```

should still import all red hat repos (can check in the UI).  
```
OPTIONS_REPOSITORY_IMPORT_FILTER="small" go run ./cmd/external-repos/main.go import
```

should import only 1 small ansible redhat repo

```
OPTIONS_REPOSITORY_IMPORT_FILTER="none" go run ./cmd/external-repos/main.go import
```
should cause no red hat repos to be imported.

For QE, you can use with bonfire: 

--set-parameter content-sources-backend/OPTIONS_REPOSITORY_IMPORT_FILTER=small
--set-parameter content-sources-backend/OPTIONS_REPOSITORY_IMPORT_FILTER=none

## Checklist

### DIFF
--- a/deployments/deployment.yaml
+++ b/deployments/deployment.yaml
@@ -123,6 +123,8 @@ objects:
                 value: ${OPTIONS_ALWAYS_RUN_CRON_TASKS}
               - name: OPTIONS_ENABLE_NOTIFICATIONS
                 value: ${OPTIONS_ENABLE_NOTIFICATIONS}
+              - name: OPTIONS_REPOSITORY_IMPORT_FILTER
+                value: ${OPTIONS_REPOSITORY_IMPORT_FILTER}
             resources:
               limits:
                 cpu: ${CPU_LIMIT}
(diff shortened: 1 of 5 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-3250

deployments/deployment.yaml
```diff
@@ -149,24 +151,6 @@ objects:
...
-                args:
-                  - /external-repos
-                  - import
-              - name: introspect-single-repo
-                inheritEnv: true
-                args:
-                  - /external-repos
-                  - introspect
-                  - https://cdn.redhat.com/content/dist/layered/rhel8/x86_64/ansible/2/os
-                  - https://cdn.redhat.com/content/dist/rhel8/8.7/x86_64/baseos/os
```
jlsherrill, line 169: these init containers seemed duplicated? I don't think its needed, so i removed from here, but its still above.

jlsherrill:
also, if you do test with ephemeral, you'll need the deployment.yaml changes here

swadeley:
Hi, I confirm:

`--set-parameter content-sources-backend/OPTIONS_REPOSITORY_IMPORT_FILTER=small` results in just the ansible repo being deployed.

PR #499, opened 2023-12-07
Fixes 3016: Reject requests with org id of -1

xbhouse:
## Summary

Adds check to middleware to reject requests if the Org ID is -1

## Testing steps

Make a request to any endpoint with an identity header with Org ID set to -1. Response should return a 403. 

## Checklist

### DIFF
--- a/pkg/middleware/enforce_identity.go
+++ b/pkg/middleware/enforce_identity.go
@@ -70,7 +70,7 @@ func EnforceOrgId(next echo.HandlerFunc) echo.HandlerFunc {
 	return func(c echo.Context) error {
 		xRHID := identity.Get(c.Request().Context())
 
-		if xRHID.Identity.Internal.OrgID == "-1" || xRHID.Identity.OrgID == "-1" {
+		if xRHID.Identity.Internal.OrgID == config.RedHatOrg || xRHID.Identity.OrgID == config.RedHatOrg {
 			err := ce.NewErrorResponse(http.StatusForbidden, "Invalid org ID", "Org ID cannot be -1")
 			c.Error(err)
 			return nil
--- a/pkg/middleware/enforce_identity.go
+++ b/pkg/middleware/enforce_identity.go
@@ -1,15 +1,26 @@
 package middleware
 
 import (
+	"encoding/base64"
+	"encoding/json"
 	"net/http"
 	"strings"
 
 	"github.com/content-services/content-sources-backend/pkg/api"
 	"github.com/content-services/content-sources-backend/pkg/config"
+	ce "github.com/content-services/content-sources-backend/pkg/errors"
 	"github.com/labstack/echo/v4"
 	echo_middleware "github.com/labstack/echo/v4/middleware"
 	"github.com/redhatinsights/platform-go-middlewares/identity"
 )
+
+type XRHID struct {
+	Type          string `json:"type"`
+	AccountNumber string `json:"account_number"`
+	Internal      struct {
+		OrgID string `json:"org_id"`
+	} `json:"internal"`
+}
 
 // WrapMiddleware wraps `func(http.Handler) http.Handler` into `echo.MiddlewareFunc`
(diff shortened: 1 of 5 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-3016

pkg/router/route.go
```diff
@@ -34,6 +34,7 @@ func ConfigureEcho(allRoutes bool) *echo.Echo {
 		Skipper:         config.SkipLogging,
 	}))
 	e.Use(middleware.EnforceJSONContentType)
+	e.Use(middleware.EnforceOrgId)
```
jlsherrill, line 37: I like the use of middleware to solve this problem!   looking at the implementation, i noticed you are parsing the identity header manually, but this shouldn't actually be needed, as something like this *should* work, which pulls it from the context:

```		xRHID := identity.Get(c.Request().Context())```

However it didn't.  I poked around, and this middleware call should add it to the context:
```
	e.Use(middleware.WrapMiddlewareWithSkipper(identity.EnforceIdentity, middleware.SkipAuth))
```
I think if you move your `e.Use(middleware.EnforceOrgId)` after this line it should work and be much simpler.

But to be honest, i think your line here is in the right place but `	e.Use(middleware.WrapMiddlewareWithSkipper(identity.EnforceIdentity, middleware.SkipAuth))
` is in the wrong place.  So rather, i'd move that up above your line.

Comment (xbhouse) [reply pkg/router/route.go:37]: ok awesome, that does simplify it quite a bit. thank you! so should i remove `e.Use(middleware.WrapMiddlewareWithSkipper(identity.EnforceIdentity, middleware.SkipAuth))` from `ConfigureEchoWithMetrics` and just have it in the `ConfigureEcho` function? 

i also am seeing the response returns a `status=200` when switching the order of middlewares for some reason, like this:

`ERR ../../../go/pkg/mod/github.com/ziflex/lecho/v3@v3.5.0/middleware.go:165 > error="code=403, message=Forbidden" bytes_in=0 bytes_out=0 host=localhost:8000 latency=0.296803 latency_human="296.803µs" method=GET referer= remote_ip=::1 request_id=XVOakqkgRgpJBCBPAUerXiilEjMFJWIF status=200 uri=/api/content-sources/v1.0/repositories/ user_agent=PostmanRuntime/7.35.0`

adding `c.Error(err)` to the EnforceOrgId middleware function forces it to return a 403 but i'm not sure if this is the best way to handle that?

File: pkg/middleware/enforce_identity.go [hunk 2]
```diff
@@ -61,3 +65,17 @@ func SkipAuth(c echo.Context) bool {
...
 }
+
+func EnforceOrgId(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		xRHID := identity.Get(c.Request().Context())
+
+		if xRHID.Identity.Internal.OrgID == "-1" || xRHID.Identity.OrgID == "-1" {
+			err := ce.NewErrorResponse(http.StatusForbidden, "Invalid org ID", "Org ID cannot be -1")
+			c.Error(err)
+			return nil
```
jlsherrill, line 76: ah! so you call c.Error() and then return nil.  Would you mind fixing this middleware too? https://github.com/content-services/content-sources-backend/blob/main/pkg/middleware/enforce_json.go#L27

pkg/middleware/enforce_identity.go
```diff
@@ -61,3 +65,17 @@ func SkipAuth(c echo.Context) bool {
 
 	return false
 }
+
+func EnforceOrgId(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		xRHID := identity.Get(c.Request().Context())
+
+		if xRHID.Identity.Internal.OrgID == "-1" || xRHID.Identity.OrgID == "-1" {
```
jlsherrill, line 73: You can refer to `config.RedHatOrg` to avoid hardcoding -1

jlsherrill (approved):
ACK!

PR #505, opened 2023-12-14
Fixes 3232: add gh action to prevent out of order migrations

xbhouse:
## Summary

Added Github Action to check for out of order database migrations.

## Testing steps

If the content in `db/migrations.latest` does not reflect the timestamp of the newest added set of migration files in `db/migrations`, the job will fail. If these timestamps match or if there are no new migrations, the job will succeed. 

Can test by:
- Adding a set of test migration files with an earlier timestamp than the latest and not updating the `db/migrations.latest`. This should cause the job to fail.
- Removing the test migration files. The `db/migrations.latest` content should reflect the newest added set of migration files. This should cause the job to succeed.
- Adding a set of test migration files with a timestamp that matches the timestamp in `db/migrations.latest`. This should cause the job to succeed.

## Checklist

### DIFF
--- a/.github/workflows/content-sources-actions.yml
+++ b/.github/workflows/content-sources-actions.yml
@@ -48,6 +48,37 @@ jobs:
           version: v1.54.2
           skip-go-installation: true
           args: --timeout=5m
+  
+  checkmigrations:
+    name: Check db migrations
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-3232

jlsherrill (approved):
ACK, nice approach!

PR #519, opened 2024-01-04
Fixes 3171: add gpg key path for red hat repos

rverdile:
## Summary
The repository config file wasn't providing the correct gpg key path for red hat repos. All the red hat repos have the same gpg key path, so this hard codes that value. 

## Testing steps
1. Import red hat repos and create a snapshot
2. Fetch the config.repo file from the `/repositories/:uuid/snapshots/:snap_uuid/config.repo/` endpoint
3. The config.repo file should have the red hat gpg key path

## Checklist

### DIFF
--- a/pkg/dao/repository_configs_test.go
+++ b/pkg/dao/repository_configs_test.go
@@ -1554,6 +1554,8 @@ func (suite *RepositoryConfigSuite) TestBulkDeleteRedhatRepository() {
 	orgID := config.RedHatOrg
 	repoConfigCount := 5
 
+	suite.tx.Exec("TRUNCATE repositories, snapshots, repositories_rpms, repositories_package_groups, repository_configurations")
+
 	err := seeds.SeedRepositoryConfigurations(suite.tx, repoConfigCount, seeds.SeedOptions{OrgID: orgID})
 	assert.Nil(t, err)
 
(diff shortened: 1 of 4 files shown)


pkg/dao/repository_configs_test.go
```diff
@@ -1554,6 +1554,8 @@ func (suite *RepositoryConfigSuite) TestBulkDeleteRedhatRepository() {
 	orgID := config.RedHatOrg
 	repoConfigCount := 5
 
+	suite.tx.Exec("TRUNCATE repositories, snapshots, repositories_rpms, repositories_package_groups, repository_configurations")
```
jlsherrill, line 1557: any idea why this is needed?

jlsherrill: i guess i should rephrase it, can we fix the test so this isn't needed?   I get its in a transaction and gets rolled back, so maybe its not a big deal, but it feels really bad

rverdile: that's fair. I can change that

jlsherrill:
https://issues.redhat.com/browse/HMS-3171

pkg/dao/repository_configs_test.go
rverdile: fixed

jlsherrill (approved):

PR #526, opened 2024-01-10
Fixes 3170: use x-forward-host hdr for repo config

jlsherrill:
## Summary

To properly show the gpg key url from the gateway

## Testing steps

1.  create a repo with a gpg key set, and snapshotting enabled
2.  fetch your repos list to get uuids for the repo and latest snapshots: GET http://localhost:8000/api/content-sources/v1/repositories/
3.    fetch this url: http://localhost:8000/api/content-sources/v1/repositories/60dd106b-d08e-436a-bf34-53180713d4bc/snapshots/8a68fc2d-c8eb-4046-8c56-df44603f8206/config.repo

replacing the UUIDs for the repo & snapshot UUID. Notice the URL is likely https://localhost:8000/.

4.  now fetch the url with the header:   'X-Forwarded-Host: example.com'
you should see the url change to  https://example.com/

## Checklist

### DIFF
--- a/pkg/handler/snapshots.go
+++ b/pkg/handler/snapshots.go
@@ -126,7 +126,12 @@ func (sh *SnapshotHandler) getRepoConfigurationFile(c echo.Context) error {
 	uuid := c.Param("uuid")
 	snapshotUUID := c.Param("snapshot_uuid")
 
-	repoConfigFile, err := sh.DaoRegistry.Snapshot.WithContext(c.Request().Context()).GetRepositoryConfigurationFile(orgID, snapshotUUID, uuid, c.Request().Host)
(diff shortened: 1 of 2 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-3170

xbhouse:
tested and works well! only thing i noticed is i think there is a typo in the test, should be setting the `x-forwarded-host` header?

jlsherrill:
good catch! i think the default of 'example.com' was being picked up, changed that to see the failure and then fixed it

xbhouse (approved):

swadeley:
All API tests pass.

PR #527, opened 2024-01-11
Refs 3372: Drop all snapshots in stage

jlsherrill:
## Summary

Deletes all snapshots in stage, so we can migrate to a new pulp server.  Will be merged after snapshotting is disabled in stage.  The repos & versions will remain in the pulp server, but we will eventually just drop the pulp server.  The rpms will remain in the s3 bucket, but we will switch to a new bucket and delete the old one.  

## Testing steps

Create a repo with snapshotting, let it snapshot, see a snapshot is created. 
run: `go run cmd/cleanup_versions/main.go  --force`
snapshot should no longer exist.

## Checklist


jlsherrill:
DO NOT MERGE until snapshotting is disabled in stage

jlsherrill:
https://issues.redhat.com/browse/HMS-3372

Andrewgdewar (approved):
Indeed! Deletes things

PR #545, opened 2024-01-23
Build: Update to use new pulp api path

jlsherrill:
## Summary

Updates our compose files to configure pulp to use /pulp/api

## Testing steps

1.  checkout PR
2. make compose-clean compose-up
3. make run
4. create a repo with snapshotting turned on, verify the snapshot gets created properly.

## Checklist


rverdile (approved):
tested and works!

PR #548, opened 2024-01-24
Fixes 2017: add retry for failed tasks

rverdile:
## Summary
Adds a periodic requeue for delete-repository-snapshot tasks that have failed. Limited up to 3 retries.
Adds backoff timer for retries, with config option (tasking.retry_wait_upper_bound) for maximum wait time (wait time for final retry). 

## Testing steps
1. Add the `retry_wait_upper_bound` field to your config, with a low value like `1s`.
2. Create a snapshot
3. Get the delete-repository-snapshot task to fail. One way to do that is to do `podman stop cs_pulp_api_1`, so requests to the pulp server get a 502.
4. Observe that the task will get requeued and dequeued 3 times before permanently finishing in a failed state.

## Checklist


rverdile:
It occurs to me that this may not be useful without some kind of delay between the requeues, so I'm gonna put this in draft and keep thinking about that.

jlsherrill:
https://issues.redhat.com/browse/HMS-2017

jlsherrill:
I could see only requeing tasks that failed more than ~4 hours ago or something like that?   these tasks aren't really meant to be user visible, so its okay if they take a while.

rverdile:
> I could see only requeing tasks that failed more than ~4 hours ago or something like that? these tasks aren't really meant to be user visible, so its okay if they take a while.

Yeah, maybe it's more complicated than it needs to be, but I've been trying to do something with exponential backoff.

jlsherrill:
oh yeah, that does sound complicated!  But if you're close that sounds promising.  If not, i think simple is fine to start with.  Either way!

pkg/tasks/queue/pgqueue.go
```diff
@@ -541,6 +547,15 @@ func (p *PgQueue) Finish(taskId uuid.UUID, taskError error) error {
 		return ErrNotRunning
 	}
 
+	var nextRetryTime *time.Time
+	if status == config.TaskStatusFailed && info.Retries <= MaxTaskRetries {
+		upperBound := config.Get().Tasking.RetryWaitUpperBound
+		retriesRemaining := float64(MaxTaskRetries - info.Retries)
+		timeToWait := time.Minute * time.Duration(upperBound.Minutes()/(retriesRemaining+1))
+		add := time.Now().Add(timeToWait)
+		nextRetryTime = &add
```
jlsherrill, line 556: i set an upper bound of '10s', but it still seems to retry every 1 minute (up to 3 times), is that expected?

jlsherrill:
It might be worth adding an explicit Error Log message when a task failed and is out of retries.  I could see this be worth investigating.  Thoughts?

pkg/tasks/queue/pgqueue.go
rverdile: It's not really expected, but I wrote this expecting to use hours (or 0 for tests). Let me see if I can adjust this to be a little more flexible.

jlsherrill: I'm seeing a little something funky.  When i set the RetryWaitUpperBound to 10s, it sets the nextRetryTime to essentially time.Now(), which i assume means that add is basically zero.  (I confirmed the nextRetryTime by looking in the db).  But the task doesn't actually execute until about 30 seconds later.

I set it to 1m, and the nextRetryTime did get set in the future, but the task didn't actually execute until 20 seconds or so after that time passed.  I this expected?

rverdile: What is your heartbeat timer set to? The tasks will get requeued once per heartbeat. Could that be why you're seeing a delay?

jlsherrill: ahhh!, that make sense.  Does the other  behavior sound fine?  as you said i don't think we'd actually use these low value, so i think its okay to me?

rverdile: I guess it is a bit confusing. But since these low values are only ever used for testing purposes, I think it's okay. It's simpler this way.

PR #560, opened 2024-02-02
Fixes 3424: add label to repository configs

jlsherrill:
## Summary

Adds a label column to the repository configurations table, and returns it from the repositories api.  
For custom repos this will currently be blank (in the future we will likely set this).
For red hat repos, this will be populated
For the snapshot config.repo api, this label is used (if present). 

## Testing steps

* check out main and run 'make repos-import'.  GET /repositories?origin=external,red_hat and see that there is no label attribute
* checkout this pr, db migrate, run `make repos-import` again
* refetch the repositories via the same api and confirm that there is now a label.
* Create a custom repo and see that its label is blank
* trigger a snapshot of a redhat repo (go run cmd/external-repos/main.go snapshot $URL)  
* fetch the config.repo file and confirm that the correct label is there:   GET /repositories/UUID/snapshots/UUID/config.repo

## Checklist


jlsherrill:
https://issues.redhat.com/browse/HMS-3424

swadeley:
Hi

I get  introspection failure  missing comps file in repo[1]. Tool tip shows:

`error getting comps: error parsing comps.xml: invalid file type: must be gzip, xz, or zstd`

I do not see any label in the Red Hat Ansible Engine repo config. Should I?

[1] 
https://github.com/content-services/content-sources-backend/pull/560/files

swadeley:
Hi, I raised Jira for the comps file detection problem:
 https://issues.redhat.com/browse/HMS-3539

jlsherrill:
the label is in the square brackets:

for example, needed_errata here:

```
[needed_errata]
name=needed errata
baseurl=http://pulp_content:24816/pulp/content/7f584fe5/9d7cac21-c81c-4b98-9adf-0539ad2a9a68/ac7891a4-0560-4af2-b1e5-e7586fe89e03/
module_hotfixes=0
gpgcheck=0
repo_gpgcheck=0
enabled=1
gpgkey=
```

db/migrations/20240202124553_AddRepoConfigLabel.up.sql
```diff
@@ -0,0 +1,8 @@
+BEGIN;
+
+
+
+ALTER TABLE repository_configurations
+    ADD COLUMN IF NOT EXISTS label VARCHAR (255)   DEFAULT FALSE ;
+
+COMMIT;
```
rverdile, line 8: is default  "false" the intention here or should this be null/blank? Also, if I create a new custom repository, the default is an empty string.

jlsherrill: good catch, this should be empty string.

jlsherrill: i'll also set it not null

rverdile (approved):
lgtm!

PR #566, opened 2024-02-06
Fixes 3420: add service notify for templates

rverdile:
## Summary
Sends an event on a new kafka topic when a template is created, updated, or deleted. 

This has nothing to do with sending an event (notification) to the notifications service or getting emails. I've used the term "notification" in places because it is still technically a notification (we're notifying the patch service), I wasn't sure what else to call it, and there's an opportunity here to share code between the templates and notifications. I'm open to other names!

## Testing steps
1. In a terminal, `make kafka-topic-consume KAFKA_TOPIC=platform.content-sources.template`
2. Start the API server
3. Create a template. Check the terminal where you've just run the kafka consumer, you should see a notification.
4. Update a template. Check for a notification.
5. Delete a template. Check for a notification.
## Checklist

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -22,20 +22,21 @@ import (
 const DefaultAppName = "content-sources"
 
 type Configuration struct {
-	Database            Database
-	Logging             Logging
-	Loaded              bool
-	Certs               Certs
-	Options             Options
-	Kafka               event.KafkaConfig
-	Cloudwatch          Cloudwatch
-	Metrics             Metrics
-	Clients             Clients            `mapstructure:"clients"`
-	Mocks               Mocks              `mapstructure:"mocks"`
-	Sentry              Sentry             `mapstructure:"sentry"`
-	NotificationsClient cloudevents.Client `mapstructure:"notification_client"`
-	Tasking             Tasking            `mapstructure:"tasking"`
-	Features            FeatureSet         `mapstructure:"features"`
+	Database                     Database
+	Logging                      Logging
+	Loaded                       bool
+	Certs                        Certs
+	Options                      Options
+	Kafka                        event.KafkaConfig
+	Cloudwatch                   Cloudwatch
+	Metrics                      Metrics
+	Clients                      Clients            `mapstructure:"clients"`
+	Mocks                        Mocks              `mapstructure:"mocks"`
+	Sentry                       Sentry             `mapstructure:"sentry"`
+	NotificationsClient          cloudevents.Client `mapstructure:"notification_client"`
+	TemplatesNotificationsClient cloudevents.Client `mapstructure:"templates_notification_client"`
+	Tasking                      Tasking            `mapstructure:"tasking"`
+	Features                     FeatureSet         `mapstructure:"features"`
 }
 
 type Clients struct {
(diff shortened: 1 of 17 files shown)


pkg/notifications/client.go
```diff
@@ -77,3 +77,34 @@ func SetEmptyToNil(value string) *string {
...
+	if config.Get().TemplatesNotificationsClient != nil && len(templates) > 0 {
+		eventNameStr := eventName.String()
+		newUUID, _ := uuid.NewRandom()
+		e := cloudevents.NewEvent()
+		e.SetSource("urn:redhat:source:console:app:repositories")
+		e.SetID(newUUID.String())
+		e.SetType("com.redhat.console.repositories." + eventNameStr)
+		e.SetSubject("urn:redhat:subject:console:rhel:" + eventNameStr)
+		e.SetTime(time.Now())
+		e.SetExtension("redhatorgid", orgID)
```
rverdile, line 92: These fields are set for the repository notifications. Not sure it's super important here, but I've copied them over as they seem like they might contain useful information.

jlsherrill:
https://issues.redhat.com/browse/HMS-3420

pkg/config/config.go
```diff
@@ -219,6 +221,7 @@ func setDefaults(v *viper.Viper) {
 	v.SetDefault("options.introspect_api_time_limit_sec", DefaultIntrospectApiTimeLimitSec)
 	v.SetDefault("options.always_run_cron_tasks", false)
 	v.SetDefault("options.enable_notifications", false)
+	v.SetDefault("options.enable_templates_notifications", false)
```
jlsherrill, line 224: we should probably default this to true.  We set notifications to false as we didn't want it turned on in production.   In fact, i might would argue we make the topic configurable (defaulting to 'platform.content-sources.template') and send it if its set.    That way, you're implicitly turning it on by setting the topic to deliver it on.

pkg/config/event.go
```diff
@@ -74,30 +75,81 @@ func readEnv(key string, def string) string {
...
+
+	c, err := v2.NewClient(protocol, v2.WithTimeNow(), v2.WithUUIDs())
+	if err != nil {
+		return nil, fmt.Errorf("failed to create cloud events client: %w", err)
+	}
+	return c, nil
+}
+
+// SetupTemplatesNotifications creates the cloud events kafka client for sending events to the patch service
+func SetupTemplatesNotifications() {
```
jlsherrill, line 106: I think i'd perfer if we referred to these as events or messages internally.  Notifications is a console dot thing that is implemented via kafka events/messaging, but also had a rest implementation as well.

jlsherrill: (and to be clear, i think its fine to call kafka messages sent to platform.notifications.ingress as notifications, but calling it all generically 'notifications' might be confusing)

pkg/config/event.go
```diff
@@ -74,30 +75,81 @@ func readEnv(key string, def string) string {
...
+
+func GetSaramaConfig() (*sarama.Config, error) {
 	saramaConfig := sarama.NewConfig()
 
 	saramaConfig.Version = sarama.V2_0_0_0
 	saramaConfig.Consumer.Offsets.Initial = sarama.OffsetOldest
 
-	if strings.Contains(LoadedConfig.Kafka.Sasl.Protocol, "SSL") {
-		saramaConfig.Net.TLS.Enable = true
-	}
```
jlsherrill, line 94: was this unnecessary?

rverdile: good catch

rverdile: I think I just missed it copying/pasting

pkg/config/event.go
rverdile: agreed, it could be confusing. I renamed everything to "events".

rverdile: I also renamed `pkg/event -> pkg/kafka` and `pkg/notifications -> pkg/event`. I think these names better represent what the packages were being used for.

jlsherrill:
You'll need to add the new topic name here:  https://github.com/content-services/content-sources-backend/blob/main/deployments/deployment.yaml#L21-L27

Otherwise this looks good and behave as expected !

jlsherrill (commented):
ACK, but we may need to reach out to Qe about the test failure.  I'll rekick it for good measure

jlsherrill (approved):

swadeley:
> ACK, but we may need to reach out to Qe about the test failure. I'll rekick it for good measure

Hi, the one fail in the test run is related to the read-only user in the rbac test. There have been some changes to user creation code. So, not related to this PR.

PR #567, opened 2024-02-06
Fixes 3353: Snapshot  inconsistent for API/UI

Andrewgdewar:
## Summary

Snapshot create-on-edit is inconsistent between API and UI 
This was caused by the URL not being checked correctly before enqueueing snapshot/introspect tasks. 
This card corrects that issue.

## Testing steps

- From the api or UI, update a repository (patch) with the original URL in the request.  
Without this change, introspection/snapshot tasks will always be enqueued. 
With it, the tasks will correctly, be ignored, as those two tasks are contingent on a URL change.

### DIFF
--- a/pkg/handler/repositories.go
+++ b/pkg/handler/repositories.go
@@ -313,7 +313,7 @@ func (rh *RepositoryHandler) update(c echo.Context, fillDefaults bool) error {
 		return ce.NewErrorResponse(ce.HttpCodeForDaoError(err), "Error updating repository", err.Error())
 	}
 
-	if urlUpdated && repoParams.URL != nil && repoConfig.URL != *repoParams.URL {
+	if urlUpdated {
 		snapInProgress, err := rh.DaoRegistry.TaskInfo.IsSnapshotInProgress(orgID, repoConfig.RepositoryUUID)
 		if err != nil {
 			return ce.NewErrorResponse(ce.HttpCodeForDaoError(err), "Error checking if snapshot is in progress", err.Error())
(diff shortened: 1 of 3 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-3353

swadeley:
lgtm

pkg/handler/repositories.go
```diff
@@ -313,7 +313,9 @@ func (rh *RepositoryHandler) update(c echo.Context, fillDefaults bool) error {
 		return ce.NewErrorResponse(ce.HttpCodeForDaoError(err), "Error updating repository", err.Error())
 	}
 
-	if repoParams.URL != nil && repoConfig.URL != *repoParams.URL {
+	urlIsDifferent := repoConfig.URL != *repoParams.URL && urlUpdated
```
rverdile, line 316: What do you think moving this comparison logic into the dao Update method and using the `urlUpdated` variable to represent if the URL has changed. I believe that was the intention of that return value.

Andrewgdewar: Hmm good point. I'll do that.

Andrewgdewar: Thanks @rverdile, much cleaner.

pkg/handler/repositories.go
```diff
@@ -313,7 +313,7 @@ func (rh *RepositoryHandler) update(c echo.Context, fillDefaults bool) error {
 		return ce.NewErrorResponse(ce.HttpCodeForDaoError(err), "Error updating repository", err.Error())
 	}
 
-	if repoParams.URL != nil && repoConfig.URL != *repoParams.URL {
+	if urlUpdated && repoParams.URL != nil && repoConfig.URL != *repoParams.URL {
```
rverdile, line 316: it seems like `urlUpdated` already checks `repoParams.URL != nil && repoConfig.URL != *repoParams.URL`, so I think you can remove everything to the right of `urlUpdated`.

Andrewgdewar: You are correct good sir! 
Thank you.

rverdile (approved):
lgtm!

PR #569, opened 2024-02-09
Fixes 3444: append /api/ to all pulp hrefs in db

xbhouse:
## Summary

Job runs on stage that updates pulp hrefs (version_href, publication_href, distribution_href) to prepend /api if not already. 

## Testing steps

- Run query on local snapshots table in db. Shouldn't update any rows if hrefs are already prepended with /api
- *Maybe* run query on stage db?  I haven't done this yet myself though
- 
## Checklist

### DIFF
--- a/cmd/cleanup_versions/main.go
+++ b/cmd/cleanup_versions/main.go
@@ -24,10 +24,24 @@ func main() {
 		log.Fatal().Msg("Requires arguments: --force")
 	}
 
-	result := db.DB.Exec("DELETE FROM snapshots")
+	query :=
+		`		
+			UPDATE snapshots
+			SET version_href = CONCAT('/api', version_href)
+			WHERE version_href NOT LIKE '/api%';
+			
+			UPDATE snapshots
+			SET publication_href = CONCAT('/api', publication_href)
+			WHERE publication_href NOT LIKE '/api%';
+
+			UPDATE snapshots
+			SET distribution_href = CONCAT('/api', distribution_href)
+			WHERE distribution_href NOT LIKE '/api%';
+		`
+	result := db.DB.Exec(query)
 	if result.Error != nil {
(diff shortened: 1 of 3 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-3444

xbhouse:
i'm not sure what caused this test failure. and when running the tests locally they just hang, never gets to the rbac tests :/

cmd/cleanup_versions/main.go
```diff
@@ -24,10 +24,24 @@ func main() {
...
+			SET distribution_href = CONCAT('/api', distribution_href)
+			WHERE distribution_href NOT LIKE '/api%';
+		`
+	result := db.DB.Exec(query)
 	if result.Error != nil {
-		log.Fatal().Err(result.Error).Msg("Could not delete snapshots.")
+		log.Fatal().Err(result.Error).Msg("Could not update hrefs.")
 	} else {
-		log.Warn().Msgf("Deleted %v snapshots", result.RowsAffected)
+		log.Warn().Msgf("Updated %v hrefs", result.RowsAffected)
```
jlsherrill, line 45: I  believe this count actually only prints out the last query's updates.   As i modified just the version_href, it updated it, but i only got 'Updated 0 hrefs'.  So it may make sense to run each one individually, in reality, in stage, the number should be the same generally as all 3 will need to be updated on each row.  

So in the end, I think this is fine since this should only actually run once.

jlsherrill (approved):

cmd/cleanup_versions/main.go
xbhouse: oh yep for each row 3 hrefs will be updated at the most. so should i change this to `updated %v snapshots` or to log after each query? or do you think this is ok for now?

PR #570, opened 2024-02-09
Fixes 3491: bootstrap candlepin connection

jlsherrill:
## Summary

This adds an initial candlepin client bootstrap process and adds a couple items to our compose-up process:
* creates an org
* imports a manifest

The thought is that in stage/prod we will use the existing orgs that match the user's org
In dev, all actions will use this created 'devel org'
This will be controlled via the devel_org boolean config 

## Testing steps

Checkout this pr
update your config.yaml with the new example values.

```go get ./...```
```make compose-down compose-up```
``` go run cmd/candlepin/main.go list-contents```  # list the contents of the devel org

## Checklist

### CODE REVIEW

**Comment by jlsherrill** [comment]
https://issues.redhat.com/browse/HMS-3491

File: pkg/candlepin_client/manifest.go [hunk 1]
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	caliri "github.com/content-services/caliri/release/v4"
+)
+
+func (c *cpClientImpl) ImportManifest(filename string) error {
+	file, err := os.Open(filename)
```
rverdile, line 11: probably want to do a `defer file.Close()`

rverdile (commented):
Looks good overall. I tested it and both commands work. Only thing I found was the one small comment I left.

pkg/candlepin_client/manifest.go
jlsherrill: great catch!

rverdile: I see a rebase, but I don't see a change here. Did you forget to update this?

jlsherrill: arg, forgot to pop the stash after the rebase, thanks!

PR #587, opened 2024-02-26
Fixes 3488: combine introspect and snapshot statuses

xbhouse:
## Summary

- Combines introspection and snapshot statuses into one field returned in the RepositoryResponse. Chart of combination logic can be found [here](https://issues.redhat.com/browse/HMS-3488)
- `status` removed from the db and replaced with `last_introspection_status`
- Handles filtering on the combined status
- Sorting is enabled on the `last_introspection_status` only. PR with the UI change to disable sorting on the new combined status can be found [here](https://github.com/content-services/content-sources-frontend/pull/219)

## Testing steps

- Introspect and snapshot a repository, the `status` returned in the response should reflect the combined status
- The different combinations can be difficult to simulate, one way to do this is to manually change the snapshot task status (in the tasks table) and the `last_introspection_status` (in the repositories table) in the db

## Checklist


db/migrations/20240215104631_add_last_introspection_status.up.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE repositories
+    RENAME COLUMN status to last_introspection_status;
```
jlsherrill, line 4: we typically try to write the up migrations to be idempotent.  You can't do something quite as easy as "IF EXISTS", but there are a couple solutions : https://stackoverflow.com/questions/48102295/rename-column-only-if-exists

pkg/dao/repository_configs.go
```diff
@@ -372,7 +372,34 @@ func (r repositoryConfigDaoImpl) filteredDbForList(OrgID string, filteredDB *gor
...
+		}
+
+		// retrieve all uuids from the filtered repo configs by status
+		var filteredRepoConfigUuids []string
+		for _, repoConfig := range filteredRepoConfigsByStatus {
+			filteredRepoConfigUuids = append(filteredRepoConfigUuids, repoConfig.UUID)
+		}
+
+		filteredDB = filteredDB.
+			Where("repository_configurations.uuid IN ?", filteredRepoConfigUuids)
```
jlsherrill, line 402: Is there any way to make this filter into a query?  I could see it quite complicated, and we'd have to essentially duplicate the combineIntrospectionAndSnapshotStatuses function to generate a where clause.  It think i'd be  something like this (this isn't complete though):

```
if filterData.Status == 'Pending' {
  return db.Where("LastIntrospectionStatus == 'pending' ").
                    Where( 
                             db.Where(LastSnapshotTask.STATUS is null").
                            Or("repoConfig.LastSnapshotTask.Status == 'running'") )
}
```

and it may be easier to just write the complicated Where() clause as pure sql if you can:

```
db.Where("LastIntrospectionStatus == 'pending' 
                     AND (LastSnapshotTask.STATUS is null OR 
                               repoConfig.LastSnapshotTask.Status == 'running')")
```

I think gorm allows that?

jlsherrill: and the reason to do this, is for performance.  It keeps the query in a single db call, rather than fetching having to fetch all results and re-query.  

Also there's a limit to the number of items you can put in an IN  clause

xbhouse: ah yea that makes sense, i'll work on changing this to a query

xbhouse: i've pushed up what i have so far for the query, still seeing incorrect results when filtering on multiple statuses though. with multiple statuses, the filter will only return the results for the first status in the list. i think it's the way i'm writing the where clauses maybe. let me know if you have ideas

pkg/dao/repository_configs.go
```diff
@@ -372,7 +372,35 @@ func (r repositoryConfigDaoImpl) filteredDbForList(OrgID string, filteredDB *gor
...
+					Or("repositories.last_introspection_status = 'Invalid' AND repository_configurations.last_snapshot_uuid IS NOT NULL AND tasks.type = 'snapshot' AND tasks.status = 'failed'").
+					Or("repositories.last_introspection_status = 'Valid' AND repository_configurations.last_snapshot_uuid IS NOT NULL AND tasks.type = 'snapshot' AND (tasks.status = 'running' OR tasks.status = 'failed')")
+			}
+			if status == "invalid" {
+				filteredDB = filteredDB.Where(
+					"repositories.last_introspection_status = 'Invalid' AND tasks.type = 'snapshot' AND tasks.status = 'completed'").
+					Or("repositories.last_introspection_status = 'Unavailable' AND repository_configurations.last_snapshot_uuid IS NULL AND tasks.type = 'snapshot' AND tasks.status = 'failed'").
+					Or("repositories.last_introspection_status = 'Valid' AND repository_configurations.last_snapshot_uuid IS NULL AND tasks.type = 'snapshot' AND tasks.status = 'failed'")
+			}
+		}
```
jlsherrill, line 403: sorry for the delay!   I think what you have is real close, the part that you were missing, is that with multiple statuses, you'd want the equivalent of something like:
```
filteredDB.Where(  db.Where(STMT1).Or(db.Where(STMT2)).Or(db.Where(STMT3))   )
```

I played around with it to make sure it worked (This pulls the statement generation out to a method as well):

```
	if filterData.Status != "" {
		statuses := strings.Split(filterData.Status, ",")

		filteredDB = filteredDB.
			Joins("LEFT JOIN tasks ON repository_configurations.last_snapshot_task_uuid = tasks.id").
			Preload("Repository").
			Preload("LastSnapshotTask")

		var filterChain *gorm.DB
		for _, status := range statuses {
			if filterChain == nil {
				// first where statement, so generate just it
				filterChain = getStatusFilter(status, r.db)
			} else {
				// after the first, OR it with the previous statements
				filterChain = filterChain.Or(getStatusFilter(status, filteredDB))
			}
		}
		filteredDB = filteredDB.Where(filterChain)
	}
	return filteredDB
}

func getStatusFilter(status string, filteredDB *gorm.DB) *gorm.DB {
	status = strings.ToLower(status)
	if status == "valid" {
		filteredDB = filteredDB.Where("(repositories.last_introspection_status = 'Valid' AND tasks.type = 'snapshot' AND tasks.status = 'completed')")
	}
	if status == "pending" {
		filteredDB = filteredDB.Where(
			"repositories.last_introspection_status = 'Pending' AND (repository_configurations.last_snapshot_task_uuid IS NULL OR (tasks.type = 'snapshot' AND (tasks.status = 'running' OR tasks.status = 'completed')))").
			Or("repositories.last_introspection_status = 'Valid' AND repository_configurations.last_snapshot_uuid IS NULL AND tasks.type = 'snapshot' AND tasks.status = 'running'")
	}
	if status == "unavailable" {
		filteredDB = filteredDB.Where(
			"repositories.last_introspection_status = 'Unavailable' AND repository_configurations.last_snapshot_uuid IS NOT NULL AND tasks.type = 'snapshot' AND (tasks.status = 'failed' OR tasks.status = 'completed')").
			Or("repositories.last_introspection_status = 'Invalid' AND repository_configurations.last_snapshot_uuid IS NOT NULL AND tasks.type = 'snapshot' AND tasks.status = 'failed'").
			Or("repositories.last_introspection_status = 'Valid' AND repository_configurations.last_snapshot_uuid IS NOT NULL AND tasks.type = 'snapshot' AND (tasks.status = 'running' OR tasks.status = 'failed')")
	}
	if status == "invalid" {
		filteredDB = filteredDB.Where(
			"repositories.last_introspection_status = 'Invalid' AND tasks.type = 'snapshot' AND tasks.status = 'completed'").
			Or("repositories.last_introspection_status = 'Unavailable' AND repository_configurations.last_snapshot_uuid IS NULL AND tasks.type = 'snapshot' AND tasks.status = 'failed'").
			Or("repositories.last_introspection_status = 'Valid' AND repository_configurations.last_snapshot_uuid IS NULL AND tasks.type = 'snapshot' AND tasks.status = 'failed'")
	}
	return filteredDB
}
```

jlsherrill: with this, the tests run, but the results fail because i presume the test needs updating

pkg/dao/repository_configs_test.go
```diff
@@ -1293,9 +1293,9 @@ func (suite *RepositoryConfigSuite) TestListFilterStatus() {
 	assert.Equal(t, 20, len(response.Data))
```
xbhouse, line 1293: @jlsherrill i think this test failure might be caused by the repoconfig seed options not including references to tasks, so the data returned doesn't include all the repos expected. but i could be wrong. how do you think i should update this?

pkg/dao/repository_configs.go
```diff
@@ -372,7 +372,53 @@ func (r repositoryConfigDaoImpl) filteredDbForList(OrgID string, filteredDB *gor
...
+				if filterChain == nil {
+					// first where statement, so generate just it
+					filterChain = getStatusFilter(status, r.db)
+				} else {
+					// after the first, OR it with the previous statements
+					filterChain = filterChain.Or(getStatusFilter(status, r.db))
+				}
+			default:
+				log.Error().Msgf("Invalid status provided: %v", status)
+				return filteredDB
```
jlsherrill, line 394: i think an invalid status should probably result in a 400 or no results (probably a 400 is better), versus here it'd just return all results.

jlsherrill:
overall looks good, just one small comment

pkg/dao/repository_configs.go
xbhouse: makes sense, pushed up that fix

xbhouse: woops looks like i've changed it to send a 404, let me fix that

jlsherrill (approved):

jlsherrill:
rebased against master

Andrewgdewar:
I believe that you may need to update the API to handle manual snapshots.
See my comment about the front-end PR [here](https://github.com/content-services/content-sources-frontend/pull/234#issuecomment-2018637501).

mshriver:
We've got some suspicious test failures in ephemeral right now with this straightforward update, but its time to move to the stage environment with this and continue investigation there. Thanks for your patience @xbhouse!

PR #595, opened 2024-03-05
Fixes 3490: update distributions for create/update templates

rverdile:
## Summary
This creates/updates/deletes distributions in pulp when a template is created or updated (and repositories are added/removed).

See the JIRA card for the breakdown of the distribution paths. In summary, distributions are created under a path associated to each template. And the paths for red hat repos and custom repos are different.

## Testing steps
1. These testing steps are just a starting point, there's a lot of cases to test here.
2. Sync 2 repositories.
3. Create a template and include the first repository.
4. After the template is created, grab the distribution_href from the "templates_repository_configurations" table. Query pulp for that href. Make sure it is created and that the path matches the requirements of the JIRA card.
5. Patch the template, including the first and second repositories in the request. Check that the old repository's distribution persists and that a new one is created for the second repository.
6. Patch the template again, removing the second repository. The distribution for that repository should be deleted.
7. If you have the means to host your own repositories, you'll also want to test changing the date of the template. I'm not sure how to test this otherwise.

## Checklist


jlsherrill:
https://issues.redhat.com/browse/HMS-3490

cmd/content-sources/main.go
```diff
@@ -45,6 +46,11 @@ func main() {
 	}
 	defer db.Close()
 
+	err = dao.SetupGormTable(db.DB)
+	if err != nil {
+		log.Fatal().Err(err).Msg("Failed to setup gorm table.")
+	}
+
```
rverdile, line 53: This is needed to use soft delete on a joined table i.e. "template_repository_configurations". Not sure if this is the best way to handle this, but it's something you just run once. I feel like it's easy to forget about this way.

pkg/tasks/update_template_distributions_test.go
```diff
@@ -0,0 +1,4 @@
+package tasks
+
+type UpdateTemplateDistributionsSuite struct {
+}
```
rverdile, line 4: I'll write this test after some review because it's all mocking and if anything changes this will definitely change with it.

rverdile: I'm also not sure if these tests have any value, since the vast majority of the code in them is mocking. I imagine anything covered here is covered by the integration test.

pkg/tasks/update_template_distributions.go
```diff
@@ -0,0 +1,200 @@
...
+	daoReg         *dao.DaoRegistry
+	pulpClient     pulp_client.PulpClient
+	payload        *payloads.UpdateTemplateDistributionsPayload
+	task           *models.TaskInfo
+	queue          *queue.Queue
+	ctx            context.Context
+	logger         *zerolog.Logger
+}
+
+func (t *UpdateTemplateDistributions) Run() error {
```
rverdile, line 75: I'm not sure if they way I've architect-ed the code here is confusing to follow. The way I organized the the flow of this task is a little bit the result of me finding bugs and then fixing them. I'm very much open to other ways handling the logic here.

cmd/content-sources/main.go
jlsherrill: would it make sense to run it in db.Connect() ?

pkg/tasks/update_template_distributions.go
```diff
@@ -0,0 +1,200 @@
...
+				return err
+			}
+		}
+
+		if reposUnchanged != nil && slices.Contains(reposUnchanged, repoConfigUUID) {
+			distHref, err := t.daoReg.Template.GetDistributionHref(t.payload.TemplateUUID, repoConfigUUID)
+			if err != nil {
+				return err
+			}
+			taskHref, err := t.pulpClient.UpdateRpmDistribution(distHref, snapshots[snapIndex].PublicationHref, distName+"crackerjack", distributionPath)
```
jlsherrill, line 171: i assume crackerjack was a test string you were using?

pkg/tasks/update_template_distributions.go
```diff
@@ -0,0 +1,200 @@
...
+			path, err := getRHRepoContentPath(repo.URL)
+			if err != nil {
+				return err
+			}
+			t.pulpClient = t.pulpClient.WithDomain(t.rhDomainName)
+			distributionPath = fmt.Sprintf("templates/%v/%v", t.payload.TemplateUUID, path)
+		} else {
+			t.pulpClient = t.pulpClient.WithDomain(t.domainName)
+			distributionPath = fmt.Sprintf("templates/%v/%v", t.payload.TemplateUUID, repoConfigUUID)
+		}
```
jlsherrill, line 110: you might extract this to a method

pkg/tasks/update_template_distributions_test.go
jlsherrill: I mostly agree.  Anything you can pull out to make smaller, and test that would be good (but i only found the one example)

pkg/tasks/update_template_distributions.go
```diff
@@ -0,0 +1,200 @@
...
+			if err != nil {
+				return err
+			}
+			_, err = t.pulpClient.PollTask(taskHref)
+			if err != nil {
+				return err
+			}
+
+			keepRepoConfigUUIDs := append(reposUnchanged, reposAdded...)
+			err = t.daoReg.Template.DeleteTemplateRepoConfigs(t.payload.TemplateUUID, keepRepoConfigUUIDs)
```
jlsherrill, line 160: This is run once per repoConfig (since its in the `for _, repoConfigUUID := range allRepos {` loop , but seems like it probably only needs to be run once at the end?

jlsherrill: because it deletes all repo configs that aren't in the 'keep list'

pkg/tasks/update_template_distributions.go
jlsherrill: One possible way i could see restructuring would be to extract 3 method (one to handle repo additions, one to handle repo updates, and one to handle repo deletions) and instead of looping over allRepos, loop over reposAdded, reposRemoved, reposUnchanged etc... in each method

```
reposAdded, reposRemoved, reposUnchanged, allRepos, err := t.daoReg.Template.GetRepoChanges(t.payload.TemplateUUID, t.payload.RepoConfigUUIDs)
```

rverdile: good idea. made that change.

cmd/content-sources/main.go
rverdile: It does make sense, but that causes a circular dependency error between db and models

pkg/tasks/update_template_distributions_test.go
rverdile: Added one test here

cmd/content-sources/main.go
```diff
@@ -107,6 +113,7 @@ func kafkaConsumer(ctx context.Context, wg *sync.WaitGroup, metrics *m.Metrics)
 		wrk.RegisterHandler(config.RepositorySnapshotTask, tasks.SnapshotHandler)
 		wrk.RegisterHandler(config.DeleteRepositorySnapshotsTask, tasks.DeleteSnapshotHandler)
 		wrk.RegisterHandler(config.DeleteTemplatesTask, tasks.DeleteTemplateHandler)
+		wrk.RegisterHandler(config.UpdateTemplateDistributionsTask, tasks.UpdateTemplateDistributionsHandler)
```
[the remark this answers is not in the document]
rverdile: added it to the list

pkg/tasks/update_template_distributions.go
rverdile: good idea, made it a function

cmd/content-sources/main.go
jlsherrill: ah i see, this is tricky :/   

one thing that might simplify it is to have the method call be fatal, maybe rename it like SetupGormTableOrFail  to make it explicit

pkg/dao/interfaces.go
```diff
@@ -52,6 +53,14 @@ func GetDaoRegistry(db *gorm.DB) *DaoRegistry {
 	return &reg
 }
 
+func SetupGormTable(db *gorm.DB) error {
+	err := db.SetupJoinTable(models.Template{}, "RepositoryConfigurations", models.TemplateRepositoryConfiguration{})
+	if err != nil {
+		return fmt.Errorf("error setting up join table for templates_repository_configurations: %w", err)
+	}
+	return nil
+}
```
jlsherrill, line 62: it might help to document the reasoning behind why we're calling this method

pkg/dao/templates.go
```diff
@@ -115,16 +119,16 @@ func (t templateDaoImpl) insertTemplateRepoConfigs(tx *gorm.DB, templateUUID str
 
 	err := tx.Clauses(clause.OnConflict{
 		Columns:   []clause.Column{{Name: "template_uuid"}, {Name: "repository_configuration_uuid"}},
-		DoNothing: true,
+		DoUpdates: clause.AssignmentColumns([]string{"deleted_at"}),
```
jlsherrill, line 122: would this cause the deleted_at column to be nulled out in the case where a user removes a repo and then adds it back?

pkg/dao/templates.go
```diff
@@ -115,16 +119,16 @@ func (t templateDaoImpl) insertTemplateRepoConfigs(tx *gorm.DB, templateUUID str
...
 	if err != nil {
 		return t.DBToApiError(err)
 	}
 	return nil
 }
 
-func (t templateDaoImpl) removeTemplateRepoConfigs(tx *gorm.DB, templateUUID string, keepRepoConfigUUIDs []string) error {
-	err := tx.Where("template_uuid = ? AND repository_configuration_uuid not in ?", UuidifyString(templateUUID), UuidifyStrings(keepRepoConfigUUIDs)).
+func (t templateDaoImpl) DeleteTemplateRepoConfigs(templateUUID string, keepRepoConfigUUIDs []string) error {
+	err := t.db.Unscoped().Debug().Where("template_uuid = ? AND repository_configuration_uuid not in ?", UuidifyString(templateUUID), UuidifyStrings(keepRepoConfigUUIDs)).
```
jlsherrill, line 131: probably don't need the Debug() here.

jlsherrill:
few comments, otherwise looks good.  doing some testing

pkg/handler/templates.go
```diff
@@ -184,11 +188,14 @@ func (th *TemplateHandler) update(c echo.Context, fillDefaults bool) error {
...
 		tempParams.FillDefaults()
 	}
-	apiTempl, err := th.DaoRegistry.Template.Update(orgID, uuid, tempParams)
+	respTemplate, err := th.DaoRegistry.Template.Update(orgID, uuid, tempParams)
 	if err != nil {
 		return ce.NewErrorResponse(ce.HttpCodeForDaoError(err), "Error updating template", err.Error())
 	}
-	return c.JSON(http.StatusOK, apiTempl)
+
+	th.enqueueUpdateTemplateDistributionsEvent(c, orgID, respTemplate.UUID, respTemplate.Date.String(), tempParams.RepositoryUUIDS)
```
jlsherrill, line 196: I think it'd make more sense to lookup the template date from the template at run time.  That way if a user changed it multiple times in rapid succession, or if a task was retried, it'd use whatever the current date is, not what the user set on it when the task was created

jlsherrill:
seeing an issue where i tried to simulate having snapshots of different dates by:

1.  creating a custom repo, letting it snapshot
2. change the url and let it snapshot again
3. run this sql command (or similar) to update the first snapshot's created_at date:
```update snapshots set created_at = '2024-03-19 09:41:24.568133+00' where uuid = 'f7454454-a6cd-47cd-a8d2-c34f34a707cd';
```
4.  Create a template and change the date to the 19th, or the 21st.  I'd expect the contents of the template path for that repo to change, but i don't see them change.  I added some debugging and it seems like the distribution is being updated with the same publication no matter which date i picked.  It always picks the one from the 19th

I thought it might be an issue with the FetchSnapshotsModelByDateAndRepository method, but when i call the api that uses it, that seems fine:

```

####
POST http://localhost:8000/api/content-sources/v1.0/snapshots/for_date/
x-rh-identity: eyJpZGVudGl0eSI6eyJvcmdfaWQiOiIxNzY4NDYzOCIsICJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJmb28ifSwiYWNjb3VudF9udW1iZXIiOiIxNzY4NDYzOCIsImludGVybmFsIjp7Im9yZ19pZCI6IjE3Njg0NjM4In19fQo=

{
  "repository_uuids": ["5079621c-c97c-4120-9aeb-60d68d3dcc29"],
  "date": "2024-03-20"

}
```

```
  {
    "repository_uuid": "5079621c-c97c-4120-9aeb-60d68d3dcc29",
    "is_after": false,
    "match": {
      "uuid": "67c68114-0a6e-4f50-a6f3-5d09e9419eaf",
      "created_at": "2024-03-21T06:08:49.060373-04:00",
      "repository_path": "cf98e9d7/5079621c-c97c-4120-9aeb-60d68d3dcc29/4866867f-3fe0-4c20-99f3-a0b342964bfa",
      "content_counts": {
        "rpm.package": 1
      },
```

pkg/tasks/update_template_distributions.go
```diff
@@ -0,0 +1,319 @@
...
+		return err
+	}
+
+	template, err := t.daoReg.Template.Fetch(t.orgId, t.payload.TemplateUUID)
+	if err != nil {
+		return err
+	}
+
+	date := template.Date.Format(time.DateOnly)
+	l := api.ListSnapshotByDateRequest{Date: date, RepositoryUUIDS: allRepos}
```
rverdile, line 95: This was this fix for that issue you saw. The type of the `Date` field in the template create/update requests (time.Time) is different from the one in the fetchSnapshotsForDate request (string), which is why it was still working on the latter endpoint.

rverdile: I was thinking we might want to make those consistent between all the endpoints, but I went the least intrusive change for now.

rverdile: The `fetchSnapshotsForDate` endpoint expects the "yyyy-mm-dd" and doesn't seem to work correctly for other time/date formats. If the type there was time.Time, you couldn't do "yyyy-mm-dd", but it could match the time/date format returned by our API.

rverdile: I opened a bug: https://issues.redhat.com/browse/HMS-3832

jlsherrill:
i noticed that if i select a repo that has no snapshots, i get an error:  

```
1:25PM ERR recovered panic in worker with error: runtime error: index out of range [-1] error="runtime error: index out of range [-1]" request_id=vWLyHErsQHhairrylRETYxAZixLBEyKE task_id=c78c02f7-676d-4fd8-a0f9-b258143fde3e task_type=update-template-distributions
```

Its probably a situation we should warn against in the ui, but ideally the task wouldn't error and would just skip that repo

jlsherrill:
just the issue with the error, otherwise this is working great!

jlsherrill (approved):

PR #597, opened 2024-03-06
Build: zest update

jlsherrill:
## Testing steps
tests pass
## Checklist


mshriver:
The pr-check tests are going to fail until an iqe-core update is released - which is actually blocked by the app being broken in stage. The job at least built and deployed the content-sources-backend image successfully to ephemeral.

mshriver (approved):

PR #602, opened 2024-03-12
Build: move pulp to a ordinary dependency

jlsherrill:
## Summary
Since with the default configuration in ephemeral, we do require pulp currently

## Testing steps

deployment passes

## Checklist


jlsherrill:
merging as the deployment occured without error

PR #611, opened 2024-03-18
Build: switch to official candlepin images

jlsherrill:
## Summary

switches to the candlepin:dev-latest image. 

Uses an init file to generate a 2nd db/user for the postgres container

## Checklist


Andrewgdewar (approved):
Tested and working!

PR #612, opened 2024-03-20
Build: handle selinux on postgres mount

jlsherrill:
## Checklist


rverdile (approved):
fixes it!

PR #623, opened 2024-04-02
Refs IQE-2748: Revert old sel testing image backend

akhil-jha:
New sel image is failing the test pipeline because it does not have a controlled shutdown command, making the pod stuck and resulting into timeout.


jlsherrill:
https://issues.redhat.com/browse/HMS-IQE-2748

mshriver (approved):

PR #627, opened 2024-04-09
Refs 3857: improve performance of rpms search query

jlsherrill:
## Summary
Improve performance of rpms search query

## Testing steps
I wasn't fully able to reproduce the ~20 seconds we are seeing in stage, but i could reproduce ~200 ms.

1.  I would turn off snapshotting to make it easier, edit config.yaml, turn the snapshotting feature off
2. ```make repos-import```
3. ```go run cmd/external-repos/main.go nightly-jobs```
4.  run this to create 10,000 'instances' of the repo:
```
for i in `seq 1 10000`; do curl -X POST  http://localhost:8000/api/content-sources/v1.0/repositories/ -H "`./scripts/header.sh $RANDOM$RANDOM`" -H "Content-Type: application/json" -d '{"name":"FOO", "url": "https://dl.fedoraproject.org/pub/epel/9/Everything/x86_64/"}'; done
```
5.  Run this request:

```
####
POST http://localhost:8000/api/content-sources/v1.0/rpms/names
x-rh-identity: eyJpZGVudGl0eSI6eyJvcmdfaWQiOiI5IiwgInR5cGUiOiJVc2VyIiwidXNlciI6eyJ1c2VybmFtZSI6ImZvbyJ9LCJhY2NvdW50X251bWJlciI6ImZvbyIsImludGVybmFsIjp7Im9yZ19pZCI6IjkifX19Cg==

{
  "urls": ["https://dl.fedoraproject.org/pub/epel/9/Everything/x86_64/"],
  "search": "kernel"
}
```

On main, i was seeing requests take ~250  ms, with this change, it dropped to ~60-80ms

## Checklist

### DIFF
--- a/pkg/dao/rpms.go
+++ b/pkg/dao/rpms.go
@@ -162,35 +162,23 @@ func (r rpmDaoImpl) Search(orgID string, request api.ContentUnitSearchRequest) (
 	urls := handleTailChars(request)
 	uuids := request.UUIDs
 
-	// This implement the following SELECT statement:
-	//
-	// SELECT DISTINCT ON (rpms.name)
-	//        rpms.name, rpms.summary
-	// FROM rpms
-	//      inner join repositories_rpms on repositories_rpms.rpm_uuid = rpms.uuid
-	//      inner join repositories on repositories.uuid = repositories_rpms.repository_uuid
-	//      left join repository_configurations on repository_configurations.repository_uuid = repositories.uuid
-	// WHERE (repository_configurations.org_id = 'acme' OR repositories.public)
-	//       AND ( repositories.url in (...)
-	//             OR repository_configurations.uuid in (...)
-	//       )
-	//       AND rpms.name LIKE 'demo%'
-	// ORDER BY rpms.name, rpms.epoch DESC
-	// LIMIT 20;
+	// Lookup repo uuids to search
+	repoUuids := []string{}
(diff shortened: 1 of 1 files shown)


pkg/dao/rpms.go
```diff
@@ -162,35 +162,23 @@ func (r rpmDaoImpl) Search(orgID string, request api.ContentUnitSearchRequest) (
...
-	//       )
-	//       AND rpms.name LIKE 'demo%'
-	// ORDER BY rpms.name, rpms.epoch DESC
-	// LIMIT 20;
+	// Lookup repo uuids to search
+	repoUuids := []string{}
+	orGroupPublicPrivatePopular := r.db.Where("repository_configurations.org_id = ?", orgID).Or("repositories.public").Or("repositories.url in ?", popularRepoUrls())
+	r.db.Model(&models.Repository{}).
+		Joins("left join repository_configurations on repositories.uuid = repository_configurations.repository_uuid and repository_configurations.org_id = ?", orgID).
+		Where(orGroupPublicPrivatePopular).
```
jlsherrill, line 170: Note that I think this "and repository_configurations.org_id = ?" is actually what is causing most of the speed up here.  Previously i think it was causing the join to include EVERY popular repo_config across orgs in the join now that the orGroupPublicPrivatePopular includes all popular repo.  

I split out the query to lookup the repo UUIDs because i thought it simplified the query quite a bit and made it easier to understand

jlsherrill:
merging to fix stage performance issue

PR #628, opened 2024-04-10
Refs 2950: add rolling distros to default introspection repos

croissanne:
Image builder switched to rolling URLs for the latest minor in order to find snapshots more easily. As a result the package search now happens against the rolling URLs.


swadeley (approved):
ACK

PR #629, opened 2024-04-10
Fixes 3937: use clowder rds ca for tangy db

jlsherrill:
## Summary

Stage/prod dbs need to use ssl, which wasn't being done because we don't have a ca cert.  Lets try using the one clowder is giving us for our db.

## Testing steps

Don't think there is an easy way  to test this


Andrewgdewar (approved):
I take it the pulp failure is expected?

jlsherrill:
yes, and this shouldn't affect integration tests at all

jlsherrill:
merging

PR #634, opened 2024-04-11
Fixes 3940: handle pending snapshot task

xbhouse:
## Summary

Fixes case where repository status is "Unknown" if a snapshot task never reaches a running state 

## Testing steps

- When a snapshot task is `pending` or `running`, the status in the repository response should show `Pending` and the UI should show it's `In progress`
- These `Pending` repositories should be displayed when filtering on status `Pending`
- I tested this by adding a repository and snapshotting, then manually changing the snapshot task status to `pending` in the db and verifying that the status shown in the repository status is `Pending`, but there is probably a better way to simulate this case
- EDIT: see [Ryan's comment](https://github.com/content-services/content-sources-backend/pull/634#issuecomment-2051854053) for a better way to test this

## Checklist

### DIFF
--- a/pkg/dao/repository_configs.go
+++ b/pkg/dao/repository_configs.go
@@ -411,9 +411,9 @@ func getStatusFilter(status string, filteredDB *gorm.DB) *gorm.DB {
 	}
 	if status == "Pending" {
 		filteredDB = filteredDB.Where(
-			"repositories.last_introspection_status = 'Pending' AND (repository_configurations.last_snapshot_task_uuid IS NULL OR (tasks.type = 'snapshot' AND (tasks.status = 'running' OR tasks.status = 'completed')))").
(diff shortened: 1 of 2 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-3488

jlsherrill:
https://issues.redhat.com/browse/HMS-3940

rverdile (approved):
tested and working!

rverdile:
To test without modifying the database, start a bunch of snapshot tasks for different repos so that some would be left in the Pending state. Then if you fetch the repository associated to one of those Pending tasks, you'll no longer see "Unknown" as the repository status.

xbhouse:
> To test without modifying the database, start a bunch of snapshot tasks for different repos so that some would be left in the Pending state. Then if you fetch the repository associated to one of those Pending tasks, you'll no longer see "Unknown" as the repository status.

ah yes, makes sense! thank you!

swadeley:
lgtm, thank you

PR #637, opened 2024-04-15
Fixes 1738: Unify on context for dao & pulp layer

jlsherrill:
## Summary
Unifies context passing into methods across Dao and PulpClient layers from handler and tasks.  Also sets up the logger to properly pull the request ID from the context. Now the context is passed fully from:
* the handler, to the dao, to the pulp_client
* the handler to the task to the doa & pulp client

All dao db calls should now have a .WithContext() call to add the context to the db call.  

## Testing steps

Run a curl request to create (and snapshot a repo):

```
UUID="1cf4b0b0-fc0a-11ee-8913-201e8814f721"
curl -H "x-rh-insights-request-id:$UUID"  http://localhost:8000/api/content-sources/v1/repositories/  -H "`./scripts/header.sh 1 1`"   -d '{"name":"foo", "url":"https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/", "snapshot":true}'  -H "Content-Type: application/json"
```

And watch the server logs:

you should see a lot of:
``` request_id=1cf4b0b0-fc0a-11ee-8913-201e8814f721```
Both at the api level and then as the task runs

Also check  the pulp workers logs:

```podman logs cs_pulp_worker_1 ```
and you should see the same request id there:
```
pulp [1cf4b0b0-fc0a-11ee-8913-201e8814f721]: pulpcore.tasking.tasks:INFO: Task completed 018ee838-baf4-7997-aa91-a4f24bcd82d9
```

## Checklist

### CODE REVIEW

**Comment by jlsherrill** [comment]
https://issues.redhat.com/browse/HMS-1738

File: pkg/dao/helpers.go [hunk 1]
```diff
@@ -112,10 +113,10 @@ func isOwnedRepository(db *gorm.DB, orgID string, repositoryConfigUUID string) (
 	return true, nil
 }
 
-func checkForValidRepoUuidsUrls(uuids []string, urls []string, db *gorm.DB) (bool, bool, string, string) {
+func checkForValidRepoUuidsUrls(ctx context.Context, uuids []string, urls []string, db *gorm.DB) (bool, bool, string, string) {
```
xbhouse, line 116: would checkForValidSnapshotUuids in this file also need a context passed in?

pkg/dao/rpms.go
```diff
@@ -529,8 +530,8 @@ func (r *rpmDaoImpl) DetectRpms(orgID string, request api.DetectRpmsRequest) (*a
 	}
 
 	// find rpms associated with the repositories that match given rpm names
-	orGroupPublicOrPrivate := r.db.Where("repository_configurations.org_id = ?", orgID).Or("repositories.public")
-	db := r.db.
+	orGroupPublicOrPrivate := r.db.WithContext(ctx).Where("repository_configurations.org_id = ?", orgID).Or("repositories.public")
```
xbhouse, line 533: is `WithContext` needed in this where clause since you're calling it again below? i'm not sure if it hurts to add it multiple times or not

xbhouse:
tested and i see the request id passed through, just a couple comments so far. @rverdile could you take a look too when you get a chance? i may have missed things

pkg/dao/interfaces.go
```diff
@@ -66,119 +64,118 @@ func SetupGormTableOrFail(db *gorm.DB) {
...
+	Fetch(ctx context.Context, orgID string, uuid string) (api.TemplateResponse, error)
+	List(ctx context.Context, orgID string, paginationData api.PaginationData, filterData api.TemplateFilterData) (api.TemplateCollectionResponse, int64, error)
+	SoftDelete(ctx context.Context, orgID string, uuid string) error
+	Delete(ctx context.Context, orgID string, uuid string) error
+	ClearDeletedAt(ctx context.Context, orgID string, uuid string) error
+	Update(ctx context.Context, orgID string, uuid string, templParams api.TemplateUpdateRequest) (api.TemplateResponse, error)
+	GetRepoChanges(ctx context.Context, templateUUID string, newRepoConfigUUIDs []string) ([]string, []string, []string, []string, error)
+	GetDistributionHref(ctx context.Context, templateUUID string, repoConfigUUID string) (string, error)
+	UpdateDistributionHrefs(ctx context.Context, templateUUID string, repoUUIDs []string, repoDistributionMap map[string]string) error
+	DeleteTemplateRepoConfigs(tctx context.Context, emplateUUID string, keepRepoConfigUUIDs []string) error
```
xbhouse, line 180: typo in templateUUID here

pkg/dao/rpms.go
jlsherrill: It probably isn't needed here only the outer most r.db, but wouldn't hurt anything.  However i added it by mistake here so i'll remove it.

pkg/dao/helpers.go
jlsherrill: good catch! i missed those when doing a rebase.  Will update

jlsherrill: fixing checkForValidSnapshotUuids, but isOwnedRepository actually passes in the DB, so in that case i think its okay to presume that the db is passed in with the context already, at least that is a pattern i used for some other things.

rverdile (commented):
ack from me

swadeley:
Hi
I see many test failures in the pr check #1133:
`Bad Request: x-rh-identity header has an invalid or missing org_id`

xbhouse (approved):
ack! thanks @rverdile

rverdile (changes requested):
Actually I think we need to double check that the request ID is the correct format. Usually they look like this: a9d3e55f2bf040ec995811920e1993d7

I see some errors in the pulp API logs, I'm not sure if they're related: 
```
pulp [None]: django_guid:WARNING: eAQguKUTcRFZtdERBGZnnuLccaFKToTA is not a valid GUID. New GUID is b7e7c7c70fa34845b441d4564553f20e
```

Might just need to update where we generate a request ID locally if one isn't specified.

pkg/router/route.go
```diff
@@ -22,11 +21,12 @@ func ConfigureEcho(allRoutes bool) *echo.Echo {
 	echoLogger := lecho.From(log.Logger,
 		lecho.WithTimestamp(),
 		lecho.WithCaller(),
-		lecho.WithLevel(echo_log.INFO),
 	)
+
 	e.Use(echo_middleware.RequestIDWithConfig(echo_middleware.RequestIDConfig{
 		TargetHeader: config.HeaderRequestId,
 	}))
+	e.Use(middleware.AddRequestId)
```
rverdile, line 29: Lecho tries to get the request ID from the header, not the context. So even if the AddRequestId middleware has added it to the context, lecho can't find it.

The echo middleware generates a request ID, and sets it on the response header, so then that request ID gets used.

jlsherrill: good catch.  I updated it to set it properly on the request/response, as well as not use the echo middleware since it doesn't set it on the context

rverdile (approved):

PR #639, opened 2024-04-16
Fixes 3599: Add candlepin content for templates

rverdile:
## Summary
Adds candlepin product/pool/environment/content creation on template create. If updating a template, content is removed from environments if necessary.

## Testing steps
1. Create  a custom repository and a red hat repository.
2. Create a content template with these repositories included. Wait for task to finish.
3. The will be many things to check in candlepin: the new product created, the new pool created, the new environment created for the template, the new product content, the new content promoted to the environment.

Here is an example to list the products. Make sure you include the Authorization header. It is the base64 encoded username:password (admin:admin by default).
```
GET http://localhost:8181/candlepin/owners/content-sources-test/products
Accept: application/json
Authorization: Basic YWRtaW46YWRtaW4=
```
4. Refer to the [candlepin REST API spec](https://petstore.swagger.io/?url=https://raw.githubusercontent.com/candlepin/candlepin/main/api/candlepin-api-spec.yaml#/) to see how to query more endpoints.
5. Make sure the content URLs and labels are set correctly. Keep in mind, these will be different for custom vs Red Hat repositories.
6. Check the content URL against the base URL of the pulp distribution associated to the repository in the template.
7. Update the template to remove a repository. Wait for the task to finish.
8. The content set for that repository should no longer be in the environment.
9. Reference the JIRA card to make sure you're checking all the things.

## Checklist


pkg/tasks/update_template_distributions.go
```diff
@@ -329,3 +347,215 @@ func getRHRepoContentPath(rawURL string) (string, error) {
 	}
 	return u.Path[1 : len(u.Path)-1], nil
 }
+
+// RunCandlepin creates an environment for the template and content sets for each repository.
+// Each content set's URL is the distribution URL created during RunPulp().
+// May promote or demote content from the environment, depending on if the repository is being added or removed from template.
+// If not created for the given org previously, this will also create a product and pool.
+func (t *UpdateTemplateDistributions) RunCandlepin() error {
```
rverdile, line 355: I saw two ways to combine the tasks here: 
1. Interweave the functionality and modify pulp/candlepin in the same method run. This might reduce the number of for loops by a little. But for candlepin we make use of bulk APIs, so the only place that would really feel the benefit of this is creating content sets.
2. Do everything in pulp, then do everything in candlepin. This way separates the candlepin logic from the pulp logic, which I feel makes the flow much easier to follow.

jlsherrill:
Do you want to go ahead and add context passing and set the X-Correlation-ID header on candelpin requests similar to what i've done here: https://github.com/content-services/content-sources-backend/pull/637
or do it as a followup issue?

xbhouse:
checking the products endpoint, i see the base url of the pulp distribution path (which i guess is also the repoconfig uuid?) in the content url for the custom repo, but not for the red hat repo. is this expected? 

red hat repo:
```
"contentUrl": "templates/34f84a07-60ee-4f19-b811-6aedbcc0f224/content/dist/layered/rhel8/x86_64/ansible/2/os",
"id": "aa19b2338be64ba3a342520979912602",
"label": "ansible-2-for-rhel-8-x86_64-rpms"
```

pulp distribution path: `aa19b233-8be6-4ba3-a342-520979912602/5275e346-a507-47a3-9ecd-0ef982aaece7`

repoconfig uuid: `aa19b233-8be6-4ba3-a342-520979912602`


custom repo:
```
"contentUrl": "templates/34f84a07-60ee-4f19-b811-6aedbcc0f224/d09b6122-5c2e-435f-b341-3a397c02b32f",
"id": "d09b61225c2e435fb3413a397c02b32f",
"label": "test1"
```

pulp distribution path: 
`ec89c87b/d09b6122-5c2e-435f-b341-3a397c02b32f/d6de4b72-f1fb-4fb3-9a27-dbf37835e3b4`

repoconfig uuid: `d09b6122-5c2e-435f-b341-3a397c02b32f`

pkg/candlepin_client/pool.go
```diff
@@ -0,0 +1,40 @@
...
+	if httpResp != nil {
+		defer httpResp.Body.Close()
+	}
+	if len(found) == 0 {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, errorWithResponseBody("couldn't fetch product", httpResp, err)
+	}
+	return &found[0], nil
```
jlsherrill, line 20: I wonder if we should log an error if there is more than 1?

jlsherrill: also, are you sure that this doesn't return red hat pools?  (i'm not actually sure myself)

pkg/dao/snapshots_test.go
```diff
@@ -594,7 +593,7 @@ func (s *SnapshotsSuite) TestGetRepositoryConfigurationFileNotFound() {
 
 	// Test bad snapshot UUID
 	mockPulpClient.On("GetContentPath").Return(testContentPath, nil).Once()
-	repoConfigFile, err := sDao.WithContext(context.Background()).GetRepositoryConfigurationFile(repoConfig.OrgID, uuid2.NewString(), host)
+	repoConfigFile, err := sDao.WithContext(context.Background()).GetRepositoryConfigurationFile(repoConfig.OrgID, uuid.NewString(), host)
```
jlsherrill, line 596: just out of curiosity, why the switch to uuid?

xbhouse:
should the environment include the red hat repos? when querying environments i see only the custom repo referenced in environmentContent, not sure if that's expected. i used this endpoint `http://localhost:8181/candlepin/owners/content-sources-test/environments` 
```
[
    {
        "created": "2024-04-18T17:14:02+0000",
        "updated": "2024-04-18T17:14:02+0000",
        "id": "34f84a0760ee4f19b8116aedbcc0f224",
        "name": "34f84a0760ee4f19b8116aedbcc0f224",
        "type": null,
        "description": null,
        "contentPrefix": null,
        "owner": {
            "id": "2c9780868ef1fe49018ef1fe632a0002",
            "key": "content-sources-test",
            "displayName": "ContentSourcesTest",
            "href": "/owners/content-sources-test",
            "contentAccessMode": "org_environment"
        },
        "environmentContent": [
            {
                "contentId": "d09b61225c2e435fb3413a397c02b32f",
                "enabled": true
            }
        ]
    }
]
```

pkg/tasks/update_template_content.go
```diff
@@ -329,3 +348,215 @@ func getRHRepoContentPath(rawURL string) (string, error) {
...
+	}
+
+	var repoLabel string
+	if repo.OrgID == config.RedHatOrg {
+		repoLabel = repo.Label
+	} else {
+		repoLabel = re.ReplaceAllString(repo.Name, "_")
+	}
+	return repoLabel, nil
+}
```
jlsherrill, line 539: I think we'd need to handle the edge case where two different custom repos would map to the same label

pkg/candlepin_client/environment.go
```diff
@@ -0,0 +1,64 @@
...
+	caliri "github.com/content-services/caliri/release/v4"
+	"github.com/openlyinc/pointy"
+)
+
+func GetEnvironmentID(templateUUID string) string {
+	return strings.Replace(templateUUID, "-", "", -1)
+}
+
+func (c *cpClientImpl) CreateEnvironment(name string, id string, prefix string) (*caliri.EnvironmentDTO, error) {
+	env, httpResp, err := c.client.OwnerAPI.CreateEnv(c.ctx, DevelOrgKey).EnvironmentDTO(caliri.EnvironmentDTO{Id: &id, Name: &name}).Execute()
```
jlsherrill, line 15: I think you need to pass the Org into the CreateEnvironment method

jlsherrill: as the idea is that we would be using a real org in stage/prod

pkg/candlepin_client/product.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	if err != nil {
+		return nil, errorWithResponseBody("couldn't fetch product", httpResp, err)
+	}
+	return found, nil
+}
+
+func (c *cpClientImpl) CreateProduct(ownerKey string) error {
+	productID := GetProductID(ownerKey)
+	found, err := c.FetchProduct(DevelOrgKey, productID)
```
jlsherrill, line 27: same comment here about the DevelOrgKey

pkg/tasks/update_template_content.go
```diff
@@ -329,3 +348,215 @@ func getRHRepoContentPath(rawURL string) (string, error) {
...
+		return err
+	}
+
+	contentPath, err := t.pulpClient.GetContentPath()
+	if err != nil {
+		return err
+	}
+
+	envID := candlepin_client.GetEnvironmentID(t.payload.TemplateUUID)
+	env, err := t.createOrFetchEnvironment(envID, contentPath)
```
jlsherrill, line 403: I'm not sure how well I described the pathing in the jira issue, but because of the way pulp domains work, we actually want to set the prefix of the environment to the red hat domain's content path.  

We then plan to use environment content overrides (which was just merged in candlepin) to override the content paths for the custom repos.  So it will look a little something like this

Prefix:   https://pulp.content.server.com/pulp/RH_DOMAIN/content/
RH Content (which already exists in candlepin) path:  /content/dist/rhel8/8/x86_64/baseos/os

Which means we can't just simply expose custom content in the same way, and custom repos will be broken right now.  The plan is to use content overrides, to override the path to be:

Override URL:  https://pulp.content.server.com/pulp/ORG_DOMAIN/content/path/to/content

This environment override was *just* merged in candlepin in 4.4.6, and i figured we'd just leave custom content broken here, but we could go ahead and integrate that here too, or still do it as a followup PR.

pkg/candlepin_client/pool.go
rverdile: locally there are no red hat pools, but I guess there normally would be? If that's the case, we'd have to store the pool ID on something in the database I think. Because we're not able to specify the pool ID during creation (we can but it doesn't work).

Since it's one pool per org, I'm thinking to store on the domains table?

pkg/tasks/update_template_content.go
rverdile: Is this case you're talking about? I don't get any errors creating things in this scenario, but I'm assuming this is an issue when consuming the config.repo file.

org1 with repo "my repo" => "my_repo"
org1 with repo "my?repo" => "my_repo"

I'm assuming for different orgs it's okay?

rverdile: Do you think it makes sense to check this in the task and return an error if there's conflicting labels? And then I guess later on we'd somehow expose that error after template creation.

jlsherrill: i'm kinda surprised candlepin doesn't prevent this! but yes, you wouldn't be able to consume both those on the client.  ideally this doesn't cause an error, but may require us to add some randomness to it if its already present.  

For example, 
"my repo"  ->  my_repo
"my?repo" -> my_repo_jrn3oiuf

its an edge case, so ideally not super common

rverdile:
> should the environment include the red hat repos? when querying environments i see only the custom repo referenced in environmentContent

Yes, this was a bug and now it should be fixed

pkg/tasks/update_template_content.go
rverdile: I pushed a change here, but I realize now this says to only add randomness _if already present_, so I need to go back and change that

pkg/tasks/update_template_content.go
rverdile: I didn't add content overrides, but I think the prefix is correct now

pkg/candlepin_client/pool.go
rverdile: I changed this to filter by the product ID, which I think would be good enough?

rverdile:
> checking the products endpoint, i see the base url of the pulp distribution

It's expected that the base URLs in pulp are different. It sorta it is for candlepin too. The content URL of the custom repo content actually doesn't really matter because we're going to override that using the context overrides feature in Candlepin.

rverdile:
Not sure why the test is failing here and not locally, need to investigate

jlsherrill:
I'm in the middle of testing, but i noticed one thing.  For red hat content sets, it looks like we're creating them with a 'randomized' name because they already exists, but for red hat contnet, they should already exist, and we should just use them as is.

```
Content:
	Type: rpm
	Name: Red Hat Enterprise Linux 9 for x86_64 - BaseOS (RPMs)
	Label: rhel-9-for-x86_64-baseos-rpms_cbFyHWHJEp

```

.github/workflows/content-sources-actions.yml
```diff
@@ -134,6 +149,10 @@ jobs:
           CLIENTS_PULP_SERVER: http://localhost:8080
           CLIENTS_PULP_USERNAME: admin
           CLIENTS_PULP_PASSWORD: password
+          CLIENTS_CANDLEPIN_SERVER: http://localhost:8181
```
jlsherrill, line 152: this should probably be  http://localhost:8181/candlepin

rverdile: thanks!

.github/workflows/content-sources-actions.yml
```diff
@@ -122,8 +122,23 @@ jobs:
...
           sleep 30
+      - name: start candlepin
+        uses: isbang/compose-action@v1.4.1
+        with:
+          compose-file: ./deployments/docker-compose.yml
+          services: |
+            candlepin
+          down-flags: --volumes
+        env:
+          CONTENT_DATABASE_PORT: 5434
```
rverdile, line 133: I want to point out that despite specifying the candlepin service, the db service is also being started (probably because the candlepin service is defined as being dependent on it). I set a different database port for the db so it doesn't interfere with the one already running.

pkg/tasks/update_template_content.go
```diff
@@ -0,0 +1,601 @@
...
+		repoLabel := repoLabels[i]
+		repoVendor := getRepoVendor(repo)
+
+		var contentURL string // TODO nothing is set for custom repos. must use content overrides to set baseurl instead.
+		if repo.OrgID == config.RedHatOrg {
+			contentURL, err = getRHRepoContentPath(repo.URL)
+			if err != nil {
+				return nil, nil, err
+			}
+		}
```
jlsherrill, line 544: I believe this is attempting to create red hat content objects (but instead might be accidentally updating them), but we don't actually want to do either of those, instead we should just leave those alone.   The only thing we really should be doing with them is promoting them into the environment

pkg/tasks/update_template_content.go
```diff
@@ -0,0 +1,601 @@
...
+
+		content = append(content, caliri.ContentDTO{
+			Id:         &id,
+			Type:       &repoType,
+			Label:      &repoLabel,
+			Name:       &repoName,
+			Vendor:     &repoVendor,
+			ContentUrl: &contentURL,
```
jlsherrill, line 552: For custom repos a blank url actually break candlepin's cert generation (kinda a bug that they allow it to be created).  We could either set it to the upstream URL of the repo, or to something not functional.  Maybe for now lets set it to the upstream repo?   It should be disabled by default so a user would have to enable it explicitly.

pkg/candlepin_client/pool.go
jlsherrill: yep, i think so too!

jlsherrill:
I noticed we weren't using the template name for the environment name, instead opting for the UUID, was there a reason for that?  These names will eventually show up on the client

rverdile:
> I noticed we weren't using the template name for the environment name, instead opting for the UUID, was there a reason for that? These names will eventually show up on the client

No, I just wasn't thinking about it that way. I can change it to the name of the template.

pkg/tasks/update_template_content.go
rverdile: Oh okay, I misunderstood that requirement on the description of the task. 

So if I'm understanding correctly, I should only do content creation for custom repos. And for red hat repos, I should just get a list of content IDs to promote, because the content already exists.

jlsherrill: yep exactly!  and if for some reason a red hat content set doesn't exist, i would log an error and continue  (I can't imagine a scenario today where that would occur though)

jlsherrill:
created a template and added the two base rhel9 repos and a custom repo and the update-template-content task failed with:

runtime error: index out of range [2] with length 1

when i removed the custom repo from the template it seemed to work fine

pkg/candlepin_client/client.go
```diff
@@ -10,26 +10,26 @@ import (
...
 	"github.com/rs/zerolog/log"
 )
 
 type cpClientImpl struct {
-	client *caliri.APIClient
-	ctx    context.Context
 }
 
 const DevelOrgKey = "content-sources-test"
+const YumRepoType = "rpm"
```
jlsherrill, line 21: This should be 'yum' instead of 'rpm'

pkg/candlepin_client/pool.go
```diff
@@ -0,0 +1,51 @@
...
+
+	productID := GetProductID(ownerKey)
+	found, err := c.FetchPool(ctx, ownerKey, productID)
+	if found != nil || err != nil {
+		return found.GetId(), err
+	}
+
+	endDate := time.Date(2049, time.December, 1, 0, 0, 0, 0, time.UTC).Format(time.RFC3339)
+	startDate := time.Now().Format(time.RFC3339)
+	found, httpResp, err := client.OwnerAPI.CreatePool(ctx, ownerKey).PoolDTO(caliri.PoolDTO{EndDate: &endDate, StartDate: &startDate, ProductId: &productID}).Execute()
```
jlsherrill, line 43: I think we may want to set the quantity here to -1 (which means unlimited).  Since everything here should be SCA it probably doesn't matter, but i think thats a good value to set.  It appears? that it may be defaulting to '1' which is kinda strange

pkg/tasks/update_template_content.go
```diff
@@ -0,0 +1,655 @@
...
+		repoLabel := repoLabels[i]
+		repoVendor := getRepoVendor(repo)
+
+		var contentURL string
+		if repo.OrgID == config.RedHatOrg {
+			contentURL, err = getRHRepoContentPath(repo.URL)
+			if err != nil {
+				return nil, nil, err
+			}
+		} else {
```
jlsherrill, line 596: I think this if block won't ever be hit due to the  'continue' above.

pkg/tasks/update_template_content.go
```diff
@@ -0,0 +1,655 @@
...
+	err = t.cpClient.AddContentBatchToProduct(t.ctx, t.ownerKey, customContentIDs)
+	if err != nil {
+		return err
+	}
+
+	contentPath, err := t.pulpClient.GetContentPath(t.ctx)
+	if err != nil {
+		return err
+	}
+	prefix := contentPath + t.rhDomainName + "/templates/" + t.payload.TemplateUUID
```
jlsherrill, line 400: I noticed that for me this is causing a resulting prefix of:
```
http://pulp.content:8080//pulp/content/80a1acac/templates/fc113564-fbcd-4ef8-8a3b-183015ef2ca7
```

the double //s isn't causing a problem, but you might use url.Parse & path.join() to do this

jlsherrill: https://stackoverflow.com/questions/34668012/combine-url-paths-with-path-join

pkg/tasks/update_template_content.go
```diff
@@ -0,0 +1,655 @@
...
+	if err != nil {
+		return err
+	}
+
+	pulpClient := pulp_client.GetPulpClientWithDomain(domainName)
+	cpClient := candlepin_client.NewCandlepinClient()
+
+	t := UpdateTemplateContent{
+		orgId:          task.OrgId,
+		ownerKey:       candlepin_client.DevelOrgKey,
```
jlsherrill, line 55: We should only use DevelOrgKey if its set (which shouldn't be for stage and prod), for those environments I *think* we can use the org_id as the owner key, but we may need to ask around about that.

rverdile: What do we do for now if it's not set? Right now the CreateOwner method errors if DevelOrg is not set.

jlsherrill: i suspect that the org_id we're storing and getting in the identity header is the owner 'key'.

pkg/tasks/update_template_content.go
rverdile: that's actually coming from the GetContentPath() function. It's because the the origin and prefix are defined with a trailing and leading slash:

```
PULP_CONTENT_ORIGIN: "http://pulp.content:8080/"
PULP_CONTENT_PATH_PREFIX: "/pulp/content/"
```

Still, it should handle that properly, so I'll add some logic there.

pkg/tasks/update_template_content.go
rverdile: so do we unblock the owner creation?

rverdile: Oh actually, this doesn't call CreateOwner(). Nevermind.

pkg/models/repository_configuration.go
```diff
@@ -78,6 +82,9 @@ func (rc *RepositoryConfiguration) BeforeUpdate(tx *gorm.DB) error {
 	if err := rc.validate(); err != nil {
 		return err
 	}
+	if err := rc.SetLabel(); err != nil {
+		return err
+	}
```
jlsherrill, line 87: I'm not sure that we want to update the label today (I'm not 100% sure we even can in candlepin).  We may want to file an issue to handle that later on and properly update it in candlepin if possible.

pkg/candlepin_client/product.go
```diff
@@ -0,0 +1,74 @@
...
+	ctx, client, err := getCandlepinClient(ctx)
+	if err != nil {
+		return err
+	}
+
+	productID := GetProductID(ownerKey)
+
+	contentMap := make(map[string]bool)
+	for _, id := range contentIDs {
+		contentMap[id] = true
```
jlsherrill, line 64: this boolean actually controls whether the content is enabled by default or not.  In this case we do not want it enabled by default (hence why we force  it to true when promoting it to the environment)

pkg/models/repository_configuration.go
```diff
@@ -105,6 +109,14 @@ func (rc *RepositoryConfiguration) ReplaceEmptyValues(tx *gorm.DB) error {
...
 }
 
+func (rc *RepositoryConfiguration) SetLabel() error {
+	label, err := getRepoLabel(*rc)
+	if err != nil {
+		return err
+	}
+	rc.Label = label
+	return nil
+}
```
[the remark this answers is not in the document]
jlsherrill: I think this is my last comment

jlsherrill:
oh one more small thing.  If i comment out the candlepin server URL it get an error on the task after creating a template.  This will be the case with stage for a bit.  We probably want to just not run the task if its not configured.

jlsherrill (approved):

PR #644, opened 2024-04-23
Refs 3897: fix sentry name

rverdile:
## Summary

I'm pretty sure this is the reason we're not getting alerts for service errors. I noticed in the logs that sentry was being configured for the task worker, but not for the service. The name for the task worker is "content-sources-glitchtip". Both pods were working fine in ephemeral if I used an env variable.

I included an error log to verify if this works. If it does, I'll remove the error log in a followup.

## Testing steps
Just merge and see I think

## Checklist


jlsherrill (approved):
makes perfect sense, ack!

PR #645, opened 2024-04-23
Refs 2422: Add pending task metrics

jlsherrill:
## Summary
    Adds new metrics:
    * Average queue time of pending task
    * Length in queue for the oldest pending task
    * Number of pending tasks
    Also:
    * Increases default metric calculations to 30 seconds from 5
    * makes interval for metric calculation configurable
    * adds log level override for metrics so you can set the rest of the app to trace while metrics are set to debugging

## Testing steps

on a new db:

```
make run
```
in another tab:
```
make repos-import
go run cmd/external-repos/main.go nightly-jobs
```
This will kick off a ton of tasks, you can monitor with:
```
 curl localhost:9000/metrics | grep task_stats
```

This will update every ~30 seconds.  

## Checklist

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -124,8 +124,9 @@ type Database struct {
 }
 
 type Logging struct {
-	Level   string
-	Console bool
+	Level        string
+	MetricsLevel string `mapstructure:"metrics_level"`
+	Console      bool
 }
 
 type Certs struct {
@@ -181,6 +182,9 @@ type Metrics struct {
 	// Defines the metrics port that the app should be configured to listen on for
 	// metric traffic.
 	Port int `mapstructure:"port"`
+
+	// How often (in seconds) to run queries to collect some metrics
+	CollectionFrequency int `mapstructure:"collection_frequency"`
 }
 
 const (
@@ -235,9 +239,11 @@ func setDefaults(v *viper.Viper) {
 	v.SetDefault("options.template_event_topic", "platform.content-sources.template")
 	v.SetDefault("options.repository_import_filter", "")
 	v.SetDefault("logging.level", "info")
+	v.SetDefault("logging.metrics_level", "")
 	v.SetDefault("logging.console", true)
 	v.SetDefault("metrics.path", "/metrics")
 	v.SetDefault("metrics.port", 9000)
+	v.SetDefault("metrics.collection_frequency", 60)
 	v.SetDefault("clients.rbac_enabled", true)
 	v.SetDefault("clients.rbac_base_url", "http://rbac-service:8000/api/rbac/v1")
 	v.SetDefault("clients.rbac_timeout", 30)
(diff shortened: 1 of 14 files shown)


jlsherrill:
note this is currently built ontop of https://github.com/content-services/content-sources-backend/pull/637

will rebase once merged.

jlsherrill:
https://issues.redhat.com/browse/HMS-2422

pkg/dao/metrics.go
```diff
@@ -104,3 +105,40 @@ func (d metricsDaoImpl) PublicRepositoriesFailedIntrospectionCount(ctx context.C
...
+	}
+
+	return output
+}
+
+func (d metricsDaoImpl) PendingTasksOldestTask(ctx context.Context) float64 {
+	var task models.TaskInfo
+	res := d.db.WithContext(ctx).
+		Model(&models.TaskInfo{}).
+		Where("status = ?", config.TaskStatusPending).Order("queued_at ASC").First(&task)
```
rverdile, line 136: Would we want to alert on this one? IIRC, when a failed task is requeued, the queued_at doesn't reset, so this might cause some false positives. It might make sense to reset the queued_at time.

pkg/tasks/queue/pgqueue.go
```diff
@@ -60,7 +60,7 @@ const (
 
 	sqlRequeue = `
 		UPDATE tasks
-		SET started_at = NULL, token = NULL, status = 'pending', retries = retries + 1
+		SET started_at = NULL, token = NULL, status = 'pending', retries = retries + 1, queued_at = statement_timestamp()
```
rverdile, line 63: I would do the same thing for the `sqlRequeueFailedTasks` query

jlsherrill:
added a test to check for queued_at updates.

Also i realized that the 'number of days' until the cdn cert expires was only calculated at startup, so i changed it to be calculated as part of the go routine that updates the metrics.  At startup it was set to zero, so to keep that from firing an alert, i set it to calculate that one metric at startup (and only that metric, because others require the db to be up, which may not actually have happened yet?)

pkg/config/config.go
```diff
@@ -181,6 +181,9 @@ type Metrics struct {
 	// Defines the metrics port that the app should be configured to listen on for
 	// metric traffic.
 	Port int `mapstructure:"port"`
+
+	// How often (in seconds) to run queries to collect some metrics
+	CollectionFrequency int `mapstructure:"collection_frequency"`
```
rverdile, line 186: this should also be in config.yaml.example

rverdile (commented):
tested and the numbers look right! just the one small comment

pkg/config/config.go
jlsherrill: yes!

jlsherrill:
added

rverdile (approved):

PR #647, opened 2024-04-24
Fixes 3793: add make command for mockery

xbhouse:
## Summary

Adds make command to install mockery and regenerate mocks

## Testing steps

- Run `make mock`. This should install the latest version of mockery if it isn't already installed in the release directory and regenerate the mocks we have

## Checklist

### DIFF
--- a/mk/mockery.mk
+++ b/mk/mockery.mk
@@ -1,0 +1,12 @@
+MOCKERY_VERSION := 2.36.1
+
+export GO_OUTPUT
+
+.PHONY: install-mockery
+install-mockery: ## Install mockery locally on your GO_OUTPUT (./release) directory
+	mkdir -p $(GO_OUTPUT) && \
+	curl -sSfL https://github.com/vektra/mockery/releases/download/v$(MOCKERY_VERSION)/mockery_$(MOCKERY_VERSION)_$(shell uname -s)_$(shell uname -m).tar.gz | tar -xz -C $(GO_OUTPUT) mockery
+
+.PHONY: mock
+mock: ## Regenerate mocks
+	go generate ./...
--- a/mk/mockery.mk
+++ b/mk/mockery.mk
@@ -2,11 +2,10 @@ MOCKERY_VERSION := 2.36.1
 
 export GO_OUTPUT
 
-.PHONY: install-mockery
-install-mockery: ## Install mockery locally on your GO_OUTPUT (./release) directory
+$(GO_OUTPUT)/mockery: ## Install mockery locally on your GO_OUTPUT (./release) directory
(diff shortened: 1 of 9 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-3793

mk/mockery.mk
```diff
@@ -0,0 +1,12 @@
...
+
+export GO_OUTPUT
+
+.PHONY: install-mockery
+install-mockery: ## Install mockery locally on your GO_OUTPUT (./release) directory
+	mkdir -p $(GO_OUTPUT) && \
+	curl -sSfL https://github.com/vektra/mockery/releases/download/v$(MOCKERY_VERSION)/mockery_$(MOCKERY_VERSION)_$(shell uname -s)_$(shell uname -m).tar.gz | tar -xz -C $(GO_OUTPUT) mockery
+
+.PHONY: mock
+mock: ## Regenerate mocks
```
jlsherrill, line 11: if you drop the .PHONY, and name the make target the name of the resulting binary:
```
$(GO_OUTPUT)/mockery: 
	mkdir -p $(GO_OUTPUT) && \
	curl -sSfL https://github.com/vektra/mockery/releases/download/v$(MOCKERY_VERSION)/mockery_$(MOCKERY_VERSION)_$(shell uname -s)_$(shell uname -m).tar.gz | tar -xz -C $(GO_OUTPUT) mockery
```

Then you can set it to be a dependency on 'mock':
```
.PHONY: mock
mock:  $(GO_OUTPUT)/mockery
```

now when you just run 'make mock', it will install it if needed (and skip the install step if not).

xbhouse: oooh thanks, this is great! will update

mk/mockery.mk
```diff
@@ -0,0 +1,11 @@
+MOCKERY_VERSION := 2.36.1
```
jlsherrill, line 1: I think this is okay as is, but we will likely forget to update this as time goes on.  you could capture that with:

curl -L https://api.github.com/repos/vektra/mockery/releases/latest | jq --raw-output .tag_name

to get something like  v2.36.1
and then you'd just need to trim off the 'v' for the URL below.  Thoughts

xbhouse: i think that would be a good addition. so then i'd need to add a check to determine which version is installed and update the target if it isn't the latest?

jlsherrill: nah, i think you could just rely on a 'make clean'  which should clear it out

jlsherrill: meaning, fresh setups would get the newest, and people could easily upgrade as needed

jlsherrill (approved):

PR #650, opened 2024-04-26
Refs 4035: set QE_ENV="ephemeral"

swadeley:
## Summary

Recent errors in pr_checks (which use clowder venv):
`HTTP response body: Bad Request: x-rh-identity header has an invalid or missing org_id`

possibly related to authentication changes, e.g. certs required, but clowder and ephemeral users are not configured for cert auth, they are user name and password only.

Lets try if ephemeral's use of caddy containers solves that problem.

## Testing steps

pr check runs OK

## Checklist


swadeley:
https://issues.redhat.com/browse/HMS-4035

jlsherrill (approved):

PR #656, opened 2024-05-01
Fixes 4055: treat filter/page param parse as info

jlsherrill:
## Summary

Changes filter/pagination parsing errors as info level logging instead of error

## Testing steps

perform a request:
GET http://localhost:8000/api/content-sources/v1.0/templates/?limit=a

watch your server, it should show logging at INFO

## Checklist

### DIFF
--- a/pkg/dao/templates.go
+++ b/pkg/dao/templates.go
@@ -245,23 +245,19 @@ func (t templateDaoImpl) List(ctx context.Context, orgID string, paginationData 
 	order := convertSortByToSQL(paginationData.SortBy, sortMap, "name asc")
 
 	// Get count
-	filteredDB.
+	if filteredDB.
 		Model(&templates).
-		Count(&totalTemplates)
-
-	if filteredDB.Error != nil {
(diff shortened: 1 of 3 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-4055

xbhouse:
i see `11:40AM ERR ../../go/pkg/mod/github.com/content-services/lecho/v3@v3.5.2/middleware.go:179 > error="error: code=500, title=Error listing templates, detail=ERROR: invalid byte sequence for encoding \"UTF8\": 0x00 (SQLSTATE 22021) \n"` when sending a request like `/templates/?name=%00`

do you need to make these updates to the templates handler as well? and maybe the task_info handler, though invalid syntax (`/tasks/?status=%00`) logs as DBG there so i'm not sure

jlsherrill:
Good catch!  The template actually had a function to detect this, but 1) it wasn't being called & 2) it wasn't checking for a query error properly, this should be fixed!

xbhouse (approved):
lgtm!

PR #657, opened 2024-05-01
Build: add make repos-import-rhel9

jlsherrill:
## Summary

Support importing only a subset of repos for testing rhel9 (for example), and a handy make target.

## Testing steps

on a fresh db:  
```
make repos-import-rhel9
```

then check the red hat repos list and see only these two red hat repos

## Checklist

### DIFF
--- a/mk/repos.mk
+++ b/mk/repos.mk
@@ -8,5 +8,9 @@ repos-download: $(GO_OUTPUT)/external-repos  ## Download external repo urls from
 	; }
 
 .PHONY: repos-import
-repos-import: ## Import External repo urls from Image Builders into the DB.  Generates pkg/external_repos/external_repos.json
+repos-import: ## Import External repo urls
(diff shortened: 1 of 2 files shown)


mk/repos.mk
rverdile, line 1: is still imports everything in external_repos.json. is that okay?

swadeley:
Hi

I deployed with bonfire and:
 `--set-parameter content-sources-backend/OPTIONS_REPOSITORY_IMPORT_FILTER=rhel9`

and under the Red Hat tab I can see the baseos and appstream rhel9 repos are present and no others.

Thank you

mk/repos.mk
jlsherrill: yeah, this filter hasn't ever actually affected the public repos, i may do that as a separate pr

swadeley:
@rverdile merge on ACK please

rverdile (approved):

PR #661, opened 2024-05-06
Refs 4043: use content-template rbac

jlsherrill:
## Summary

Adds content template permission to content template apis

## Testing steps
Probably needs to be tested in stage

## Checklist


jlsherrill:
https://issues.redhat.com/browse/HMS-4043

pkg/test/mocks/rbac/rbac.go
```diff
@@ -154,6 +154,12 @@ func MockRbac(c echo.Context) error {
 			{
 				Permission: "content-sources:repositories:write",
 			},
+			{
+				Permission: "content-sources:content_templates:read",
```
xbhouse, line 158: do these need to match the ResourceTemplates string in the permission map exactly? if so, i think this would need to be `content_sources:templates:read` etc. if not, ignore this :)

pkg/rbac/permission_map.go
```diff
@@ -28,6 +28,7 @@ type Resource string
 
 const (
 	ResourceRepositories Resource = "repositories"
+	ResourceTemplates             = "templates"
```
xbhouse, line 31: does permission_map_test.go need to be updated as well?

jlsherrill: yes it does!

pkg/test/mocks/rbac/rbac.go
jlsherrill: they do thanks, i changed them and forgot to update here, good catch

xbhouse:
this looks good, just needs some updated tests (unless i missed them)

jlsherrill:
added a small test, but we don't really test rbac for each handler, we more test the infrastructure around it.  Probably better suited for an api integration suite or iqe tests

xbhouse (approved):

PR #670, opened 2024-05-15
Fixes 4093: Override candlepin environment paths

jlsherrill:
for custom repos

## Summary

Adds overriding of custom content baseurls 

## Testing steps

Follow the directions here: https://github.com/content-services/content-sources-backend/blob/main/docs/register_client.md

At the end you should have a client configured to use a  custom repository and a red hat repository.  You should now be able to 'dnf install' a package from the custom repo

## Checklist

### DIFF
--- a/pkg/candlepin_client/environment.go
+++ b/pkg/candlepin_client/environment.go
@@ -2,7 +2,6 @@ package candlepin_client
 
 import (
 	"context"
-	"fmt"
 	"strings"
 
 	caliri "github.com/content-services/caliri/release/v4"
@@ -109,21 +108,10 @@ func (c *cpClientImpl) DemoteContentFromEnvironment(ctx context.Context, envID s
 	return nil
 }
 
-func (c *cpClientImpl) OverrideContentPaths(ctx context.Context, environmentId string, contentLabelToUrl map[string]string) error {
+func (c *cpClientImpl) UpdateContentOverrides(ctx context.Context, environmentId string, dtos []caliri.ContentOverrideDTO) error {
 	ctx, client, err := getCandlepinClient(ctx)
 	if err != nil {
 		return err
-	}
-
-	dtos := []caliri.ContentOverrideDTO{}
-	for k, v := range contentLabelToUrl {
-		label := k
-		url := v
-		dtos = append(dtos, caliri.ContentOverrideDTO{
-			Name:         pointy.Pointer("baseurl"),
-			ContentLabel: &label,
-			Value:        &url,
-		})
 	}
 
 	_, httpResp, err := client.EnvironmentAPI.PutEnvironmentContentOverrides(ctx, environmentId).ContentOverrideDTO(dtos).Execute()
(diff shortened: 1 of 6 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-4093

test/integration/update_template_content_test.go
```diff
@@ -96,20 +99,23 @@ func (s *UpdateTemplateContentSuite) TestCreateCandlepinContent() {
 
 	// Create initial template
 	reqTemplate := api.TemplateRequest{
-		Name:            pointy.Pointer("test template"),
+		Name:            pointy.Pointer(fmt.Sprintf("test template %v", rand.Int())),
```
jlsherrill, line 102: I randomized the name here, otherwise i couldn't actually re-run the test without clearing the db

pkg/candlepin_client/environment.go
```diff
@@ -107,3 +108,46 @@ func (c *cpClientImpl) DemoteContentFromEnvironment(ctx context.Context, envID s
...
+			Value:        &url,
+		})
+	}
+
+	_, httpResp, err := client.EnvironmentAPI.PutEnvironmentContentOverrides(ctx, environmentId).ContentOverrideDTO(dtos).Execute()
+	if httpResp != nil {
+		defer httpResp.Body.Close()
+	}
+	if err != nil {
+		return fmt.Errorf("could not override contents urls %w", err)
```
rverdile, line 134: this is missing the "errorWithResponseBody()" pattern

pkg/candlepin_client/environment.go
```diff
@@ -107,3 +108,46 @@ func (c *cpClientImpl) DemoteContentFromEnvironment(ctx context.Context, envID s
...
+	if err != nil {
+		return []caliri.ContentOverrideDTO{}, err
+	}
+
+	overrides, httpResp, err := client.EnvironmentAPI.GetEnvironmentContentOverrides(ctx, environmentId).Execute()
+	if httpResp != nil {
+		defer httpResp.Body.Close()
+	}
+	if err != nil {
+		return []caliri.ContentOverrideDTO{}, fmt.Errorf("could not fetch env overrides %w", err)
```
rverdile, line 150: also missing the "errorWithResponseBody()" pattern

rverdile (commented):
should this also delete content overrides if a repository is removed from a template?

pkg/candlepin_client/environment.go
jlsherrill: good catch!

jlsherrill:
I'm not sure it fully matters, as its ignored if the content isn't in the environment.  It should be deleted when the content is deleted.... which we need to do on repo deletion, i'll file a task.  If you feel like we should be removing them, i can add them

pkg/tasks/update_template_content.go
```diff
@@ -508,6 +528,75 @@ func (t *UpdateTemplateContent) getContentList() ([]caliri.ContentDTO, []string,
...
 }
 
+// getOverrideDTOs uses the RepoConfigUUIDs to query the db and generate a mapping of content labels to distribution URLs
+// for the snapshot within the template.  For all repos, we include an override for an 'empty' sslcacert, so it does not use the configured default
+// on the client.  For custom repos, we override the base URL, due to the fact that we use different domains for RH and custom repos.
+func (t *UpdateTemplateContent) getOverrideDTOs(contentPath string) ([]caliri.ContentOverrideDTO, error) {
+	mapping := []caliri.ContentOverrideDTO{}
+	uuids := strings.Join(t.payload.RepoConfigUUIDs, ",")
+	origins := strings.Join([]string{config.OriginExternal, config.OriginRedHat}, ",")
+	customRepos, _, err := t.daoReg.RepositoryConfig.List(t.ctx, t.orgId, api.PaginationData{Limit: -1}, api.FilterData{UUID: uuids, Origin: origins})
```
rverdile, line 538: this is redhat and custom, not just custom, right?

rverdile (commented):
I created a template with RHEL 9 baseos and appstream, but I only see a content override for the cert of appstream. Is that a bug?

```
[
  {
    "created": "2024-05-21T17:30:42+0000",
    "updated": "2024-05-21T17:30:42+0000",
    "name": "sslcacert",
    "contentLabel": "rhel-9-for-x86_64-appstream-rpms",
    "value": " "
  }
]
```

pkg/tasks/update_template_content.go
jlsherrill: correct, we have to override the sslcacert for both

rverdile (approved):

PR #685, opened 2024-06-03
Fixes 4195: check for multiple tags in openapi spec

xbhouse:
## Summary

Adds a linter to check for multiple tags in the openapi spec when running `make openapi`. Each tag is associated with the resulting data. For example, for listing repository package groups, the tag would be `packagegroups`.

## Testing steps

- If there are any cases of multiple tags in the openapi spec, `make openapi` should fail and tell you which endpoint has more than one tag
- `make openapi` should not fail if each endpoint has only one tag
- For QE, some endpoints will be moved around (for example, if an endpoint was originally found under the `repositories_api` and it lists repository rpms, it would now only be found under the `rpms_api`) 

## Checklist

### DIFF
--- a/api/docs.go
+++ b/api/docs.go
@@ -30,7 +30,6 @@ const docTemplate = `{
                     "application/json"
                 ],
                 "tags": [
-                    "repositories",
                     "environments"
                 ],
                 "summary": "Search environments",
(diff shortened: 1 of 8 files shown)


xbhouse:
i can't see what's causing the snyk failure here, how can i see/fix that?

jlsherrill:
https://issues.redhat.com/browse/HMS-4195

jlsherrill (approved):

PR #687, opened 2024-06-04
Fixes 4249: Reuse HTTP zest transport

lzap:
Hello, I noticed this small detail - a new transport is created for each zest client, this effectively means that no HTTP connection pooling can be done. Here is an example of the [DefaultTransport](https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/net/http/transport.go) from the standard library that is supposed to be reused - it is thread safe.

Easy help, this patch just extracts the variable into the package scope so it can be reused on every new client. In case you are wondering, all those timeouts are coming from the `DefaultTransport` from the standard library plus the `ResponseHeaderTimeout` that was already defined here.

_Edit: I was trying out github.dev, did not go well, force pushed the correct version._

### DIFF
--- a/pkg/pulp_client/client.go
+++ b/pkg/pulp_client/client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-4249

jlsherrill:
looks good!  i think you're just missing the 'net' import

lzap:
Sure, rebased.

jlsherrill (approved):

jlsherrill:
thanks @lzap !

PR #690, opened 2024-06-05
Fixes 3974: add priority to tasks

rverdile:
## Summary
Adds the priority column to tasks, an integer representing the priority of the task. Tasks with a higher integer will be dequeued first. Ready tasks view is sorted by priority then queued_at time i.e. highest priority, then oldest, goes first.

Make the priority of update_template_content tasks 1, so it is always dequeued first.

## Testing steps
1. Run the db migration
2. Change worker_count in your config.yaml to 1
3. Create two EPEL repos, one after the other.
4. While the first one is snapshotting, create or update a template.
5. The order of the task completion should be: snapshot (for the first EPEL repo), update_template_content, snapshot (for the second EPEL repo).

## Checklist


jlsherrill:
https://issues.redhat.com/browse/HMS-3974

xbhouse (approved):
looks good to me!

PR #704, opened 2024-06-12
Build: Update to go 1.22

jlsherrill:
## Checklist


rverdile:
is this a draft on purpose? also, should we update yummy and tang as well?

jlsherrill:
yes, i can bump those after this goes in (would rather do it in that order, as its backwards compatible but not forward compatible).

PR #719, opened 2024-06-25
Fixes 3861: reject cancellation of certain task types

xbhouse:
## Summary

- Adds a list of cancellable tasks to check against and rejects cancelling any tasks not included in the list. Currently includes only snapshot and introspect tasks
- Unrelated but attempting to cancel a RH task resulted in a panic due to dereferencing an error that doesn't exist so this includes that fix also

## Testing steps

- Start a few different tasks in addition to introspect and snapshot and take note of the task uuid. List of all task types [here](https://github.com/content-services/content-sources-backend/blob/main/pkg/config/tasks.go)
- Make a request to cancel any task that isn't snapshot or introspect: `POST /tasks/:uuid/cancel/`. You should get a 400
- Cancelling snapshot or introspect tasks works as before

## Checklist

### DIFF
--- a/pkg/handler/task_info.go
+++ b/pkg/handler/task_info.go
@@ -9,6 +9,7 @@ import (
 	ce "github.com/content-services/content-sources-backend/pkg/errors"
 	"github.com/content-services/content-sources-backend/pkg/rbac"
 	"github.com/content-services/content-sources-backend/pkg/tasks/client"
+	"github.com/content-services/content-sources-backend/pkg/tasks/queue"
 	"github.com/labstack/echo/v4"
 	"github.com/rs/zerolog/log"
 )
@@ -104,11 +105,15 @@ func (t *TaskInfoHandler) cancel(c echo.Context) error {
 		return ce.NewErrorResponse(ce.HttpCodeForDaoError(err), "error canceling task", err.Error())
 	}
 	if task.OrgId == config.RedHatOrg {
-		return ce.NewErrorResponse(http.StatusBadRequest, "Cannot cancel a Red Hat Task", err.Error())
+		return ce.NewErrorResponse(http.StatusBadRequest, "error canceling task", "Cannot cancel a Red Hat Task")
 	}
 	err = t.TaskClient.SendCancelNotification(c.Request().Context(), id)
 	if err != nil {
-		return ce.NewErrorResponse(http.StatusInternalServerError, "error canceling task", err.Error())
+		if err == queue.ErrNotCancellable {
+			return ce.NewErrorResponse(http.StatusBadRequest, "error canceling task", err.Error())
+		} else {
+			return ce.NewErrorResponse(http.StatusInternalServerError, "error canceling task", err.Error())
+		}
 	}
 	return c.NoContent(http.StatusNoContent)
 }
--- a/pkg/tasks/client/client.go
+++ b/pkg/tasks/client/client.go
@@ -3,9 +3,11 @@ package client
 import (
 	"context"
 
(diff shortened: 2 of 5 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-3861

pkg/handler/task_info.go
```diff
@@ -104,7 +105,10 @@ func (t *TaskInfoHandler) cancel(c echo.Context) error {
 		return ce.NewErrorResponse(ce.HttpCodeForDaoError(err), "error canceling task", err.Error())
 	}
 	if task.OrgId == config.RedHatOrg {
-		return ce.NewErrorResponse(http.StatusBadRequest, "Cannot cancel a Red Hat Task", err.Error())
+		return ce.NewErrorResponse(http.StatusBadRequest, "error canceling task", "Cannot cancel a Red Hat Task")
+	}
+	if !slices.Contains(config.CancellableTasks, task.Typename) {
+		return ce.NewErrorResponse(http.StatusBadRequest, "error canceling task", "Can only cancel introspect or snapshot tasks")
```
rverdile, line 111: What do you think about moving this check into this method?
https://github.com/content-services/content-sources-backend/blob/main/pkg/tasks/client/client.go#L38

This way if a task is cancelled outside of the handler, the cancel is still blocked

xbhouse: makes sense! will do

pkg/handler/task_info.go
```diff
@@ -104,11 +104,11 @@ func (t *TaskInfoHandler) cancel(c echo.Context) error {
 		return ce.NewErrorResponse(ce.HttpCodeForDaoError(err), "error canceling task", err.Error())
 	}
 	if task.OrgId == config.RedHatOrg {
-		return ce.NewErrorResponse(http.StatusBadRequest, "Cannot cancel a Red Hat Task", err.Error())
+		return ce.NewErrorResponse(http.StatusBadRequest, "error canceling task", "Cannot cancel a Red Hat Task")
 	}
 	err = t.TaskClient.SendCancelNotification(c.Request().Context(), id)
 	if err != nil {
-		return ce.NewErrorResponse(http.StatusInternalServerError, "error canceling task", err.Error())
+		return err
```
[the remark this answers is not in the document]
rverdile: I would add a new error [here](https://github.com/content-services/content-sources-backend/blob/main/pkg/tasks/queue/queue.go#L55-L62) for "task not cancellable" or something like that

Then you can return that error from SendCancelNotification() and check for it in the handler before deciding the response code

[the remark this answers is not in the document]
xbhouse: thank you!

rverdile (commented):
code looks good!! 

could you add a test to pgqueue_test.go to check that SendCancelNotification() errors for an un-cancelable type?

swadeley:
go lint fails

pkg/tasks/client/client.go
```diff
@@ -3,9 +3,11 @@ package client
 import (
 	"context"
 
+	"github.com/content-services/content-sources-backend/pkg/config"
 	"github.com/content-services/content-sources-backend/pkg/tasks"
 	"github.com/content-services/content-sources-backend/pkg/tasks/queue"
 	"github.com/google/uuid"
+	"golang.org/x/exp/slices"
```
rverdile, line 10: one small thing and this is good to go! this import should be "slices" since it's now included in the standard library

rverdile (approved):
nice job!

PR #721, opened 2024-06-26
Fixes 4334: lower parse filters log level

rverdile:
## Summary
Changes all the parse filter error returns to log at the same level (INFO)

## Testing steps
1. Make a request to `tasks/?exclude_red_hat_org=notabool`, where the params are the wrong type
2. In the server logs, there will be an error logged at the INFO level
## Checklist


jlsherrill:
https://issues.redhat.com/browse/HMS-4334

PR #738, opened 2024-07-10
Fixes 4420: add ca support to candlepin client

jlsherrill:
## Summary

support a custom ca cert in teh candlepin connection

## Testing steps

I would just do a code review on this.  I was able to see this work using custom certs, but it was a very manual process of loading files from the filesystem into the config (since the config takes the file contents, not file names).   We should see it fix the current issue in stage quickly

## Checklist

### DIFF
--- a/deployments/deployment.yaml
+++ b/deployments/deployment.yaml
@@ -150,6 +150,12 @@ objects:
                     name: content-sources-candlepin
                     key: key
                     optional: true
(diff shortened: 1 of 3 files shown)


Andrewgdewar (approved):

jlsherrill:
https://issues.redhat.com/browse/HMS-4420

PR #755, opened 2024-07-24
Fixes 4446: improve task context canceled logging

rverdile:
## Summary
We are seeing what's likely expected error logs from tasks that are cancelled. We want to avoid logging these errors because they are expected when a task is cancelled and can be ignored.

## Testing steps
1. `make repos-import`
2. `go run cmd/external-repos/main.go nightly-jobs`. Turn off snapshotting to save time.
3. Keep exiting the server with CTRL+C and restarting it
4. You shouldn't see error logs, expect maybe `error requeuing task with task_id: 00000000-0000-0000-0000-000000000000`. Ignore that one.

## Checklist

### DIFF
--- a/cmd/content-sources/main.go
+++ b/cmd/content-sources/main.go
@@ -61,12 +61,12 @@ func main() {
 
 	// Setup cancellation context
 	var wg sync.WaitGroup
-	ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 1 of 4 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-4446

PR #762, opened 2024-08-05
Test the test

jlsherrill:


cmd/candlepin/main.go.orig
```diff
@@ -0,0 +1,77 @@
+package main
```
xbhouse, line 1: did you mean to add this file?

jlsherrill: doh! no good catch ahha

xbhouse (approved):

PR #764, opened 2024-08-06
Fixes 4539: fix indentation in example yaml config

dominikvagner:
## Summary
There is a wrong indentation in the `configs/config.yaml.example` file which when copied to remove `.example` extension doesn't work and causes `make compose-up` fail. 

## Testing steps
Fixed indentation and it works now after copying the file.


jlsherrill:
https://issues.redhat.com/browse/HMS-4539

xbhouse (approved):
nice!!

PR #769, opened 2024-08-08
Fixes 4422: edited repo returns incorrect snapshot task

xbhouse:
## Summary

- Adds a clause to a few queries to check only repo configs that have not been soft-deleted

## Testing steps

- Add a repository, let it snapshot, and grab the snapshot task uuid
- Restart the app with pulp disabled to force the delete-repository-snapshots task to fail
- Delete the repository
- Fetch the task associated with that repository, you should get a 404
- Listing tasks filtered on that repository uuid should return an empty list
- Searching environments, package groups, and rpms should work as before
- For QE, this should fix the issue that caused `test_repo_snapshot_functions` to fail, so we should make sure this test passes


jlsherrill:
https://issues.redhat.com/browse/HMS-4422

swadeley:
Hi, API version of `test_repo_snapshot_functions` now passes, I'll revert the changes that disabled the test.

rverdile (approved):

PR #773, opened 2024-08-14
Fixes 4432: log request body for server errors

dominikvagner:
## Summary
Added a middleware which stores the body of incoming requests (cutoff to the limit of 1000B) and logs the request body when the response error code is of the 500 level (i.e. Server Error).

Wrote automated tests for checking the body storing it's recovery, cutoff limit and logging. Also tested this manually and checked for `request-id` retention and logging.

### 2. Comment out the body of it and put this line in the handler.
```
return ce.NewErrorResponse(http.StatusInternalServerError, "Test Error", "5xx testing error")
```
### 3. Start the server and send a request to it.
example request:
```
http http://localhost:8000/api/content-sources/v1.0/rpms/names "$( ./scripts/header.sh acme 1111)" limit:=100 uuids:='["cdb499cd-fc3b-44f9-9d4b-81824a418090"]' search='YfbrYLBvVcnY' x-Rh-Insights-Request-Id:9d229d34-7219-405d-ba3d-8f99cf7c36ae -v
```
### 4. Look at logs for an error message with the `request body` and `request-id` printed out.
image

### DIFF
--- a/pkg/middleware/log_server_error_request.go
+++ b/pkg/middleware/log_server_error_request.go
@@ -1,0 +1,62 @@
+package middleware
+
+import (
+	"bytes"
+	"errors"
+	"io"
+	"net/http"
+
+	ce "github.com/content-services/content-sources-backend/pkg/errors"
+	"github.com/labstack/echo/v4"
+)
+
+const BodyDumpLimit = 1000
+const BodyStoreKey = "body_backup"
+
+func LogServerErrorRequest(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) (err error) {
+		if c.Get(BodyStoreKey) == nil {
+			storeRequestBody(c)
+		}
+		if err = next(c); err != nil {
+			if containsServerError(err) {
+				logRequestBody(c)
+			}
+			return err
+		}
+		return nil
+	}
+}
+
+func containsServerError(err error) bool {
+	httpError := new(ce.ErrorResponse)
+	if errors.As(err, httpError) {
+		for _, e := range httpError.Errors {
+			if e.Status >= http.StatusInternalServerError {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func logRequestBody(c echo.Context) {
+	if body := c.Get(BodyStoreKey); body != nil {
+		storedBodyBytes, ok := body.([]byte)
+		if !ok {
+			c.Logger().Error("Error reading request body")
+		}
+		c.Logger().Errorf("Request body: %v", string(storedBodyBytes))
+	}
+}
+
+func storeRequestBody(c echo.Context) {
+	var reqBody []byte
+	if c.Request().Body != nil {
+		reqBody, _ = io.ReadAll(c.Request().Body)
+	}
+	c.Request().Body = io.NopCloser(bytes.NewBuffer(reqBody))
+
+	limit := min(len(reqBody), BodyDumpLimit)
+	c.Set(BodyStoreKey, reqBody[:limit])
+}
--- a/pkg/middleware/log_server_error_request.go
+++ b/pkg/middleware/log_server_error_request.go
@@ -54,7 +54,7 @@ func storeRequestBody(c echo.Context) {
 	limit := BodyDumpLimit
 	buffered := BufferedReadCloser{bufio.NewReader(c.Request().Body), c.Request().Body}
 
-	bytes, err := buffered.Peek(1000)
+	bytes, err := buffered.Peek(BodyDumpLimit)
 	if errors.Is(err, io.EOF) {
 		limit = len(bytes)
 		err = nil
@@ -73,9 +73,9 @@ func storeRequestBody(c echo.Context) {
 
 func isBodiedMethod(method string) bool {
 	switch method {
-	case "GET":
+	case http.MethodGet:
 		return false
-	case "DELETE":
+	case http.MethodDelete:
 		return false
 	default:
 		return true
--- a/pkg/middleware/log_server_error_request.go
+++ b/pkg/middleware/log_server_error_request.go
@@ -1,7 +1,7 @@
 package middleware
 
 import (
-	"bytes"
+	"bufio"
 	"errors"
 	"io"
 	"net/http"
@@ -15,11 +15,11 @@ const BodyStoreKey = "body_backup"
 
 func LogServerErrorRequest(next echo.HandlerFunc) echo.HandlerFunc {
 	return func(c echo.Context) (err error) {
-		if c.Get(BodyStoreKey) == nil {
+		if c.Get(BodyStoreKey) == nil && isBodiedMethod(c.Request().Method) {
 			storeRequestBody(c)
 		}
 		if err = next(c); err != nil {
-			if containsServerError(err) {
+			if containsServerError(err) && isBodiedMethod(c.Request().Method) {
 				logRequestBody(c)
 			}
 			return err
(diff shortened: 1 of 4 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-4432

pkg/middleware/log_server_error_request_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+func TestReadBodyStoreAndRestore(t *testing.T) {
+	e := echo.New()
+	req := httptest.NewRequest(http.MethodGet, "/", bytes.NewBufferString(TestBody))
```
xbhouse, line 22: small thing, since GET requests don't usually have a body, maybe we should check for both GET and POST requests (or just POST) in the tests? it could be good to test what's logged for requests with no body too. right now for requests with no body i see this: 
`11:34AM ERR ../../go/pkg/mod/github.com/content-services/lecho/v3@v3.5.2/logger.go:88 > Request body:  request_id=a80aff68-f533-4677-a628-b245295ac4b0`

xbhouse:
overall i think this looks great and works well! @jlsherrill can you take a look in case i missed something?

pkg/middleware/log_server_error_request_test.go
dominikvagner: oh, must have missed that, for clarity in the tests it probably should be a `POST` or a method usually having a body (not that it would change the outcome/functionality in the case of these tests 😅)
that gets me thinking, if I should add a check for the request method and if it's `GET` then don't log this message at all 🤔

pkg/middleware/log_server_error_request.go
```diff
@@ -0,0 +1,62 @@
...
+
+func storeRequestBody(c echo.Context) {
+	var reqBody []byte
+	if c.Request().Body != nil {
+		reqBody, _ = io.ReadAll(c.Request().Body)
+	}
+	c.Request().Body = io.NopCloser(bytes.NewBuffer(reqBody))
+
+	limit := min(len(reqBody), BodyDumpLimit)
+	c.Set(BodyStoreKey, reqBody[:limit])
```
jlsherrill, line 61: I worry a bit about the performance hit of reading in the entire request body, especially for large requests coming in.

I played around a bit with bufio.Reader which implements a buffered Reader. This allows for 'peeking' at the first 'n' bytes without reading the entire Body. 

 This doesn't come with a Close() method though so you have to implement that:

```
type BufferedReadCloser struct {
	*bufio.Reader
	io.ReadCloser
}

func (rw BufferedReadWriteCloser) Close() error {
	return rw.ReadCloser.Close()
}

func (rw BufferedReadWriteCloser) Read(p []byte) (int, error) {
	return rw.Reader.Read(p)
}
```

Then you can create a buffered ReadCloser  from the request body:
```
		buffered := BufferedReadCloser{bufio.NewReader(c.Request().Body), c.Request().Body}
```

and 'peek' at the first 1000 bytes:

```
bytes, err := buffered.Peek(1000)
```

And then you could just set the buffered reader as the body:
```
c.Request().Body =  buffered
```

This way we aren't duplicating the entire body in memory, we're just buffering the first ~1000 bytes we read, at least until the request is actually processed.  
What do you think?

jlsherrill:
one comment, otherwise everything else looks great!

pkg/middleware/log_server_error_request.go
dominikvagner: True, this could be a more performant solution. 🤔 
And the `Peek()` method seems to **not** advance the reader so maybe I wouldn't have to restore the body on the request by writing the bytes read back to it. 😅
Will test it out. 👍🏼

dominikvagner: Did some profiling for this solution and it seems to be more memory efficient, but at the cost of CPU time. Although still an improvement I think, as the memory footprint might be a more important metric, so we should go with it 😄 
\
Methodology:
- 1,000,000 POST requests wrapped in the middleware each with body of ~12500 bytes.
- Code was an adjusted version of the `TestLargeBodyCutoff()` test case for this middleware.
- Profiled with the JetBrains (GoLand) built-in profiler.
- Averaged 3 runs.

Results:
|   | `ReadAll()` (Baseline)  | `Peek()` | Diff |
| -------------| ------------- | ------------- | ------------- |
| **Allocated Space**  | 97 GB | 74 GB | - 23.7 %  |
| **CPU Time** | 84 sec. | 104 sec. | + 23.8 % |

jlsherrill: interesting!!  Out of curiosity what does the cpu/memory look like without the middleware in use?

dominikvagner: found a mistake in my test case, had added some debug/check read and forgot it there, updated table here and it seems the buffered reader is a better solution all around 🆒😁 

updated results:
|                | W/O       | ReadAll() | Peek()    |
|----------------|-----------|-----------|-----------|
| Alocated Space | 32.8 GB   | 69.2 GB   | 37.2 GB   |
| CPU Time       | 52.3 sec. | 85.0 sec. | 68.3 sec. |

jlsherrill: wow! very cool, nice work getting real data!

dominikvagner: just to be sure, tested it in a more expected normal scenario, along with the regular more real like `echo` setup (ran the test for creating a repository 1M times with the regular payload, so not a crazy big payload that has to be read/binded after the middleware) 😄 
and if the reads after the middleware with the buffered reader wrapped request body, are any slower for those normal sized requests then it's not really noticeable, even at 1M requests, so I think we should be fine 😁

pkg/middleware/log_server_error_request_test.go
dominikvagner: [resolved in the latest commit](https://github.com/content-services/content-sources-backend/pull/773/commits/8a34ffb865d15367b1bdda71ef5420a7bcab6422), turned off logging for `GET` and `DELETE` methods as they normally shouldn't have a body

pkg/middleware/log_server_error_request.go
dominikvagner: [resolved in the latest commit](https://github.com/content-services/content-sources-backend/pull/773/commits/8a34ffb865d15367b1bdda71ef5420a7bcab6422), implemented the new buffered reader after finding that it provides considerable performance gains, good catch/proposal 👍🏼😄

jlsherrill (dismissed):
Nice work!!

pkg/middleware/log_server_error_request.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		c.Logger().Errorf("Request body: %v", string(storedBodyBytes))
+	}
+}
+
+func storeRequestBody(c echo.Context) {
+	limit := BodyDumpLimit
+	buffered := BufferedReadCloser{bufio.NewReader(c.Request().Body), c.Request().Body}
+
+	bytes, err := buffered.Peek(1000)
```
jlsherrill, line 57: oh! you might wanna use the constant you defined above BodyDumpLimit

dominikvagner: oops, missed that, [resolved in the latest commit](https://github.com/content-services/content-sources-backend/pull/773/commits/9414a6f6a60bfc10e30a61f42aada445adbc9a41)

xbhouse (approved):
awesome work!!

PR #776, opened 2024-08-14
Fixes 4426: Remove NewRepositoryFiltering feature

Andrewgdewar:
## Summary

Associated ticket [here](https://issues.redhat.com/browse/HMS-4426). 

This removes the NewRepositoryFiltering feature option. 

## Testing steps
- Call the features API, the NewRepositoryFiltering variable should no longer appear.

### DIFF
--- a/compose_files/pulp/docker-compose.yml
+++ b/compose_files/pulp/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '3'
+version: "3"
 services:
   postgres:
     image: "docker.io/library/postgres:13"
@@ -8,8 +8,8 @@ services:
       POSTGRES_USER: pulp
       POSTGRES_PASSWORD: password
       POSTGRES_DB: pulp
-      POSTGRES_INITDB_ARGS: '--auth-host=scram-sha-256'
-      POSTGRES_HOST_AUTH_METHOD: 'scram-sha-256'
+      POSTGRES_INITDB_ARGS: "--auth-host=scram-sha-256"
+      POSTGRES_HOST_AUTH_METHOD: "scram-sha-256"
     volumes:
       - "pg_data:/var/lib/postgresql"
       - "./assets/postgres/passwd:/etc/passwd:Z"
@@ -22,6 +22,7 @@ services:
 
   migration_service:
     image: "quay.io/redhat-services-prod/pulp-services-tenant/pulp:latest"
+    platform: linux/amd64
     depends_on:
       postgres:
         condition: service_healthy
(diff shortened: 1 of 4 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-4426

pkg/dao/repository_configs.go
```diff
@@ -387,7 +387,8 @@ func (r repositoryConfigDaoImpl) filteredDbForList(OrgID string, filteredDB *gor
 		filteredDB = filteredDB.Where("name = ?", filterData.Name)
 	}
 
-	if !config.Get().Features.NewRepositoryFiltering.Enabled && filterData.ContentType == "" {
+	// Default type to ContentTypeRpm if not specified
+	if filterData.ContentType == "" {
 		filterData.ContentType = config.ContentTypeRpm
```
jlsherrill, line 392: I think we want to not default to RPM if its disabled.  The point of 'newrepositoryfiltering' was that there wasn't any filtering and a fetch of  /repositories/  returns everything

pkg/dao/repository_configs.go
```diff
@@ -422,9 +423,11 @@ func (r repositoryConfigDaoImpl) filteredDbForList(OrgID string, filteredDB *gor
 			Where("name ILIKE ? OR url ILIKE ?", containsSearch, containsSearch)
 	}
 
-	if !config.Get().Features.NewRepositoryFiltering.Enabled && filterData.Origin == "" {
+	// Default origin to "external" if not specified
+	if filterData.Origin == "" {
 		filterData.Origin = config.OriginExternal
```
jlsherrill, line 428: same comment here, we don't want to default to originExternal now, as we want the feature to be 'turned on' permanently

compose_files/pulp/docker-compose.yml
```diff
@@ -22,6 +22,7 @@ services:
 
   migration_service:
     image: "quay.io/redhat-services-prod/pulp-services-tenant/pulp:latest"
+    platform: linux/amd64
```
xbhouse, line 25: i don't think we want to push the changes here in case it messes up anyone else, i'm not sure if it would but just in case. maybe we can ask someone using fedora to test if this breaks anything and add this to another PR if not, it would be nice to have for us mac users :)

xbhouse:
i think a few test cases are failing in dao/snapshots_test.go because all the repo types (external, redhat, and upload) are being listed now, so you may need to adjust the values being asserted

swadeley:
Hi @Andrewgdewar 

When you have time to fix the unit tests, please rebase to pick up change in https://github.com/content-services/content-sources-backend/pull/788

Thank you

compose_files/pulp/docker-compose.yml
jlsherrill: i can test it

jlsherrill: works fine, ack on this change

xbhouse: awesome!! 🎉 thank you for testing!

xbhouse (approved):

PR #780, opened 2024-08-19
Fixes 4478: support cancel pending tasks

rverdile:
## Summary
Adds support for cancelling pending tasks. Anywhere in our code that was cancelling a running task will now cancel a running or pending task.

## Testing steps
1. Add all the EPEL repos, so their snapshot tasks will use all 3 workers. These snapshot tasks will be running.
2. Create another repo with snapshot. This snapshot task will be pending.
3. Delete all the repos you just created.
4. Both the running and pending snapshot tasks should be canceled.

### DIFF
--- a/pkg/candlepin_client/candlepin_client_mock.go
+++ b/pkg/candlepin_client/candlepin_client_mock.go
@@ -6,6 +6,7 @@ import (
 	context "context"
 
 	caliri "github.com/content-services/caliri/release/v4"
+
 	mock "github.com/stretchr/testify/mock"
 )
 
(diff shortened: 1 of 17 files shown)


xbhouse:
i see some logs like this after adding 4 repos and deleting them all while pending or running:
`DBG CancelTask: Status Conflict`
`WRN error canceling task: error deleting heartbeat: heartbeat not found. was this task requeued recently?`

it does seem like all the tasks do get cancelled though 🤔

xbhouse:
code looks good to me! just needs a rebase

rverdile:
those logs are okay. they're just warnings and in this case they don't indicate that something went wrong

xbhouse (approved):

swadeley:
Hi @rverdile 

I see some `update-latest-snapshot` tasks pending.

rverdile:
thanks @swadeley, that should be fixed now

swadeley:
Hi

all tasks cancelled as expected.

Thank you.

PR #782, opened 2024-08-20
Fixes 4541: fix template patch endpoint error on empty date

dominikvagner:
## Summary
There is a problem when sending a `PATCH` to update a template and supplying only the `use_latest=true` and not the date (supplying a null or an empty string), that errors out right now, but it should zero out the date instead.
This PR fixes that, by changing the way the date is decoded/unmarshaled from the JSON request body by wrapping it in a custom type (it's necessary to it this way, so that the parsing doesn't quit after first binding error, can't just silence the error returned by `Bind()` as that would ignore any following data).
Accepting `date=""` as zeroed date was also propagated to accompanying template `PUT` and `POST` endpoints.

#### 1. Create a new template
```bash
http :8000/api/content-sources/v1.0/templates/ "$( ./scripts/header.sh acme 1111)" arch='x86_64' version='9' repository_uuids:='[]' description='testing patch date latest' name='test' use_latest:=false date="2024-07-20T11:15:17+02:00"
```

#### 2. Update `use_latest` to true with `PATCH` and _don't supply_ the date.
```bash
http PATCH :8000/api/content-sources/v1.0/templates/{_NEW TEMPLATE ID_} "$( ./scripts/header.sh acme 1111)" use_latest:=true
```
Should pass and zero out the date. (`"date": "0001-01-01T00:57:44+00:57"`)

#### 3. Revert the update
```bash
http PATCH :8000/api/content-sources/v1.0/templates/{_NEW TEMPLATE ID_} "$( ./scripts/header.sh acme 1111)" use_latest:=false date="2024-07-20T11:15:17+02:00"
```

#### 4. Update `use_latest` to true with `PATCH` and _supply_ the date as empty string.
```bash
http PATCH :8000/api/content-sources/v1.0/templates/{_NEW TEMPLATE ID_} "$( ./scripts/header.sh acme 1111)" use_latest:=true date="" name="renamed"
```
Should pass and zero out the date. (`"date": "0001-01-01T00:57:44+00:57"`) and the name should also change to "renamed" as the data following the date is also correctly parsed.

### DIFF
--- a/pkg/api/templates.go
+++ b/pkg/api/templates.go
@@ -106,7 +106,7 @@ func (d EmptiableDate) MarshalJSON() ([]byte, error) {
 
 func (d *EmptiableDate) UnmarshalJSON(b []byte) error {
 	str := string(b)
-	if b == nil || len(b) == 0 || str == "null" || str == `""` || str == "" {
+	if len(b) == 0 || str == "null" || str == `""` || str == "" {
 		*d = EmptiableDate(time.Time{})
 		return nil
 	}
--- a/pkg/api/templates.go
+++ b/pkg/api/templates.go
@@ -1,22 +1,23 @@
 package api
 
 import (
+	"encoding/json"
 	"time"
 
 	"github.com/content-services/content-sources-backend/pkg/utils"
 )
 
 type TemplateRequest struct {
-	UUID            *string    `json:"uuid" readonly:"true" swaggerignore:"true"`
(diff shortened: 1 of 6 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-4541

xbhouse (approved):
working well and looks good! nice job! :)

PR #783, opened 2024-08-22
Fixes 4549: add more tasks to task cleanup

rverdile:
## Summary
Adds more task types to the tasks that get cleaned up nightly. Now all tasks get cleaned up if they are older than 10 days. The deletion tasks only get cleaned if they completed.

## Testing steps
1. Create a repository and template.
2. Delete a repository and template.
3. This gives a few task types to work with. Update the tasks to change the finished_at times and verify they get cleaned up.

You can use this SQL query to update a task that will always get cleaned up
```
UPDATE tasks
SET queued_at = '2024-08-08 15:24:52.793340 +00:00', started_at = '2024-08-09 15:24:52.793340 +00:00', finished_at = '2024-08-10 15:24:52.793340 +00:00'
WHERE id = '<task id>';
```

You can use this SQL query to update a task that will get cleaned up only if it succeeded. The only difference here is the status is failed.
```
UPDATE tasks
SET queued_at = '2024-08-09 15:24:52.793340 +00:00', started_at = '2024-08-09 15:24:52.793340', finished_at = '2024-08-10 15:24:52.793340 +00:00', status = 'failed'
WHERE id = '<task id>';
```
4. run `go run cmd/external-repos/main.go nightly-jobs`
5. Verify the expected tasks are cleanedup


jlsherrill:
https://issues.redhat.com/browse/HMS-4549

pkg/dao/task_info.go
```diff
@@ -116,12 +116,12 @@ func (t taskInfoDaoImpl) List(
...
-	// Delete all completed or failed introspection tasks that are older than 10 days
-	// Delete all finished Repo delete tasks that are older 10 days
+	// Delete all completed or failed specified tasks that are older than 10 days
+	// Delete all finished delete tasks that are older 10 days
 	q := "delete from tasks where " +
-		"(type = '%v' and (status = 'completed' or status = 'failed') and finished_at < (current_date - interval '10' day)) OR" +
-		"(type = '%v' and status = 'completed' and  finished_at < (current_date - interval '10' day))"
-	q = fmt.Sprintf(q, config.IntrospectTask, config.DeleteRepositorySnapshotsTask)
+		"(type in (%v) and (status = 'completed' or status = 'failed') and finished_at < (current_date - interval '10' day)) OR" +
+		"(type in (%v) and status = 'completed' and  finished_at < (current_date - interval '10' day))"
```
jlsherrill, line 123: can we adjust this 10 days to 20 days?  10 was fine for something like introspection failures, but for other failures, i worry 10 isn't enough retention to be able to investigate problems adequately.

jlsherrill: sorry, not this one, the one above it, with 'or status = failed'

jlsherrill:
tested well and looks good, just one small comment.

pkg/dao/task_info.go
rverdile: updated!

jlsherrill (approved):

PR #789, opened 2024-08-27
Fixes 4681: Update to go 1.23

jlsherrill:
## Summary
you can upgrade with
```
go get go@1.23.0
```
## Testing steps
tests pass


swadeley:
Hi @jlsherrill 

You mentioned the bug in openapi that changes the order of these values:
```
"sha256",
 "file"
```
and I notice the only change to the api docs in this PR is that change you mentioned.

Does that mean I can just ignore this, you will you revert that change, or do I still need to regenerate API docs?

jlsherrill:
The ordering of  those items be non-deterministic, and possibly change anytime 'make openapi' is run until https://github.com/getkin/kin-openapi is merged.   I think we can just ignore this until that happens

jlsherrill:
but i guess i could change it back to the correct order :)

jlsherrill:
actually this seems like the more common order that is generated, so lets leave it like this until that new dependency is released

jlsherrill:
rebased and removed openapi change

swadeley:
Hi

pr check test run says:
The ENV_FOR_DYNACONF is ephemeral

yet this test ran (it is tagged for stage only):
```
    @pytest.mark.stage
    def test_search_non_added_popular_repo
```

I'll trigger a retest, see if it happens again (might have to specifically omit it).

jlsherrill:
https://issues.redhat.com/browse/HMS-4681

PR #790, opened 2024-08-28
Fixes 4523: use clock timestamp for started_at

jlsherrill:
## Summary

statement_timestamp() calculates the timestamp of when the statement starts.  I suspect that a new task is being added while this query is executing in rare cases (but don't actually have much evidence for this).  By using clock_timestamp(), we could (in theory) ensure that the started_at is always set after the queued_at is set.

We'll probably have to commit this and monitor errors for a couple weeks to see if its still occuring.

## Testing steps

tests pass


jlsherrill:
https://issues.redhat.com/browse/HMS-4523

rverdile (commented):
this sounds like it could be the cause to me. could you change the other uses of statement_timestamp() to clock_timestamp()? clock_timestamp() seems like the better function to use

jlsherrill:
[test]

rverdile (approved):

PR #792, opened 2024-08-28
Build: add optional live reloading server setup

dominikvagner:
## Summary
This is just a DX improvement for speeding up local development by adding an optional live reload server setup. I set this up for myself, but I think it would be a great addition to the project and is useful to everyone, try it out if you like it/find it useful we could add it to the repo. 😁😇

### DIFF
--- a/.air.toml
+++ b/.air.toml
@@ -1,0 +1,50 @@
+root = "."
+tmp_dir = "tmp"
+
+[build]
+  args_bin = ["api", "consumer", "instrumentation", "mock_rbac"]
+  bin = "./release/content-sources"
+  cmd = "make build"
+  delay = 1000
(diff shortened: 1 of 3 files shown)


.air.toml
```diff
@@ -0,0 +1,50 @@
+root = "."
+tmp_dir = "tmp"
+
+[build]
+  args_bin = ["api", "consumer", "instrumentation", "mock_rbac"]
+  bin = "./release/content-sources"
+  cmd = "make build"
```
jlsherrill, line 7: You might change this to 
```go build -o ./release/content-sources cmd/content-sources/main.go```

so then its only rebuilding the binary used by the server.  That will speed it up a lot!

dominikvagner: changed! thanks for the tip 👌🏼

jlsherrill (approved):

PR #794, opened 2024-08-30
Fixes 4525: add subscription check endpoint

rverdile:
## Summary
Adds an endpoint to check if the user has the RHEL for x86_64 product. Will be used for gating content template creation in the UI.

## Testing steps
**For local testing**

1. Turn off "devel_org" in the config.yaml
2. `GET http://localhost:8000/api/content-sources/v1.0/subscription_check/` should return false
3. Turn on "devel_org"
4. `GET http://localhost:8000/api/content-sources/v1.0/subscription_check/` should return true
5. Running it again, you should see a faster response time from the caching


jlsherrill:
https://issues.redhat.com/browse/HMS-4525

xbhouse (approved):
this looks good to me! just needs a rebase, that might fix the unit tests too

PR #797, opened 2024-09-04
Fixes 4666: lookup content sets instead of listing

jlsherrill:
## Summary

The newer version of candlepin limits the number of content sets returned and errors if more than 3000 are returned. 
Candlepin does support  paging, but these options are not exposed in the api spec for reasons.
We really don't need to list them all, so this changes it to fetch the contet sets by label (which is a new feature)

## Testing steps

Create a content template with some subset of rhel 9 repos and a custom repo.  Verify that a client sees just those pieces of content (or check the db and verify that the  cp_environment_content table in candlepin has what is expected


jlsherrill:
https://issues.redhat.com/browse/HMS-4666

rverdile (approved):
looks good, seeing only expected content sets on registered client

jlsherrill:
merging since openapi was fixed by another PR

PR #799, opened 2024-09-05
Build: use kin-openapi with order fix

jlsherrill:


jlsherrill:
note the test will fail until https://github.com/content-services/content-sources-backend/pull/797 is merged, but this doesn't affect those tests

xbhouse (approved):

PR #803, opened 2024-09-06
Fixes 4575: use identity hdr for pulp auth

jlsherrill:
## Testing steps

Going to see if i can configure local devel to prove it exists, but have to wait for pulp team to be be back


jlsherrill:
https://issues.redhat.com/browse/HMS-4575

PR #814, opened 2024-09-12
Fixes 4712: add content override for sslverifystatus

jlsherrill:
## Summary

By default clients registered to stage/prod have  sslsslverifystatus set to 1, which works fine for cdn.redhat.com, but not for s3.   This overrides that setting for each content set.

## Testing steps

Register a system and associate it to a template: https://github.com/content-services/content-sources-backend/blob/main/docs/register_client.md

on the client run 'subscription-manager refresh'

```
# grep sslverifystatus /etc/yum.repos.d/redhat.repo 
sslverifystatus = 0
```

### DIFF
--- a/pkg/candlepin_client/content.go
+++ b/pkg/candlepin_client/content.go
@@ -9,6 +9,7 @@ import (
 
 const OverrideNameBaseUrl = "baseurl"
 const OverrideNameCaCert = "sslcacert"
+const OverrideSSLVerifyStatus = "sslverifystatus"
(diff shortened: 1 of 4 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-4712

Andrewgdewar (approved):
Was able to reproduce the expected outcome! Ack!

jlsherrill:
apologies, this will have to be qe'd in stage!

PR #821, opened 2024-09-17
Build: update subman config command in register client doc

dominikvagner:
## Summary
When setting up a testing RHEL VM for the first time, the command in the docs for configuring subscription-manager didn't work, but a one I found in our chat did, so it should be updated. 😄

### DIFF
--- a/docs/register_client.md
+++ b/docs/register_client.md
@@ -42,7 +42,7 @@ for example, if a VM has an ip address of 192.168.122.58, the Host's IP address 
 ```
 4. Configure the client, within the VM:
 ```bash
-subscription-manager config --server.hostname=pulp.content --server.port=8444  --server.prefix=/candlepin --server.insecure=1
+subscription-manager config --server.hostname=subscription.rhsm.stage.redhat.com --server.port=443  --server.prefix=/subscription --server.insecure=0  --server.proxy_hostname=squid.corp.redhat.com  --server.proxy_port=3128
 ```
 5. Register the client within the VM:
 ```bash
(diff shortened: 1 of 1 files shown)


docs/register_client.md
```diff
@@ -42,7 +42,7 @@ for example, if a VM has an ip address of 192.168.122.58, the Host's IP address
 ```
 4. Configure the client, within the VM:
 ```bash
-subscription-manager config --server.hostname=pulp.content --server.port=8444  --server.prefix=/candlepin --server.insecure=1
+subscription-manager config --server.hostname=subscription.rhsm.stage.redhat.com --server.port=443  --server.prefix=/subscription --server.insecure=0  --server.proxy_hostname=squid.corp.redhat.com  --server.proxy_port=3128
```
jlsherrill, line 45: so this command is for registering to stage, not for a local server (which is what this doc targeted before). 

we probably should document this, maybe in a new section at the bottom?

dominikvagner: oh, didn't know that they are for different use cases 😅 but in that case the one for local server didn't work for me with the setup from the guide 🤔

dominikvagner: alright, retested it and I found a mistake on my part, setting my firewall zone to "trusted" (`sudo firewall-cmd --set-default-zone=trusted`) didn't seem to be enough, had to completely turn off the firewall for it to work 🤔😅 
and @xbhouse also verified that it works (on Mac) as it is now, so this PR should be scraped 😄

jlsherrill: yeah i think the stage config command is valuable.  I always have to search my slack for it :)

jlsherrill (approved):

PR #824, opened 2024-09-17
Fixes 4735: snapshot repos even if last task was deleted

jlsherrill:
## Summary

If a repo's last_snapshot_task got cleaned up the repo was never considered for nightly snapshotting because:
1)  the query that queries for repos to snapshot assumed that either the last_snapshot_task_uuid would be null, or it would exist and have a finished date
2) we have code to snapshot repos even if not needed yet to help spread out the load, but apparently sorting in ascending order puts NULLs last, so they would never have been selected

## Testing steps
On main:
1.  Create a few repos, or import some red hat ones and trigger them to sync with ```go run cmd/external-repos/main.go nightly-jobs```
2.   After the repos have synced, run:
 ```go run cmd/external-repos/main.go nightly-jobs 6```

This should trigger at least 1 snapshot task:
```

```
3.  Now "clean" the task:

```
make db-cli-connect

delete from tasks;
```

4.  Try to run the nightly jobs again:
 ```go run cmd/external-repos/main.go nightly-jobs 6```

You won't see any of these snapshot task get enqueued 

5.  Checkout this pr and ```make db-migrate-up```
6. Re-run the nightly jobs:

 ```go run cmd/external-repos/main.go nightly-jobs 6```

You should see their snapshot task be enqueued.


jlsherrill:
https://issues.redhat.com/browse/HMS-4735

pkg/tasks/queue/pgqueue.go
```diff
@@ -118,7 +118,7 @@ const (
                 DELETE FROM task_heartbeats
                 WHERE id = $1`
 	sqlDeleteAllTasks = `
-                TRUNCATE tasks, task_heartbeats, task_dependencies`
+                TRUNCATE task_heartbeats, task_dependencies; DELETE FROM TASKS;`
```
jlsherrill, line 121: can't use truncate because now that is a foreign key reference to this table

rverdile (approved):
looks good, just needs rebase!

PR #825, opened 2024-09-17
Fixes 3319: Add metrics/ping test

Andrewgdewar:
## Summary

- Adds a ping and metrics health check to integration tests.


jlsherrill:
https://issues.redhat.com/browse/HMS-3319

test/integration/health_check_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	s.metricsServer = &http.Server{
+		Addr:    "127.0.0.1:9005",
+		Handler: metricsRouter,
+	}
+}
+
+func (s *HealthCheckSuite) TearDownTest() {
+	s.cancel()
+	err := s.metricsServer.Shutdown(context.Background())
```
xbhouse, line 68: we should make sure both servers are shut down here

test/integration/health_check_test.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		log.Error().Err(err).Msg("Could not shutdown server")
+	}
+	s.Suite.TearDownTest()
+}
+
+func (s *HealthCheckSuite) serveMetricRouter(req *http.Request) (int, []byte, error) {
```
xbhouse, line 75: you could use just one serveRouter method and pass in an *http.Server, then call it with either the metricsServer or the pingServer in your tests

xbhouse (approved):
looks good!! ack!

PR #827, opened 2024-09-19
Fixes 4745: use domain list to detect pulp is up

jlsherrill:
## Summary

Changes our ping call to pulp to use domain list as we won't have permission for remotes list
The health check (using metrics) already uses this!

## Testing steps

monitor the pulp api:
```
 podman logs -f cs_pulp_api_1 
```

launch the server
```
go run cmd/content-sources/main.go api instrumentation consumer
```

call the ping api:
```
 curl localhost:9000/ping  
````

you should only see calls to /api/pulp/default/api/v3/domains/?name=default
You'll only see ping hit this endpoint once per server start.  You should see it get hit by the metrics go routine every ~30 minutes.

### DIFF
--- a/pkg/handler/api.go
+++ b/pkg/handler/api.go
@@ -4,11 +4,12 @@ import (
 	_ "embed"
 	"encoding/json"
 	"fmt"
-	"github.com/content-services/content-sources-backend/pkg/cache"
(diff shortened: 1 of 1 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-4745

xbhouse (approved):

PR #833, opened 2024-09-26
Refs 4739: new metric check for failing snapshot tasks on RH repos

dominikvagner:
## Summary
This PR adds a new metric for checking if the snapshot tasks are failing on RH repos, specifically if there isn't a successful one in the last 36 hours and returns number of repos on which is this happening.

## Testing steps
1. Import RH repos.
2. Make the snapshot task fail for some repo(s) (i.e. status = failed) and see if the metric increases.

### DIFF
--- a/pkg/dao/interfaces.go
+++ b/pkg/dao/interfaces.go
@@ -134,6 +134,7 @@ type MetricsDao interface {
 	PendingTasksAverageLatency(ctx context.Context) float64
 	PendingTasksCount(ctx context.Context) int64
 	PendingTasksOldestTask(ctx context.Context) float64
+	RHReposNoSuccessfulSnapshotTaskIn36Hours(ctx context.Context) int64
 }
 
 //go:generate $GO_OUTPUT/mockery --name TaskInfoDao --filename task_info_mock.go --inpackage
--- a/pkg/dao/metrics.go
+++ b/pkg/dao/metrics.go
@@ -142,3 +142,24 @@ func (d metricsDaoImpl) PendingTasksOldestTask(ctx context.Context) float64 {
 	}
 	return time.Since(*task.Queued).Seconds()
 }
+
+func (d metricsDaoImpl) RHReposNoSuccessfulSnapshotTaskIn36Hours(ctx context.Context) int64 {
+	var output int64 = -1
+	date := time.Now().Add(-36 * time.Hour).Format(time.RFC3339)
+
+	subQuery := d.db.WithContext(ctx).
+		Model(&models.RepositoryConfiguration{}).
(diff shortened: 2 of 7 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-4739

xbhouse (approved):
metrics look good to me! @rverdile can you take a look when you get a chance in case i missed anything?

rverdile (commented):
Small comment, looks really good overall

pkg/dao/interfaces.go
```diff
@@ -134,6 +134,7 @@ type MetricsDao interface {
 	PendingTasksAverageLatency(ctx context.Context) float64
 	PendingTasksCount(ctx context.Context) int64
 	PendingTasksOldestTask(ctx context.Context) float64
+	RHReposNoSuccessfulSnapshotTaskIn36Hours(ctx context.Context) int64
```
rverdile, line 137:
```suggestion
	RHReposSnapshotNotCompletedInLast36HoursCount(ctx context.Context) int64
```
This is a nitpick, but what do you think about renaming this metric?

"No" can also be an abbreviation for "number", so "RHReposNoSuccessfulSnapshotTaskIn36Hours" could be interpreted as "Number of sucessful snapshot tasks"

dominikvagner: Haven't though of that interpretation, changed! 😁👍🏼

pkg/dao/metrics.go
```diff
@@ -142,3 +142,24 @@ func (d metricsDaoImpl) PendingTasksOldestTask(ctx context.Context) float64 {
...
 	return time.Since(*task.Queued).Seconds()
 }
+
+func (d metricsDaoImpl) RHReposNoSuccessfulSnapshotTaskIn36Hours(ctx context.Context) int64 {
+	var output int64 = -1
+	date := time.Now().Add(-36 * time.Hour).Format(time.RFC3339)
+
+	subQuery := d.db.WithContext(ctx).
+		Model(&models.RepositoryConfiguration{}).
+		Select("repository_configurations.uuid, bool_or(tasks.status ILIKE ? AND tasks.finished_at > ?) AS has_successful_tasks", fmt.Sprintf("%%%s%%", config.TaskStatusCompleted), date).
```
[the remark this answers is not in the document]
dominikvagner: Though I had tested for that edge case, but seems like I did not. I think it definitely should be included.
Added the solution in the latest commit 👍🏼

rverdile (approved):
looks good!!

mshriver:
@swadeley @jlsherrill @dominikvagner  I like this approach to production monitoring rather than trying to write an IQE test around this situation and identifying the failure mode from production test results. thanks!

PR #834, opened 2024-09-27
Fixes 4777: remove custom param in list candlepin products

rverdile:
## Summary
We don't need the custom param defined to correctly query products here. I think I misunderstood how the parameter worked when I first implemented this. 

Without this change, only custom products are included in the response, but we want to return custom and red hat products (which is the default behavior).

## Testing steps
I tested the API call in stage so it should work there

To test locally:
1. Turn off devel_org in the config file
2. `GET /api/content-sources/v1/subscription_check/` should return false
3. Turn on devel_org
4.  `GET /api/content-sources/v1/subscription_check/` should return true


jlsherrill:
https://issues.redhat.com/browse/HMS-4777

PR #843, opened 2024-10-10
Adding empty directory and mounting it

mayurilahane:
Need to create and mount an empty directory inside the pod using a container. 
So that you can upload any files from that directory or can use it as storage while testing if you download anything.

### DIFF
--- a/deployments/testing-integration.yaml
+++ b/deployments/testing-integration.yaml
@@ -27,6 +27,9 @@ objects:
         - name: sel-shm
           emptyDir:
             medium: Memory
+        - name: sel-downloads
+          emptyDir:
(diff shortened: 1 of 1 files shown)


jlsherrill:
test failure doesn't look related?

swadeley:
> test failure doesn't look related?

No it is not and that should be fixed now, I'll try a retest

swadeley:
Hi @mayurilahane, I see you added this to `deployments/testing-integration.yaml`

 is this upload test only going to run in integration testing?

To use this in content-sources testing you need to add to `deployments/deployment.yaml`

swadeley (approved):
ACK

PR #854, opened 2024-10-17
Fixes 4860: use timestamptz for date comparisons

jlsherrill:
## Summary

Previously the change to this query improperly used the timestamp postgresql type which ignores timezones, by using timestamptz, the query now works properly.  We'll also just convert the date to UTC to be more consistent with using UTC dates.

## Testing steps

Fly to somewhere that doesn't use UTC.

Run the TestUseLatest integration test in test/integration/update_template_content_test.go


jlsherrill:
https://issues.redhat.com/browse/HMS-4860

rverdile (approved):
lgtm!

PR #856, opened 2024-10-21
Fixes 4552: fix missing method bug with mock gen

dominikvagner:
## Summary
This PR addresses the missing method bug (i.e. automates commenting out the `registry_mock.go` file) when generating mocks and updates the way we use the `mockery` tool to a new config as the old way be deprecated.

## Testing steps
- Verify that after adding a new method to a `dao` interface and to its "impl struct", `make mock` runs successfully without errors.
- Also `make mock` should now run without warnings about using an outdated config.


jlsherrill:
https://issues.redhat.com/browse/HMS-4552

mk/mockery.mk
```diff
@@ -8,4 +8,6 @@ $(GO_OUTPUT)/mockery: ## Install mockery locally on your GO_OUTPUT (./release) d
 
 .PHONY: mock
 mock: $(GO_OUTPUT)/mockery ## Install mockery if it isn't already in ./release directory and regenerate mocks
-	go generate ./...
\ No newline at end of file
+	sed -i '2,$$ s/^/\/\//' ./pkg/dao/registry_mock.go
```
xbhouse, line 11: i see this error running `make mock` on a mac:

`sed: 1: "./pkg/dao/registry_mock.go": invalid command code .`

i think it's because the sed tools for GNU and BSD work slightly differently. the commands you have do work on fedora. maybe we can conditionally run one or the other depending on the user's OS? or maybe there's a command that works cross-platform? 

looks like this works on macs:
```
sed -i '' '2,$$ s/^/\/\//' ./pkg/dao/registry_mock.go
sed -i '' '2,$$ s/^\/\///' ./pkg/dao/registry_mock.go
```

mk/mockery.mk
```diff
@@ -8,4 +8,6 @@ $(GO_OUTPUT)/mockery: ## Install mockery locally on your GO_OUTPUT (./release) d
...
+	mockery
```
xbhouse, line 12: wouldn't this call mockery installed outside of the project? i don't have mockery installed locally and this fails for me 😞 maybe it should be `$(GO_OUTPUT)/mockery`?

dominikvagner: yeah, it definitely should be `$(GO_OUTPUT)/mockery`, didn't realize I was using my locally installed mockery 🤦🏼 
thanks for catching that 👍🏼

mk/mockery.mk
dominikvagner: hrm, didn't think sed would require behave differently between GNU and BSD 🤔
pushed a fix for this, so please try again 😄

xbhouse: yep it works now! 🎉

xbhouse (approved):
lgtm! nice work!! 🚀

PR #869, opened 2024-10-31
Fixes 4610: don't trigger snapshot on adding slash to url

xbhouse:
## Summary

Adjusts comparison to check formatted url to keep from triggering snapshots if only a slash was added or removed.

## Testing steps

1. Create a repository and let it snapshot
2. Update the repository URL by adding an extra slash or removing a slash. Snapshots should not be triggered with either PUT or PATCH


jlsherrill:
https://issues.redhat.com/browse/HMS-4610

jlsherrill (approved):

swadeley:
Hi

works as described.

Thank you

PR #870, opened 2024-10-31
Fixes 3877: Add required fields for API spec requests

Andrewgdewar:
## Summary

- Add required fields for API spec requests.

### DIFF
--- a/pkg/api/content_units.go
+++ b/pkg/api/content_units.go
@@ -1,16 +1,16 @@
 package api
 
 type ContentUnitListRequest struct {
-	UUID   string `param:"uuid"`    // Identifier of the repository
-	Search string `query:"search"`  // Search string based query to optionally filter-on
-	SortBy string `query:"sort_by"` // SortBy sets the sort order of the result
+	UUID   string `param:"uuid" validate:"required"`    // Identifier of the repository
+	Search string `query:"search" validate:"required"`  // Search string based query to optionally filter-on
+	SortBy string `query:"sort_by" validate:"required"` // SortBy sets the sort order of the result
 }
 
 type ContentUnitSearchRequest struct {
-	URLs   []string `json:"urls,omitempty"`  // URLs of repositories to search
-	UUIDs  []string `json:"uuids,omitempty"` // List of repository UUIDs to search
-	Search string   `json:"search"`          // Search string to search content unit names
-	Limit  *int     `json:"limit,omitempty"` // Maximum number of records to return for the search
+	URLs   []string `json:"urls,omitempty"`             // URLs of repositories to search
+	UUIDs  []string `json:"uuids,omitempty"`            // List of repository UUIDs to search
+	Search string   `json:"search" validate:"required"` // Search string to search content unit names
+	Limit  *int     `json:"limit,omitempty"`            // Maximum number of records to return for the search
 }
 
 const ContentUnitSearchRequestLimitDefault int = 100
--- a/pkg/api/content_units.go
+++ b/pkg/api/content_units.go
@@ -1,16 +1,16 @@
 package api
 
 type ContentUnitListRequest struct {
-	UUID   string `param:"uuid" validate:"required"`    // Identifier of the repository
-	Search string `query:"search" validate:"required"`  // Search string based query to optionally filter-on
-	SortBy string `query:"sort_by" validate:"required"` // SortBy sets the sort order of the result
+	UUID   string `param:"uuid"`    // Identifier of the repository
+	Search string `query:"search"`  // Search string based query to optionally filter-on
+	SortBy string `query:"sort_by"` // SortBy sets the sort order of the result
 }
 
 type ContentUnitSearchRequest struct {
-	URLs   []string `json:"urls,omitempty"`             // URLs of repositories to search
-	UUIDs  []string `json:"uuids,omitempty"`            // List of repository UUIDs to search
-	Search string   `json:"search" validate:"required"` // Search string to search content unit names
-	Limit  *int     `json:"limit,omitempty"`            // Maximum number of records to return for the search
+	URLs   []string `json:"urls,omitempty"`  // URLs of repositories to search
+	UUIDs  []string `json:"uuids,omitempty"` // List of repository UUIDs to search
+	Search string   `json:"search"`          // Search string to search content unit names
+	Limit  *int     `json:"limit,omitempty"` // Maximum number of records to return for the search
 }
 
 const ContentUnitSearchRequestLimitDefault int = 100
--- a/pkg/api/repositories.go
+++ b/pkg/api/repositories.go
@@ -39,20 +39,19 @@ type RepositoryResponse struct {
 
 // RepositoryRequest holds data received from request to create repository
 type RepositoryRequest struct {
-	UUID        *string `json:"uuid" readonly:"true" swaggerignore:"true"`
-	AccountID   *string `json:"account_id" readonly:"true" swaggerignore:"true"`   // Account ID of the owner
-	OrgID       *string `json:"org_id" readonly:"true" swaggerignore:"true"`       // Organization ID of the owner
-	Origin      *string `json:"origin" readonly:"true"`                            // Origin of the repository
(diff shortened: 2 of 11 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-3877

swadeley:
Hi

How to test these changes?

If I take the first one:
```
   "definitions": {
        "api.AddUploadsRequest": {
            "type": "object",
            "required": [
                "artifacts",
                "uploads"
            ],
```
Looking at "add_upload" in ephemeral I do not see anything difenret to that in stage:
```
app.content_sources.rest_client.repositories_api.add_upload.
                                                                      allowed_values  call_with_http_info() headers_map           params_map           
                                                                      api_client            callable()            location_map          settings             
                                                                      attribute_map         collection_format_map openapi_types         validations        
```

jlsherrill:
I don't know that the changes need to be tested, but if you regenerate the api bindings you will see that code will likely need to be changed.  for example, this method signature changes:

```
@@ -102,9 +102,12 @@ class ApiContentUnitSearchRequest(ModelNormal):
     ])

     @convert_js_args_to_python_args
-    def __init__(self, *args, **kwargs):  # noqa: E501
+    def __init__(self, search, *args, **kwargs):  # noqa: E501
         """ApiContentUnitSearchRequest - a model defined in OpenAPI

+        Args:
+            search (str): Search string to search content unit names
+
```

So  if you are using the generated bindings its clearer that 'search' is required to be specified.   This doesn't actually change the behavior of the API at all (hence why the tests all pass without generating the bindings)

pkg/api/content_units.go
```diff
@@ -1,16 +1,16 @@
 package api
 
 type ContentUnitListRequest struct {
-	UUID   string `param:"uuid"`    // Identifier of the repository
-	Search string `query:"search"`  // Search string based query to optionally filter-on
-	SortBy string `query:"sort_by"` // SortBy sets the sort order of the result
+	UUID   string `param:"uuid" validate:"required"`    // Identifier of the repository
+	Search string `query:"search" validate:"required"`  // Search string based query to optionally filter-on
+	SortBy string `query:"sort_by" validate:"required"` // SortBy sets the sort order of the result
```
rverdile, line 6: search and sort_by are likely not required inputs

pkg/api/content_units.go
```diff
@@ -1,16 +1,16 @@
...
 type ContentUnitSearchRequest struct {
-	URLs   []string `json:"urls,omitempty"`  // URLs of repositories to search
-	UUIDs  []string `json:"uuids,omitempty"` // List of repository UUIDs to search
-	Search string   `json:"search"`          // Search string to search content unit names
-	Limit  *int     `json:"limit,omitempty"` // Maximum number of records to return for the search
+	URLs   []string `json:"urls,omitempty"`             // URLs of repositories to search
+	UUIDs  []string `json:"uuids,omitempty"`            // List of repository UUIDs to search
+	Search string   `json:"search" validate:"required"` // Search string to search content unit names
+	Limit  *int     `json:"limit,omitempty"`            // Maximum number of records to return for the search
 }
```
rverdile, line 14: The required inputs for ContentUnitSearchRequest are URLs and UUIDs. With that said, only one of them is required. So I'm not sure if it makes sense to mark them both as required, or neither?

In theory we allow an empty search, so I wouldn't say search is required

pkg/api/rpms.go
rverdile, line 1: For all of these content unit search and list requests, I believe only the UUID is required. 

For the ones where you can specify UUIDs or URLs, similar to above to the question above, not sure if it should require both or neither.

pkg/api/templates.go
```diff
@@ -10,15 +10,15 @@ import (
...
+	Name            *string        `json:"name" validate:"required"`                                            // Name of the template
+	Description     *string        `json:"description"`                                                         // Description of the template
+	RepositoryUUIDS []string       `json:"repository_uuids" validate:"required"`                                // Repositories to add to the template
+	Arch            *string        `json:"arch" validate:"required"`                                            // Architecture of the template
+	Version         *string        `json:"version" validate:"required"`                                         // Version of the template
+	Date            *EmptiableDate `json:"date"`                                                                // Latest date to include snapshots for
+	OrgID           *string        `json:"org_id" readonly:"true" swaggerignore:"true" validate:"required"`     // Organization ID of the owner
+	User            *string        `json:"created_by" readonly:"true" swaggerignore:"true" validate:"required"` // User creating the template
```
rverdile, line 20: These come from the header, but they're not required in the request body, so I wouldn't include "required" here

pkg/api/uploads.go
```diff
@@ -3,8 +3,8 @@ package api
 import "time"
 
 type AddUploadsRequest struct {
-	Uploads   []Upload   `json:"uploads"`   // List of unfinished uploads
-	Artifacts []Artifact `json:"artifacts"` // List of created artifacts
+	Uploads   []Upload   `json:"uploads" validate:"required"`   // List of unfinished uploads
+	Artifacts []Artifact `json:"artifacts" validate:"required"` // List of created artifacts
```
xbhouse, line 7: i believe artifacts also isn't required, just the uploads. there isn't a public api to finish the uploads (which creates the artifacts) so someone using only the public apis would only ever send upload uuids

Andrewgdewar:
Thanks for the review @rverdile @xbhouse!

rverdile:
there's some tests failing :)

pkg/api/templates.go
```diff
@@ -45,7 +45,7 @@ type TemplateResponse struct {
 
 // We use a separate struct because version and arch cannot be updated
 type TemplateUpdateRequest struct {
-	UUID            *string        `json:"uuid" readonly:"true" swaggerignore:"true"`
+	UUID            *string        `json:"uuid" readonly:"true" swaggerignore:"true" validate:"required"`
```
rverdile, line 48: When updating the template, the uuid comes from the url param and not from the request body, so I wouldn't mark it as required here

rverdile (commented):
I thought I saw them before, but TemplateRequest and RepositoryRequest are missing

pkg/api/repositories.go
```diff
@@ -146,11 +146,11 @@ func (r *RepositoryRequest) FillDefaults() {
 }
 
 type RepositoryIntrospectRequest struct {
-	ResetCount bool `json:"reset_count"` // Reset the failed introspections count
+	ResetCount bool `json:"reset_count" validate:"required"` // Reset the failed introspections count
```
rverdile, line 149: ResetCount is not required for RepositoryIntrospectRequest

Andrewgdewar:
> I thought I saw them before, but TemplateRequest and RepositoryRequest are missing

Thanks for the feedback here, I did add to the template and repository request. Please confirm if those changes make sense.

pkg/api/content_units.go
```diff
@@ -1,9 +1,9 @@
 package api
 
 type ContentUnitListRequest struct {
-	UUID   string `param:"uuid"`    // Identifier of the repository
-	Search string `query:"search"`  // Search string based query to optionally filter-on
-	SortBy string `query:"sort_by"` // SortBy sets the sort order of the result
+	UUID   string `param:"uuid" validate:"required"`   // Identifier of the repository
+	Search string `query:"search" validate:"required"` // Search string based query to optionally filter-on
```
rverdile, line 5: search nor UUID is required here. Search is an optional filter and the UUID comes from the URL, so it's not actually passed by the user through the request body.

rverdile:
that's my last comment :)

Andrewgdewar:
@rverdile @xbhouse Thank you for your thorough review here, I'm still learning 🍼 👶, I appreciate your patience!!

rverdile (approved):
lgtm!

swadeley:
Hi

I found that I had to add `origin="external` to all repo create calls.
No changes to anything related to search this time (that had `origin="external`  from a while  back)

rverdile:
Ah, it seems origin is actually not a required field. We should remove that then.

PR #875, opened 2024-11-06
Fixes 4927: add flag to prevent requeuing canceled tasks

xbhouse:
## Summary

Adds a `cancel_attempted` field to the tasks table and prevents requeuing a task if it was canceled

## Testing steps

1. On main, create a repo and let it snapshot
2. You can use this script to reproduce the "could not find template with UUID" errors (you'd need to adjust it to use your identity header and change the `repository_uuids` to include the repo you've created). Running this on main should reproduce those errors (you may not need to loop as many times):

```
for i in {1..50}; do
  UUID=$(curl -s -X POST --location "http://localhost:8000/api/content-sources/v1.0/templates/"     -H "x-rh-identity: eyJpZGVudGl0eSI6eyJvcmdfaWQiOiIxNzY4NDYzMiIsICJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJicnl0dGFuaWVob3VzZSJ9LCJhY2NvdW50X251bWJlciI6IjExNDkxNzg2IiwiaW50ZXJuYWwiOnsib3JnX2lkIjoiMTc2ODQ2MzIifX19Cg=="     -H "Content-Type: application/json"     -d "{
          \"name\":\"test-$RANDOM\",
          \"arch\": \"x86_64\",
          \"version\": \"8\",
          \"repository_uuids\": [\"0de58827-5301-44da-a226-85db660d6d1f\"],
          \"date\": \"2024-11-05T14:04:05Z\"
        }"  | jq -r .uuid)

  curl -s -X DELETE --location "http://localhost:8000/api/content-sources/v1.0/templates/$UUID"     -H "x-rh-identity: eyJpZGVudGl0eSI6eyJvcmdfaWQiOiIxNzY4NDYzMiIsICJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJicnl0dGFuaWVob3VzZSJ9LCJhY2NvdW50X251bWJlciI6IjExNDkxNzg2IiwiaW50ZXJuYWwiOnsib3JnX2lkIjoiMTc2ODQ2MzIifX19Cg=="

done
```
3. Checkout this PR, run the migration, and run the script again. You shouldn't see any requeues or "could not find template with UUID" errors


jlsherrill:
https://issues.redhat.com/browse/HMS-4927

pkg/tasks/queue/pgqueue.go
```diff
@@ -120,6 +120,10 @@ const (
                 WHERE id = $1`
 	sqlDeleteAllTasks = `
                 TRUNCATE task_heartbeats, task_dependencies; DELETE FROM TASKS;`
+	sqlSetCancelAttempted = `
+				UPDATE tasks 
+				SET cancel_attempted = true 
+				WHERE id = $1`
```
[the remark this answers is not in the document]
xbhouse: ah yep that does make sense, i set it here thinking this is where a request to cancel is first detected. setting it in the cancel query does get rid of the heartbeat warnings too 💡

pkg/tasks/queue/pgqueue.go
```diff
@@ -669,6 +669,9 @@ func (p *PgQueue) Requeue(taskId uuid.UUID) error {
 	if err == pgx.ErrNoRows {
 		return ErrNotExist
 	}
+	if info.CancelAttempted {
+		return ErrTaskCanceled
+	}
```
rverdile, line 674: One more thing to add here, I would also check that the task isn't "running" before returning this error.

This Requeue method is currently used by the tasking system to requeue running tasks. We wouldn't want to prevent a "running" task from re-queuing here if it were canceled, because then it would be stuck in the "running" state.

pkg/tasks/queue/pgqueue_test.go
rverdile, line 1: Could you add a test for the "RequeueFailedTasks" query? I imagine it would look like this test, except you'd cancel the task before failing it: https://github.com/content-services/content-sources-backend/blob/main/pkg/tasks/queue/pgqueue_test.go#L274

rverdile: And in this case it wouldn't requeue the task

pkg/tasks/queue/pgqueue.go
rverdile: This isn't a realistic scenario, but it could be an edge case

rverdile (approved):
nice job!

PR #879, opened 2024-11-07
Fixes 4964: add support for docker and importing arm repos

xbhouse:
## Summary

- Adjusts the make command for container volume removal to add support for docker
- Adds a selector to RHEL9 arm repos so it's easier to import those repos

## Testing steps

1. Using docker, `make compose-clean` should clear out all unused volumes
2. Using podman, `make compose-clean` should still work as before
3. Running `OPTIONS_REPOSITORY_IMPORT_FILTER=arm go run cmd/external-repos/main.go import && go run cmd/external-repos/main.go nightly-jobs` should import and snapshot RHEL9 arm repos

### DIFF
--- a/mk/compose.mk
+++ b/mk/compose.mk
@@ -31,7 +31,7 @@ compose-down: ## Shut down service  depdencies using podman(docker)-compose
 compose-clean: compose-down
 	if [ "$(DOCKER)" == "docker" ]; then \
 		$(DOCKER) volume prune --force --all; \
-    elif [ "$(DOCKER)" == "podman" ]; then \
+	elif [ "$(DOCKER)" == "podman" ]; then \
 		$(DOCKER) volume prune --force; \
 	fi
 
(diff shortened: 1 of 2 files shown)


mk/compose.mk
```diff
@@ -29,7 +29,7 @@ compose-down: ## Shut down service  depdencies using podman(docker)-compose
 
 .PHONY: compose-clean ## Clear out data (dbs, files) for service dependencies
 compose-clean: compose-down
-	$(DOCKER) volume prune --force
+	$(DOCKER) volume prune --force --all
```
dominikvagner, line 32: sadly this flag isn't supported in Podman and when I checkout this PR and run `make compose-clean` it fails with this error:
```
podman volume prune --force --all
Error: unknown flag: --all
```
\
is it necessary to add this flag? 🤔

dominikvagner (changes requested):

mk/compose.mk
xbhouse: unfortunately it seems either this flag or `--filter all=true` is necessary for docker to remove all anon and named volumes now, but neither are supported by podman 😞  added in a conditional check to get around this. should work for both now 🤞

dominikvagner: alrighty, `make compose-clean` works now, thanks! 🎉

dominikvagner (approved):
everything works now, both the compose and the import of rhel 9 arm repos! 👌🏼 
approved, good job! 👏🏼😄

PR #883, opened 2024-11-11
Fixes 4962: remove content from product on delete

dominikvagner:
## Summary
This PR fixes an issues caused by a recent candlepin update which removed the convenience feature of auto-unlink content on deletion from products, now a separate call to remove content from the product has to be done before deleting the content.

## Testing steps
Tests and pipeline passes successfully.

### DIFF
--- a/pkg/candlepin_client/interface.go
+++ b/pkg/candlepin_client/interface.go
@@ -24,6 +24,7 @@ type CandlepinClient interface {
 	CreateContentBatch(ctx context.Context, orgID string, content []caliri.ContentDTO) error
 	CreateContent(ctx context.Context, orgID string, content caliri.ContentDTO) error
 	AddContentBatchToProduct(ctx context.Context, orgID string, contentIDs []string) error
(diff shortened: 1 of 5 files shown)


pkg/candlepin_client/interface.go
```diff
@@ -24,6 +24,7 @@ type CandlepinClient interface {
 	CreateContentBatch(ctx context.Context, orgID string, content []caliri.ContentDTO) error
 	CreateContent(ctx context.Context, orgID string, content caliri.ContentDTO) error
 	AddContentBatchToProduct(ctx context.Context, orgID string, contentIDs []string) error
+	RemoveContentFromProduct(ctx context.Context, ownerKey, repoConfigUUID string) error
```
jlsherrill, line 27:
```suggestion
	RemoveContentFromProduct(ctx context.Context, orgId, repoConfigUUID string) error
```

jlsherrill (approved):
ack from me too

PR #885, opened 2024-11-12
Fixes 4887: use livez for checking pulp connectivity

dominikvagner:
## Summary
This PR changes the pulp connectivity check in the metrics collector to use the `livez` endpoint instead of listing the domains for checking this.

## Testing steps
- Start all the necessary containers and the server, and verify that metrics report the pulp connectivity as '1'.
- Stop the `cs_pulp_api_1` and `cs_pulp_content_1` containers.
- Verify the pulp connectivity metric is now '0'.


xbhouse (approved):
working as expected! mock fix looks good too 🎉

PR #893, opened 2024-11-18
Fixes 4925: fixup DBErrorToApi funcs

jlsherrill:

### DIFF
--- a/pkg/dao/repository_configs.go
+++ b/pkg/dao/repository_configs.go
@@ -81,7 +81,7 @@ func RepositoryDBErrorToApi(e error, uuid *string) *ce.DaoError {
 		}
 	} else {
 		daoErr = ce.DaoError{
-			Message:  "Database Error",
+			Message:  e.Error(),
 			NotFound: ce.HttpCodeForDaoError(e) == 404, // Check if isNotFoundError
 		}
 	}
--- a/pkg/dao/repository_configs.go
+++ b/pkg/dao/repository_configs.go
@@ -39,7 +39,7 @@ func GetRepositoryConfigDao(db *gorm.DB, pulpClient pulp_client.PulpClient) Repo
 	}
 }
 
-func DBErrorToApi(e error) *ce.DaoError {
+func RepositoryDBErrorToApi(e error, uuid *string) *ce.DaoError {
 	var dupKeyName string
 	if e == nil {
 		return nil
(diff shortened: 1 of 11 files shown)


pkg/dao/snapshots.go
```diff
@@ -219,17 +207,36 @@ func (sDao *snapshotDaoImpl) FetchModel(ctx context.Context, uuid string, includ
...
 }
 
+func SnapshotsDBToApiError(e error, uuid *string) *ce.DaoError {
+	if e == nil {
+		return nil
+	}
+
+	daoError := ce.DaoError{}
+	if errors.Is(e, gorm.ErrRecordNotFound) {
+		msg := "Task not found"
```
xbhouse, line 222:
```suggestion
		msg := "Snapshot not found"
```

pkg/dao/snapshots.go
```diff
@@ -219,17 +207,36 @@ func (sDao *snapshotDaoImpl) FetchModel(ctx context.Context, uuid string, includ
...
+		if uuid != nil {
+			msg = fmt.Sprintf("Could not find snapshot with UUID %s", *uuid)
```
xbhouse, line 224: nitpicky but i think just to be consistent with the others, the error message here could be `Snapshot with UUID %s not found`

pkg/dao/repository_configs.go
```diff
@@ -69,9 +69,21 @@ func DBErrorToApi(e error) *ce.DaoError {
...
+		if uuid != nil {
+			msg = fmt.Sprintf("Repository with UUID %s not found", *uuid)
+		}
+		daoErr = ce.DaoError{
+			Message:  msg,
+			NotFound: true,
+		}
+	} else {
+		daoErr = ce.DaoError{
+			Message:  "Database Error",
```
xbhouse, line 84: i think this message should also display e.Error() if it's not a NotFound error

pkg/dao/templates.go
```diff
@@ -21,7 +22,7 @@ type templateDaoImpl struct {
 	db *gorm.DB
 }
 
-func (t templateDaoImpl) DBToApiError(e error) *ce.DaoError {
+func (t templateDaoImpl) DBToApiError(e error, uuid *string) *ce.DaoError {
```
xbhouse, line 25: what do you think about either renaming this to TemplateDBToApiError and removing the receiver or also using a receiver on the other DBToApiError funcs and removing the object name? 

or using one DBToApiError method for all objects and just passing in the uuid and the object name? would that work?

pkg/dao/snapshots_test.go
```diff
@@ -996,7 +996,7 @@ func (s *SnapshotsSuite) TestBulkDeleteNotFound() {
 
 	errs := sDao.BulkDelete(context.Background(), []string{s2.UUID, s3.UUID})
 	assert.Len(t, slices.DeleteFunc(errs, func(e error) bool { return e == nil }), 1)
-	assert.Equal(t, fmt.Sprintf("Could not find snapshot with UUID %s", s2.UUID), errs[0].Error())
+	assert.Equal(t, fmt.Sprintf("Could not find snapshot with UUID %s: record not found", s2.UUID), errs[0].Error())
```
swadeley, line 999: see comment about about "Could not find snapshot"

pkg/dao/templates.go
jlsherrill: Some of them check for specific contraints, so as we are doin git today, i'm not sure we can do one method, but yes moving away from a reciever makes sense.

PR #901, opened 2024-11-21
HMS-4747: automatic deletion of outdated snapshots

dominikvagner:
## Summary
This PR adds automatic deletion of outdated snapshots to the `nightly-jobs` command. The date after which the snapshots should be deleted is settable in the config.
Also adds a list of snapshots which are going to be deleted in the next 14 days to the fetch/list template endpoints.

## Testing steps
1. Create a repository and let it snapshot.
2. Create a template that uses the repo and is using the latest snap.
3. Connect to the DB CLI `make db-cli-connect` and set the snapshot date to the past.
    - `UPDATE snapshots SET created_at = (NOW() - CAST('400 days' AS INTERVAL)) WHERE uuid = '$SNAPSHOT_UUID';`
4. Run nightly jobs (`go run cmd/external-repos/main.go nightly-jobs`) and verify that there wasn't a "delete-snapshots" task enqueued.
5. Verify the outdated snaphot is listed in both the "snapshots" and "to_be_deleted_snapshots" when querying templates.
6. Edit the repo so that it snapshots again and change the template to use the old one.
7. Run the nightly jobs again and verify the snapshot has been deleted and the template updated to use the other one.


pkg/config/config.go
```diff
@@ -243,6 +244,7 @@ func setDefaults(v *viper.Viper) {
 	v.SetDefault("options.template_event_topic", "platform.content-sources.template")
 	v.SetDefault("options.repository_import_filter", "")
 	v.SetDefault("options.external_url", "http://pulp.content:8000")
+	v.SetDefault("options.snapshot_retain_days_limit", 90)
```
jlsherrill, line 247: i'd probably set the default to a higher value (365), and then we can customize it in app-interface to do less for stage

jlsherrill:
https://issues.redhat.com/browse/HMS-4747

swadeley:
Hi, I am stuck with this:
```
E               iqe_content_sources_api.exceptions.ApiTypeError: Invalid type for variable 'to_be_deleted_snapshots'. Required value type is list and passed type was NoneType at ['received_data']['to_be_deleted_snapshots']
```

swadeley:
> Hi, I am stuck with this:
> 
> ```
> E               iqe_content_sources_api.exceptions.ApiTypeError: Invalid type for variable 'to_be_deleted_snapshots'. Required value type is list and passed type was NoneType at ['received_data']['to_be_deleted_snapshots']
> ```
 
Hi, thanks for the update. IQE tests now pass.

xbhouse:
i'm seeing this error when running nightly-jobs if there is a template using an outdated snapshot:

```
"error deleting rpm repository versions: 400 Bad Request: [\"The repository version cannot be deleted because it (or its publications) are currently being used to distribute content. Please update the necessary distributions first.\"]"
```

it seems like maybe the template isn't being updated to the next snapshot before the snapshot is deleted? though on the next requeue it does seem like the distribution was removed (getting a 404 when trying to view the content url and seeing resource not found errors), but the snapshot was removed from the template and still exists in our db. 

i may just be missing something when testing 😄 to reproduce, i'm using these steps:

1. create a template (set to the current date) with a RH repo and a custom repo
2. update the custom repo's snapshot's created_at field to be more than 365 days in the past 
3. edit the custom repo's url to get a new snapshot 
4. run nightly-jobs

cmd/external-repos/main.go
```diff
@@ -290,6 +297,86 @@ func enqueueSnapshotRepos(ctx context.Context, urls *[]string, interval *int) er
...
+	}
+
+	for _, repo := range repoConfigs {
+		// Fetch snapshots for repo and find those which are to be deleted
+		snaps, err := snapshotDao.FetchForRepoConfigUUID(ctx, repo.UUID)
+		if err != nil {
+			return fmt.Errorf("error fetching snapshots for repository %v", repo.Name)
+		}
+		if len(snaps) < 2 {
+			return nil
```
xbhouse, line 322: i think this will stop checking repos once it gets to the first repo with just 1 snapshot. you might want to `continue` here instead of returning?

xbhouse:
> i'm seeing this error when running nightly-jobs if there is a template using an outdated snapshot:
> 
> ```
> "error deleting rpm repository versions: 400 Bad Request: [\"The repository version cannot be deleted because it (or its publications) are currently being used to distribute content. Please update the necessary distributions first.\"]"
> ```
> 
> it seems like maybe the template isn't being updated to the next snapshot before the snapshot is deleted? though on the next requeue it does seem like the distribution was removed (getting a 404 when trying to view the content url and seeing resource not found errors), but the snapshot was removed from the template and still exists in our db.
> 
> i may just be missing something when testing 😄 to reproduce, i'm using these steps:
> 
> 1. create a template (set to the current date) with a RH repo and a custom repo
> 2. update the custom repo's snapshot's created_at field to be more than 365 days in the past
> 3. edit the custom repo's url to get a new snapshot
> 4. run nightly-jobs

your latest commit seems to have fixed this! 🎉

xbhouse (approved):
this looks good to me! @jlsherrill may want to take another look?

pkg/config/config.go
jlsherrill, line 247: i'd probably suggest we default this to 365, and then lower it in stage to 90

jlsherrill:
(using your PR to test some automation real quick)

jlsherrill:
https://issues.redhat.com/browse/HMS-HMS-4747

jlsherrill:
https://issues.redhat.com/browse/Fixes HMS-4747

jlsherrill (approved):

PR #902, opened 2024-11-21
Fixes 5034: update latest snapshot after uploading content

xbhouse:
## Summary

Triggers the update-latest-snapshot task after uploading content to a repo. Without this, use_latest templates with upload repos wouldn't get updated with the latest content.

## Testing steps

(It's probably easiest to do most of this in the UI)

1. Add a RH repo and an upload repo with 1 rpm, let them snapshot
2. Create a template using those repos and set it to use latest content
3. Upload another rpm to the upload repo
4. Check the template's packages via the UI or the API (`/templates/:uuid/rpms?search=<uploaded_rpm_name>`). They should include all the uploaded packages

### DIFF
--- a/pkg/handler/repositories_test.go
+++ b/pkg/handler/repositories_test.go
@@ -176,6 +176,15 @@ func mockTaskClientEnqueueAddUploads(repoSuite *ReposSuite, repo api.RepositoryR
 		repo.OrgID,
 		repo.RepositoryUUID,
 	).Return(nil)
+	repo.LastSnapshotTaskUUID = "00000000-0000-0000-0000-000000000000"
+	repoSuite.tcMock.On("Enqueue", queue.Task{
+		Typename:     config.UpdateLatestSnapshotTask,
+		Payload:      tasks.UpdateLatestSnapshotPayload{RepositoryConfigUUID: repo.UUID},
+		Dependencies: []uuid.UUID{dao.UuidifyString(repo.LastSnapshotTaskUUID)},
+		ObjectUUID:   &repo.RepositoryUUID,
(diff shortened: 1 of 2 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-5034

pkg/handler/repositories_test.go
```diff
@@ -176,6 +176,14 @@ func mockTaskClientEnqueueAddUploads(repoSuite *ReposSuite, repo api.RepositoryR
 		repo.OrgID,
 		repo.RepositoryUUID,
 	).Return(nil)
+	repoSuite.tcMock.On("Enqueue", queue.Task{
+		Typename:     config.UpdateLatestSnapshotTask,
+		Payload:      tasks.UpdateLatestSnapshotPayload{RepositoryConfigUUID: repo.UUID},
+		Dependencies: []uuid.UUID{uuid.Nil},
```
rverdile, line 182: I think the add uploads task should be a dependency, similar to here: https://github.com/content-services/content-sources-backend/blob/main/pkg/handler/repositories.go#L954

xbhouse: ahh i see, just pushed that up. is that what you mean?

rverdile: oh sorry I didn't realize I was looking at a test :man_facepalming:. What you had was okay!

rverdile: Either way is fine

xbhouse: ok lol i was a little confused, i'll just keep it as it is now then :)

rverdile (approved):
looks good!

swadeley:
Hi, looks good:

image

image



image

PR #844, opened 2024-10-10
Fixes 4716: Add "latest" snapshot distribution

Andrewgdewar:
## Summary

- When a new snapshot is taken either via a sync, or via an upload, a new distribution there is a new distribution in pulp pointing to /latest.

- The full URL to the above 'latest' snapshot distribution is returned with each snapshotted & upload repository, as the attribute:  "latest_snapshot_url"

- New route to download the repo-config with associated 'latest' url @/repositories/UUID/config.repo  

## Testing steps

- Create a custom repository > wait for snapshot > check the response for the "latest_snapshot_url"
- Call the /repositories/UUID/config.repo endpoint with the above repository uuid, see the latest_snapshot_url represented there as "url"
- Check that you can download the above distribution.

PR #903, opened 2024-11-22
Fixes 4926: job to create latest snap dist for existing repos

rverdile:
## Summary
A job to create the latest snapshot distribution for existing repos, because existing ones would have been missed by this PR: https://github.com/content-services/content-sources-backend/pull/844#discussion_r1815501837

## Testing steps
1. Apply this patch `git apply mypatch.patch` to prevent the latest snapshot distribution being created in the snapshot task.
```
diff --git a/pkg/tasks/snapshot_helper.go b/pkg/tasks/snapshot_helper.go
index 1c0e812..75e9bc6 100644
--- a/pkg/tasks/snapshot_helper.go
+++ b/pkg/tasks/snapshot_helper.go
@@ -63,12 +63,12 @@ func (sh *SnapshotHelper) Run(versionHref string) error {
                return err
        }
 
-       latestPathIdent := helpers.GetLatestRepoDistPath(*sh.payload.GetSnapshotIdent())
-
-       _, _, err = helper.CreateOrUpdateDistribution(sh.orgId, sh.repo.UUID, latestPathIdent, publicationHref)
-       if err != nil {
-               return err
-       }
+       //latestPathIdent := helpers.GetLatestRepoDistPath(*sh.payload.GetSnapshotIdent())
+       //
+       //_, _, err = helper.CreateOrUpdateDistribution(sh.orgId, sh.repo.UUID, latestPathIdent, publicationHref)
+       //if err != nil {
+       //      return err
+       //}
        version, err := sh.pulpClient.GetRpmRepositoryVersion(sh.ctx, versionHref)
        if err != nil {
                return err
```
2. Create multiple repos for multiple orgs, some with and without snapshotting
3. Notice that the for the "latest_snapshot_url" returned by those repos, the URL gives 404.
4. Run the job: `go run cmd/create_latest_distributions/main.go`
5. Those URLs that were giving 404s should now serve the latest snapshot


jlsherrill:
https://issues.redhat.com/browse/HMS-4926

pkg/jobs/create_latest_distributions.go
```diff
@@ -0,0 +1,69 @@
...
+	domains, err := daoReg.Domain.List(ctx)
+	if err != nil {
+		log.Fatal().Err(err).Msg("failed to list domains")
+	}
+	for _, domain := range domains {
+		pulpClient := pulp_client.GetPulpClientWithDomain(domain.DomainName)
+		distHelper := helpers.NewPulpDistributionHelper(ctx, pulpClient)
+
+		pageData := api.PaginationData{Limit: -1}
+		filterData := api.FilterData{Origin: config.OriginExternal + "," + config.OriginUpload}
```
xbhouse, line 34: i'm seeing 404s for the latest snapshot urls in RH repos, looks like RH repos aren't included here. was that intentional?

pkg/api/snapshots.go
```diff
@@ -6,15 +6,16 @@ import (
...
+	UUID            string           `json:"uuid"`
+	CreatedAt       time.Time        `json:"created_at"`                            // Datetime the snapshot was created
+	RepositoryPath  string           `json:"repository_path"`                       // Path to repository snapshot contents
+	ContentCounts   map[string]int64 `json:"content_counts"`                        // Count of each content type
+	AddedCounts     map[string]int64 `json:"added_counts"`                          // Count of each content type
+	RemovedCounts   map[string]int64 `json:"removed_counts"`                        // Count of each content type
+	URL             string           `json:"url"`                                   // URL to the snapshot's content
+	RepositoryName  string           `json:"repository_name"`                       // Name of repository the snapshot belongs to
+	RepositoryUUID  string           `json:"repository_uuid"`                       // UUID of the repository the snapshot belongs to
+	PublicationHref string           `json:"publication_href" swaggerignore:"true"` // Publication href of the snapshot in pulp
```
xbhouse, line 18: if swaggerignore is true, does that require an api doc update? looking at the PR check failure: `failed on setup with "iqe_content_sources_api.exceptions.ApiAttributeError: ApiSnapshotResponse has no attribute 'publication_href' at ['received_data']['data'][0]['last_snapshot']['publication_href']"`

pkg/jobs/create_latest_distributions.go
rverdile: I guess I was thinking red hat repos won't really have issue because they get updated frequently. But it probably makes sense to include them anyway! Just in case.

deployments/deployment.yaml
```diff
@@ -323,6 +323,64 @@ objects:
...
       jobs:
+        - name: create-latest-distributions
+          podSpec:
+            securityContext:
+              runAsNonRoot: true
+              runAsUser: 1001
+            image: ${IMAGE}:${IMAGE_TAG}
+            inheritEnv: true
+            command:
+              - jobs
```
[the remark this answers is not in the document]
rverdile: i don't know either, but probably :)

pkg/api/snapshots.go
rverdile: oh, I also meant to do `json:"-"`. That will allow it to be used in the struct, but not in API response

xbhouse (approved):
looks good!

PR #905, opened 2024-11-25
Fixes 4928: fix worker task requeuing on start/stop

dominikvagner:
## Summary
This PR fixes the problems introduced with the changes made to how we close the pg pool, which were required to prevent leaving hanging connections. The introduced problem was when workers tried requeuing tasks on the context stop, but the pool was already closed at that point, so the requeuing was moved to when the workers start.

## Testing steps
1. Enqueue a bunch of tasks.
2. CTRL+C to close the server.
3. Start the server again.
4. Verify that there are no errors:
    - error requeuing task with task_id: 959ae228-6974-499b-b049-8600832355a7
    - error starting database transaction: closed pool

### DIFF
--- a/pkg/tasks/queue/pgqueue.go
+++ b/pkg/tasks/queue/pgqueue.go
@@ -226,11 +226,6 @@ func NewPgxPool(ctx context.Context, url string) (*pgxpool.Pool, error) {
 		return nil, fmt.Errorf("error establishing connection: %w", err)
 	}
 
-	// Allow for context cancellation to release the pool
-	go func() {
-		<-ctx.Done()
-	}()
-
 	return pool, nil
 }
 
--- a/pkg/tasks/queue/pgqueue.go
+++ b/pkg/tasks/queue/pgqueue.go
@@ -131,6 +131,7 @@ const (
 type Pool interface {
 	Transaction
 	Acquire(ctx context.Context) (Connection, error)
+	Close()
 }
 
 // Transaction mimics the pgx.Tx struct
@@ -228,7 +229,6 @@ func NewPgxPool(ctx context.Context, url string) (*pgxpool.Pool, error) {
 	// Allow for context cancellation to release the pool
 	go func() {
 		<-ctx.Done()
-		pool.Close()
 	}()
 
 	return pool, nil
@@ -883,6 +883,10 @@ func (p *PgQueue) ListenForCancel(ctx context.Context, taskID uuid.UUID, cancelF
 		logger.Debug().Msg("[Canceled Task]")
 		cancelFunc(ErrTaskCanceled)
 	}
+}
+
+func (p *PgQueue) Close() {
+	p.Pool.Close()
 }
 
 func isContextCancelled(ctx context.Context) bool {
--- a/test/integration/suite_test.go
+++ b/test/integration/suite_test.go
@@ -91,7 +91,7 @@ func (s *Suite) SetupTest() {
 	go func() {
 		<-wkrCtx.Done()
 		wrk.Stop()
-		defer wkrQueue.Close()
+		wkrQueue.Close()
 	}()
 }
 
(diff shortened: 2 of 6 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-4928

pkg/tasks/queue/pgqueue.go
rverdile, line 1: I have one idea that might help simplify fix a bit. And I apologize for writing a vague card with no direction :sweat_smile: .

We could add a Close() method to the Queue interface, which would close the pool. And then also remove [closing the pool directly with canceled context](https://github.com/content-services/content-sources-backend/blob/main/pkg/tasks/queue/pgqueue.go#L228-L232), which was previously implemented.

Then in main.go we'd just have:
```
pgQueue, err := NewPgQueue()
defer pgQueue.Close()

// and then further down in the function
workerPool.Stop()
```
The added benefit of this is that, when the server exits, the workers are still able to requeue the tasks, because the pool won't close until the Stop() function has returned.

dominikvagner: Yeah that definitely looks like a nicer solution, I tried replicating it in a mock scenario and it seems to work even when the `workerPool.Stop()` function would be called inside of a goroutine, which I didn't know if it would work at first. 😅 
I will try changing this and if works I will update this 👍🏼

pkg/tasks/queue/pgqueue.go
```diff
@@ -228,7 +229,6 @@ func NewPgxPool(ctx context.Context, url string) (*pgxpool.Pool, error) {
 	// Allow for context cancellation to release the pool
 	go func() {
 		<-ctx.Done()
-		pool.Close()
 	}()
```
rverdile, line 232: this entire go routine can be removed because it's only used for closing the pool

test/integration/suite_test.go
```diff
@@ -91,6 +91,7 @@ func (s *Suite) SetupTest() {
 	go func() {
 		<-wkrCtx.Done()
 		wrk.Stop()
+		defer wkrQueue.Close()
```
rverdile, line 94: It doesn't matter functionally, just pointing it for understanding; you don't need the `defer` keyword, since it's already being called after `wrk.Stop()`. `defer` just means "call this  when the function ends".

pkg/tasks/queue/pgqueue.go
rverdile, line 1: After testing, it seems like your changes have fixed the "closed pool" issue. But it also seemed to surface this other issue, where sometimes (when exiting the server) tasks are failed before they are re-queued.

I can re-create this by CTRL+C interrupting the server while a snapshot is running.

```
2:37PM WRN [Finished Task] task failed error="Get \"http://localhost:8080/api%2Fpulp%2Fcs-f193155fee%2Fapi%2Fv3%2Ftasks%2F019388e1-1b07-7838-8d31-e0a4659fc275%2F\": server exited: no body" request_id=c7e896da-22ed-4591-b152-fa64c13e4d88 task_id=61c620b0-3f16-4616-a8a3-31c7a9e95913 task_type=snapshot
2:37PM ERR error requeuing task with task_id: 00000000-0000-0000-0000-000000000000 error="error requeuing task: task does not exist"

```

rverdile: To fix this, we'd probably have to do something like prevent the task from failing on server exit, which would allow it to requeue. The task fails only because closing the server cancels the context.

rverdile: Do you want to fix that here? Or I can open a card

pkg/tasks/queue/pgqueue.go
dominikvagner: removed 🫡

pkg/tasks/queue/pgqueue.go
dominikvagner: hmm, I was able to reproduce it as well 🤔
I would probably split that into a new task and leave this as is 😅

rverdile (approved):
looks good! opened a card for the other issue

PR #913, opened 2024-12-04
HMS-5063: create job to cleanup missing domains

xbhouse:
## Summary

Adds a job to clean up the domains in our DB that are not in pulp

## Testing steps

You can use the pulp CLI to test this, setup docs [here](https://pulpproject.org/pulp-cli/)

1. Create 2 repos in different orgs and let them snapshot
2. List domains in pulp via the pulp CLI: `pulp domain list | jq '.[].name'`
3. Delete one of the repos
4. Delete the domain associated to the deleted repo's org ID in pulp via the pulp CLI: `pulp domain destroy --name <domain_name>`
5. List the domains in pulp again to confirm the domains that are left
6. Run this job to clean up the domains in our DB: `go run cmd/jobs/main.go cleanup-missing-domains`
7. List the domains in our DB. They should now match pulp's cs-prefixed domains

### DIFF
--- a/pkg/jobs/cleanup_missing_domains.go
+++ b/pkg/jobs/cleanup_missing_domains.go
@@ -27,22 +27,20 @@ func CleanupMissingDomains() {
 		pulpHref, err := pulp_client.GetGlobalPulpClient().LookupDomain(ctx, domain.DomainName)
 		if err != nil {
 			log.Error().Err(err).Msg("failed to lookup pulp domain")
-		} else {
-			if pulpHref == "" {
-				var snapCount int64 = 0
-				result := db.DB.Model(models.Snapshot{}).
-					Joins("inner join repository_configurations on repository_configurations.uuid = snapshots.repository_configuration_uuid").
-					Where("org_id = ?", domain.OrgId).Count(&snapCount)
-				if result.Error != nil {
-					log.Fatal().Err(err).Msg("failed to fetch snapshots for org")
-				}
-				if snapCount > 0 {
-					log.Error().Err(err).Msg("skipping domain deletion, snapshots exist in this org")
-				} else {
-					err := daoReg.Domain.Delete(ctx, domain.OrgId, domain.DomainName)
-					if err != nil {
-						log.Error().Err(err).Msg("failed to delete domain")
-					}
+		} else if pulpHref == "" {
+			var snapCount int64 = 0
+			result := db.DB.Model(models.Snapshot{}).
+				Joins("inner join repository_configurations on repository_configurations.uuid = snapshots.repository_configuration_uuid").
+				Where("org_id = ?", domain.OrgId).Count(&snapCount)
+			if result.Error != nil {
+				log.Fatal().Err(err).Msg("failed to fetch snapshots for org")
+			}
+			if snapCount > 0 {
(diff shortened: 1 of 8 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-5063

pkg/jobs/cleanup_missing_domains.go
```diff
@@ -0,0 +1,37 @@
...
+
+	domains, err := daoReg.Domain.List(ctx)
+	if err != nil {
+		log.Fatal().Err(err).Msg("failed to list domains")
+	}
+	for _, domain := range domains {
+		pulpHref, err := pulp_client.GetGlobalPulpClient().LookupDomain(ctx, domain.DomainName)
+		if err != nil {
+			log.Error().Err(err).Msg("failed to lookup pulp domain")
+		}
```
jlsherrill, line 29: it'd be better if we explicitly checked for a 404, in case some other error occured.

pkg/jobs/cleanup_missing_domains.go
```diff
@@ -0,0 +1,37 @@
...
+		if pulpHref == "" {
+			err := daoReg.Domain.Delete(ctx, domain.OrgId, domain.DomainName)
+			if err != nil {
+				log.Error().Err(err).Msg("failed to delete domain")
+			}
```
jlsherrill, line 34: I wonder if we should check if any snapshots exist in the given org?  if they did and the domain didn't exist in pulp there may be a more complicated situation we should investigate (hopefully that doesn't happen).  So i guess i'm suggesting if the domain does not exist in pulp:
1.  check if any snapshots exist for that org
2. if so, log an error and skip domain db deletion

jlsherrill:
couple of comments, but otherwise looks good!  I'll do a full test after updates

jlsherrill:
https://issues.redhat.com/browse/Fixes 5063

pkg/jobs/cleanup_missing_domains.go
xbhouse: it looks like we don't return an error if the domain we filter by isn't found. should [this](https://github.com/content-services/content-sources-backend/blob/79e0a48e59ef3388f8f295050201da4aac04107f/pkg/pulp_client/domains.go#L86) return a 404 if the domain isn't found?

jlsherrill: ohh i see, it returns empty string.  Thats fine, i would just ignore the pulpHref if an err is thrown and skip to the next one?

pkg/jobs/cleanup_missing_domains.go
```diff
@@ -0,0 +1,50 @@
...
+	domains, err := daoReg.Domain.List(ctx)
+	if err != nil {
+		log.Fatal().Err(err).Msg("failed to list domains")
+	}
+	for _, domain := range domains {
+		pulpHref, err := pulp_client.GetGlobalPulpClient().LookupDomain(ctx, domain.DomainName)
+		if err != nil {
+			log.Error().Err(err).Msg("failed to lookup pulp domain")
+		} else {
+			if pulpHref == "" {
```
jlsherrill, line 31: a small  nitpick, but this could be `else if`

jlsherrill (approved):
ACK! tested and worked great

PR #917, opened 2024-12-10
Fixes 5144: nightly job failing

xbhouse:
## Summary

This fixes the nightly job failure by checking if the context was canceled before closing the pool. Here I added back a separate context for the listener, there might be a better way to do this without that though 🤔 

## Testing steps

1. Start the server and run `make repos-import-rhel9 && go run cmd/external-repos/main.go nightly-jobs`
2. Should not panic and RH repos should be snapshotted

### DIFF
--- a/pkg/tasks/queue/pgqueue.go
+++ b/pkg/tasks/queue/pgqueue.go
@@ -154,8 +154,9 @@ type Connection interface {
 
 // PgQueue a task queue backed by postgres, using pgxpool.Pool using a wrapper (PgxPoolWrapper) that implements a Pool interface
 type PgQueue struct {
-	Pool      Pool
-	dequeuers *dequeuers
+	Pool         Pool
+	dequeuers    *dequeuers
+	stopListener func()
 }
 
 // thread-safe list of dequeuers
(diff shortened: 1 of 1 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-5144

jlsherrill:
Also worth noting, on main "go run cmd/external-repos/main.go snapshot https://cdn.redhat.com/content/dist/layered/rhel8/x86_64/ansible/2/os/" seems to hang, but this resolves it!

rverdile:
I took a look at this 
> 9:14AM ERR Error unlistening for tasks in dequeue error="conn closed"

I don't really have a good solution. If closing the pool is causing too much weirdness, then I'm not sure we _need_ to close the pool outside of the tests. Similarly, I don't know how important it really is to unlisten if the connection is already closed?

jlsherrill:
@rverdile and i chatted about this, and think a waitgroup may help here?

jlsherrill:
we might also go ahead and merge this so fix the stage jobs and then we can deal with this issue separately ?

xbhouse:
> we might also go ahead and merge this so fix the stage jobs and then we can deal with this issue separately ?

ok, yea i think merging to fix the nightly-jobs is a good idea. could you open a card for the other issue?

jlsherrill:
will do!

jlsherrill (approved):

PR #923, opened 2024-12-17
Build: make pulp docker-compose reusable

jlsherrill:
## Summary

By allowing ports to be specified, we can reuse this for tang and other places

This also moves minio out of the pulp compose into the main compose


xbhouse (approved):
lgtm!

PR #927, opened 2024-12-19
HMS-5062: fix del-templ task fail if repo is deleted first

dominikvagner:
## Summary
This PR fixes a bug that happens when a repository exists within a template, and both are deleted, but the repository config is deleted before the template is deleted, then the `delete-templates` task gives a "repository not found" error and if this happens the template-repository distribution doesn't get deleted and remains an orphan.

The ticket assumed that the `delete-repository-snapshots` task didn't try deleting the template-repository distribution, but it actually already was trying to do so, but that step just needed to be moved earlier in the task and also do that for soft deleted templates. The `delete-template` task just needed to be changed to ignore not found repository configs.

## Testing steps
1. Create a custom repository and add it to a new template.
2. Note down the template-repository distribution href (look it up by uuids in `SELECT * FROM templates_repository_configurations;`) and verify it exists (`http -a admin:password :8080/${DISTRIBUTION_HREF}`).
3. Change `retry_wait_upper_bound` in `config.yaml` to 3m.
4. Edit the `Run()` method of `delete_templates.go` to force a failure by returning an error (add `return errors.new("test error")` at the beginning).
5. Delete the template.
6. Delete the repository configuration.
7. Once the `delete-repository-snapshot` task finishes, stop the server.
8. Fix the `Run()` from step 3, so it runs correctly, then restart the server.
9. The `delete-templates` tasks will retry and it should be completed successfully.
10. Verify the template-repository distribution was deleted (`http -a admin:password :8080/${DISTRIBUTION_HREF}` -> `404`).

### DIFF
--- a/pkg/dao/interfaces.go
+++ b/pkg/dao/interfaces.go
@@ -168,7 +168,7 @@ type TemplateDao interface {
 	Create(ctx context.Context, templateRequest api.TemplateRequest) (api.TemplateResponse, error)
 	Fetch(ctx context.Context, orgID string, uuid string, includeSoftDel bool) (api.TemplateResponse, error)
 	InternalOnlyFetchByName(ctx context.Context, name string) (models.Template, error)
-	List(ctx context.Context, orgID string, paginationData api.PaginationData, filterData api.TemplateFilterData) (api.TemplateCollectionResponse, int64, error)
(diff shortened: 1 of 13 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-5062

jlsherrill:
code looks good, testing now

jlsherrill (approved):

PR #934, opened 2025-01-07
HMS-5127: force task requeue on server exit

rverdile:
## Summary
This will make sure a running task requeues  on server exit, rather than possibly failing before having the chance to requeue

## Testing steps
1. You can apply this patch to help recreate the conditions of the bug
```
diff --git a/pkg/tasks/queue/queue.go b/pkg/tasks/queue/queue.go
index cbf0105..b97c5bd 100644
--- a/pkg/tasks/queue/queue.go
+++ b/pkg/tasks/queue/queue.go
@@ -11,7 +11,7 @@ import (
 	"github.com/google/uuid"
 )
 
-const MaxTaskRetries = 3 // Maximum number of times a task can be retried before failing
+const MaxTaskRetries = 100 // Maximum number of times a task can be retried before failing
 
 type Task struct {
 	Typename     string
diff --git a/pkg/tasks/worker/worker.go b/pkg/tasks/worker/worker.go
index 3fac3e8..f8754f4 100644
--- a/pkg/tasks/worker/worker.go
+++ b/pkg/tasks/worker/worker.go
@@ -84,6 +84,7 @@ func (w *worker) start(ctx context.Context) {
 	defer beat.Stop()
 
 	for {
+		time.Sleep(time.Second)
 		select {
 		case <-w.stopChan:
 			if w.runningTask.id != uuid.Nil {
```
2. With this PR, closing the server (CTRL+C) should always allow the running tasks to requeue
3. Without this PR (and with this patch), you should see the tasks fail and therefore fail to requeue as well

### DIFF
--- a/pkg/tasks/worker/worker.go
+++ b/pkg/tasks/worker/worker.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/content-services/content-sources-backend/pkg/config"
+	ce "github.com/content-services/content-sources-backend/pkg/errors"
 	m "github.com/content-services/content-sources-backend/pkg/instrumentation"
 	"github.com/content-services/content-sources-backend/pkg/models"
 	"github.com/content-services/content-sources-backend/pkg/tasks"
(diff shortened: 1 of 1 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-5127

xbhouse (approved):
looks good and works well! ack!

PR #936, opened 2025-01-09
HMS-5259: cleanup canceled tasks

rverdile:
## Summary
Adds canceled tasks to the query that cleans up all the old the tasks. Cleans up canceled with a queued_at time older than 20 days, not finished_at because canceled tasks do not have a finished_at time

## Testing steps
1. Create and then delete a repository so that the snapshot and/or update-latest-snapshot tasks get canceled
2. Manually edit the queued_at time so it is older than 20 days
3. Run nightly-jobs and the task should be cleaned up


jlsherrill:
https://issues.redhat.com/browse/HMS-5259

dominikvagner (approved):
looks good and works as expected! 🎉 👏🏼 
approved ✅

PR #939, opened 2025-01-14
HMS-5244: fix snapshot task cleanup

dominikvagner:
## Summary
This PR fixes up the snapshot task's cleanup that happens on context cancellation. This caused errors that were being reported by glitchtip. 
(`error deleting rpm repository versions: 400 Bad Request: [\"The repository version cannot be deleted because it (or its publications) are currently being used to distribute content. Please update the necessary distributions first.\"]`) 

## Testing steps
1. Create a repo and let is snapshot.
2. Apply this git patch:
```
diff --git a/pkg/tasks/snapshot_helper.go b/pkg/tasks/snapshot_helper.go
--- a/pkg/tasks/snapshot_helper.go	(revision 8c9b3c2e8f371d28878423f47acd3f19039f46cb)
+++ b/pkg/tasks/snapshot_helper.go	(date 1736855224959)
@@ -111,7 +111,7 @@
 		return fmt.Errorf("unable to save distribution task href: %w", err)
 	}
 
-	return nil
+	return context.Canceled
 }
```
3. Update the created repo and create one more.
4. Let them snapshot and verify there are no errors in the logs and that the first repo has only one snapshot.
5. Revert the change made in the 2. step.
6. Trigger snapshots on the 2 created repos and verify that after the snapshot tasks complete their status is valid.


jlsherrill:
https://issues.redhat.com/browse/HMS-5224

jlsherrill:
https://issues.redhat.com/browse/HMS-5244

pkg/tasks/snapshot_helper.go
```diff
@@ -97,33 +102,79 @@ func (sh *SnapshotHelper) Run(versionHref string) error {
 	if err != nil {
 		return err
 	}
+	err = sh.payload.SaveSnapshotIdent(snap.UUID)
+	if err != nil {
+		return fmt.Errorf("unable to save snapshot ident: %w", err)
+	}
```
rverdile, line 108: It's a little confusing, but the SnapshotIdent is set previously and it's actually not the snapshot UUID, just a random UUID, because the snapshot object has not been created yet.

I see that you're setting it so that the snapshot can be deleted, which I think makes sense. To avoid conflating the two values, it probably makes sense to make a new "SnapshotUUID" field for the purpose of tracking snapshot creation?

pkg/tasks/snapshot_helper.go
```diff
@@ -97,33 +102,79 @@ func (sh *SnapshotHelper) Run(versionHref string) error {
...
+	err = sh.payload.SaveDistributionTaskHref(snap.DistributionHref)
+	if err != nil {
+		return fmt.Errorf("unable to save distribution task href: %w", err)
+	}
+
```
rverdile, line 113: the distribution href is saved earlier in the function, so I think it's okay to remove this

pkg/tasks/snapshot_helper.go
dominikvagner: yeah, the `SnapshotIdent` confused me a little, but it wasn't used anywhere and I thought it should be the UUID so I "repurposed" it, but you are right it would be better to use a different separate field 😅 
changed it 👍🏼

pkg/tasks/snapshot_helper.go
dominikvagner: oops, removed 🫡

rverdile (commented):
looks good but needs rebase!

rverdile (approved):

PR #941, opened 2025-01-14
HMS-5056: add layered RH repos with feature name

xbhouse:
## Summary

- Adds layered RH repos (OCP and HA) to redhat_repos.json and allows importing of RH repos to be controlled by the feature name
- Adds feature_name as a new column on the repository_configurations table and imports this for RH repos. For custom repos, this value will be null

## Testing steps

1. Set `options.feature_filter` in config.yaml to include only the "RHEL-OS-x86_64" feature name (see config.yaml.example)
2. Run `make repos-import && go run cmd/external-repos/main.go nightly-jobs`. Only the 14 pre-existing RH repos should be imported and snapshotted
3. Update `options.feature_filter` to also include the "OPENSHIFT-OCP-x86_64" and "RHEL-HA-x86_64" features and restart the server
4. Run`make repos-import && go run cmd/external-repos/main.go nightly-jobs` again. You should see the 5 additional layered RH repos (OCP and HA) imported and snapshotted
5. If `options.feature_filter` is not set in config.yaml, the default ("RHEL-OS-x86_64") should be used and only the pre-existing RH repos should be imported and snapshotted


jlsherrill:
https://issues.redhat.com/browse/HMS-5056

pkg/external_repos/redhat_repos.go
```diff
@@ -27,6 +27,7 @@ type RedHatRepo struct {
 	Selector            string `json:"selector"`
 	GpgKey              string `json:"gpg_key"`
 	Label               string `json:"content_label"`
+	FeatureName         string `json:"feature_name"`
```
jlsherrill, line 30: It might be helpful to go ahead and add the ability to update the feature if it changes, i think its just adding it to the list here: https://github.com/content-services/content-sources-backend/blob/main/pkg/dao/repository_configs.go#L1281

jlsherrill: oh you know, i realized its not actaully importing the feature name into the db.  mind adding that? :)

db/migrations/20250115112007_add_feature_name_to_repo_configs.down.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+ALTER TABLE repository_configurations DROP COLUMN feature_name;
+
+COMMIT;
```
jlsherrill, line 5: I'd add an index on this column, as we'll be querying it in the future (and might as well do it in the same migration that adds it).

jlsherrill:
small comment, otherwise looks good!

jlsherrill (approved):

PR #942, opened 2025-01-16
HMS-5292: Add list subs-as-features

rverdile:
## Summary
Adds `GET  /admin/features`, which lists the features from the subs-as-features API.

Returns `{ features: ["feature1", "feature2", ...] }`

## Testing steps
1. Grab the certs you need
2. Make a GET request to `/admin/features` and you should see the features listed


pkg/config/certificates.go
```diff
@@ -0,0 +1,151 @@
...
+)
+
+type CertUser interface {
+	ClientCert() string
+	ClientKey() string
+	CACert() string
+	ClientCertPath() string
+	ClientKeyPath() string
+	CACertPath() string
+}
```
rverdile, line 19: I pushed this as a second commit because it over-complicate things a little, but I was trying to make using certs a little more streamlined in the future. Open to thoughts

jlsherrill:
https://issues.redhat.com/browse/HMS-5292

pkg/config/certificates.go
xbhouse: +1, i think this is a great idea!

pkg/config/certificates.go
```diff
@@ -0,0 +1,151 @@
...
+		transport.TLSClientConfig = tlsConfig
+	}
+	return transport, nil
+}
+
+func certPool(caCert []byte) (*x509.CertPool, error) {
+	pool := x509.NewCertPool()
+	ok := pool.AppendCertsFromPEM(caCert)
+	if !ok {
+		return nil, fmt.Errorf("could not parse candlepin ca cert")
```
xbhouse, line 94: small suggested change to the error message since i think technically we're only dealing with candlepin ca certs here now, but will likely use these methods for more in the future?

```suggestion
		return nil, fmt.Errorf("could not parse ca cert")
```

pkg/config/config.go
```diff
@@ -96,6 +97,16 @@ type Candlepin struct {
 	DevelOrg   bool   `mapstructure:"devel_org"` // For use only in dev envs
 }
 
+type SubsAsFeatures struct {
```
xbhouse, line 100: i'm wondering if we might need to have fields like these set in the deployment.yaml for subs_as_features too, i *think* the secretKeyRefs would be the same:
```
  - name: CLIENTS_CANDLEPIN_SERVER
    value: ${CLIENTS_CANDLEPIN_SERVER}
  - name: CLIENTS_CANDLEPIN_CLIENT_CERT
    valueFrom:
      secretKeyRef:
        name: content-sources-candlepin
        key: cert
        optional: true
  - name: CLIENTS_CANDLEPIN_CLIENT_KEY
    valueFrom:
      secretKeyRef:
        name: content-sources-candlepin
        key: key
        optional: true
  - name: CLIENTS_CANDLEPIN_CA_CERT
    valueFrom:
      secretKeyRef:
        name: content-sources-candlepin
        key: ca
        optional: true
```

pkg/config/certificates.go
rverdile: agreed!

pkg/config/config.go
rverdile: yes! thanks

pkg/admin_client/client.go
```diff
@@ -0,0 +1,83 @@
+package admin_client
```
xbhouse, line 1: what do you think about calling this features_client? or are you anticipating using this for more admin only (and not features-related) things?

pkg/config/config.go
xbhouse: oh we may need the parameter set too at the bottom of the deployment.yaml:

```
- name: CLIENTS_FEATURE_SERVICE_SERVER
  value: 'https://subscription.stage.api.redhat.com/svcrest/subscription/v5/features'
``` 

though the endpoint to query whether a user has access to certain features is different: https://subscription.stage.api.redhat.com/svcrest/subscription/v5/featureStatus. what do you think about setting this value to https://subscription.stage.api.redhat.com/svcrest/subscription/v5 so we can use it for your API and to query the other endpoint?

pkg/admin_client/client.go
rverdile: My thinking was it could be used for things non feature related since it's just making http calls (without a generated binding like the other clients). I think I'll know more about how to organize things with the next part of this

xbhouse: gotcha, makes sense!

rverdile: I think it does make sense to rename this for now, so I'm going to push that as well

xbhouse (approved):
nice job! lgtm!

PR #944, opened 2025-01-17
Build: speed up integration tests

jlsherrill:
## Summary

* By moving the wait for pulp to after candlepin, we hopefully don't have to wait as long
* unsure if we realy need these 'sleep 30' statements, lets try to get rid of them


jlsherrill:
one more time to see if we get a 3rd pass in a row

/retest

xbhouse (approved):

PR #949, opened 2025-01-22
HMS-5380: create_upload fixes

xbhouse:
## Summary

Fixes some issues when creating an upload:
- changes artifact href to be a string instead of null if empty (this is blocking IQE from calling this api)
- rejects request if chunk_size is 0 

## Testing steps

1. Artifact href should be set to an empty string if it doesn't exist
2. Request should be rejected if chunk size is 0 or less


jlsherrill:
https://issues.redhat.com/browse/HMS-5380

mayurilahane:
not facing the issue now and was able to create the upload 

image

mayurilahane:
When chunk size is "0" it is giving error as expected 

image

mayurilahane:
ACK

Andrewgdewar (approved):

PR #951, opened 2025-01-23
HMS-5293: add list feature content

rverdile:
## Summary
This adds an API to list the content if a feature from the feature service API

## Testing steps
1. Add the correct certs path to the clients.feature_service option in the config.yaml
2. Use the `GET /api/content-sources/v1/admin/features/` endpoint list the features
3. Choose one and use the name with `GET /api/content-sources/v1/admin/features/:name/content/` to list the content
4. Check that the fields match what is described by the JIRA ticket

### DIFF
--- a/pkg/feature_service_client/features.go
+++ b/pkg/feature_service_client/features.go
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"strings"
 
+	caliri "github.com/content-services/caliri/release/v4"
 	"github.com/content-services/content-sources-backend/pkg/api"
 	"github.com/content-services/content-sources-backend/pkg/cache"
 	"github.com/content-services/content-sources-backend/pkg/config"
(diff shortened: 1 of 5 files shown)


pkg/feature_service_client/features.go
```diff
@@ -66,3 +69,78 @@ func (fs featureServiceImpl) ListFeatures(ctx context.Context) (FeaturesResponse
...
+		}
+	}
+	return version
+}
+
+func getArchFromArches(arches string) string {
+	if strings.Contains(arches, "x86_64") {
+		return "x86_64"
+	}
+	return arches
```
jlsherrill, line 145: probably want to check for   aarch64 too?

jlsherrill:
overall looked good.  There are some issues with the older release (such as RHEL 4,5.etc..) with distribution & arch not being correct,  but IDK if need to worry about it right now.  

I think we'll probably just wanna adjust the logic as we go based on what we need.   It may change some with RHEL 10 and we may need to adjust regardless

pkg/feature_service_client/features.go
rverdile: good point! fixed

PR #958, opened 2025-01-29
HMS-5379: add metrics for templates

dominikvagner:
## Summary
This PR adds new Prometheus metrics that are gonna be used in our grafana dashboard.
```python
# HELP content_sources_templates_age_average Average age (days between the set template date and now) of templates.
content_sources_templates_age_average 
# HELP content_sources_templates_count Total number of templates.
content_sources_templates_count 
# HELP content_sources_templates_updated_in_last_24_hour_count Number of templates that have been updated in the last 24 hours.
content_sources_templates_updated_in_last_24_hour_count 
# HELP content_sources_templates_use_date_count Number of templates which have a set date.
content_sources_templates_use_date_count 
# HELP content_sources_templates_use_latest_count Number of templates which are set to use latest.
content_sources_templates_use_latest_count 
# HELP content_sources_task_pending_time_average_by_type Average pending time of the tasks by their type.
content_sources_task_pending_time_average_by_type{task_type="add-uploads-repository"}
content_sources_task_pending_time_average_by_type{task_type="delete-repository-snapshots"}
content_sources_task_pending_time_average_by_type{task_type="delete-snapshots"}
content_sources_task_pending_time_average_by_type{task_type="delete-templates"}
content_sources_task_pending_time_average_by_type{task_type="introspect"}
content_sources_task_pending_time_average_by_type{task_type="snapshot"}
content_sources_task_pending_time_average_by_type{task_type="update-latest-snapshot"}
content_sources_task_pending_time_average_by_type{task_type="update-repository"}
content_sources_task_pending_time_average_by_type{task_type="update-template-content"}
```
## Testing steps
1. Verify that the newly added metrics are showing up on our  `:9000/metrics` endpoint.
2. Create some templates which are using latest (at least 1) and have set dates (at least two, with different dates).
3. Verify that the `*_count` metrics have increased/changed accordingly.
4. Verify that the `content_sources_templates_age_average` is the average age of templates you created.
5. Run `make repos-import` and nightly jobs.
6. Wait for the pending time metrics to show up in the output of the metrics output.

\
_(tip: if you have the `bat` tool installed, you can pipe the output from that endpoint into `bat` and when you specify the language as Python, it will be a lot more readable, ex.: `http :9000/metrics | bat --language=python -P -p`)_ 😅

### DIFF
--- a/pkg/dao/metrics.go
+++ b/pkg/dao/metrics.go
@@ -178,3 +178,51 @@ func (d metricsDaoImpl) RHReposSnapshotNotCompletedInLast36HoursCount(ctx contex
 
 	return output + outputTaskless
 }
+
+func (d metricsDaoImpl) TemplatesUpdateTaskPendingTimeAverage(ctx context.Context) float64 {
+	var output float64 = -1
+	d.db.WithContext(ctx).
+		Model(&models.TaskInfo{}).
+		Select("coalesce(avg(extract(epoch from now() - queued_at)), 0.0) as pending_task_avg").
+		Where("status = ? and type = ?", config.TaskStatusPending, config.UpdateTemplateContentTask).
+		Pluck("pending_task_avg", &output)
+	return output
+}
+
+func (d metricsDaoImpl) TemplatesUseLatestCount(ctx context.Context) int {
+	var output int64 = -1
+	d.db.WithContext(ctx).
+		Model(&models.Template{}).
+		Where("use_latest is True").
+		Count(&output)
+	return int(output)
+}
+
+func (d metricsDaoImpl) TemplatesUseDateCount(ctx context.Context) int {
+	var output int64 = -1
+	d.db.WithContext(ctx).
+		Model(&models.Template{}).
+		Where("use_latest is False").
+		Count(&output)
+	return int(output)
+}
+
+func (d metricsDaoImpl) TemplatesUpdatedInLast24HoursCount(ctx context.Context) int {
+	var output int64 = -1
+	d.db.WithContext(ctx).
+		Model(&models.Template{}).
+		Where("created_at >= now() - interval '1' day").
+		Or("updated_at >= now() - interval '1' day").
+		Count(&output)
+	return int(output)
+}
+
+func (d metricsDaoImpl) TemplatesAgeAverage(ctx context.Context) float64 {
+	var output float64 = -1
+	d.db.WithContext(ctx).
+		Model(&models.Template{}).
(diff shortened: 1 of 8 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-5379

pkg/dao/metrics.go
```diff
@@ -178,3 +178,51 @@ func (d metricsDaoImpl) RHReposSnapshotNotCompletedInLast36HoursCount(ctx contex
...
+
+func (d metricsDaoImpl) TemplatesUpdateTaskPendingTimeAverage(ctx context.Context) float64 {
+	var output float64 = -1
+	d.db.WithContext(ctx).
+		Model(&models.TaskInfo{}).
+		Select("coalesce(avg(extract(epoch from now() - queued_at)), 0.0) as pending_task_avg").
+		Where("status = ? and type = ?", config.TaskStatusPending, config.UpdateTemplateContentTask).
+		Pluck("pending_task_avg", &output)
+	return output
+}
```
rverdile, line 190: There's an existing method for getting the average pending time of all tasks, I wonder if that method can be updated to include an optional filter to return a particular task type? That would make it easy to do this in the future for a different task type.

pkg/instrumentation/metrics.go
```diff
@@ -45,6 +51,12 @@ type Metrics struct {
 	OrgTotal                                       prometheus.Gauge
 	RHCertExpiryDays                               prometheus.Gauge
 	RHReposSnapshotNotCompletedInLast36HoursCount  prometheus.Gauge
+	TemplatesUpdateTaskPendingTimeAverage          prometheus.Gauge
```
rverdile, line 54: I could see this being a GaugeVector of pending times, partitioned by the task type. That may simplify doing this again for another task type. Not sure if it would work, but might be worth trying?

dominikvagner: Good idea, I like it 😁 Implemented it so have a look again at how it is now. 😅
Right now it does it for all task types, which can be changed or left as is, since it's still only one query.

pkg/dao/metrics.go
dominikvagner: Changed this method according to the other comment, and we still need to have 2, one for the total average and the per type averages.

pkg/instrumentation/metrics.go
rverdile: good idea!

rverdile: just gonna test now

rverdile (approved):
nice job!

PR #973, opened 2025-02-11
Build: use full path for compose files

jlsherrill:
## Summary

specifies full path to compose files to work around podman-compose bug

## Testing steps

make compose-clean compose-up


Andrewgdewar (approved):
Still works on mac! Ack from me!

dominikvagner (approved):
awesome, works with updated `podman-compose` _(1.3.0-2)_ on fedora _(41)_ 🎉 
thanks! 🫶

PR #983, opened 2025-02-17
HMS-5477: fix rh repo not updated for use-latest

rverdile:
## Summary
Red Hat repos within "use-latest" templates were not being updated on new snapshots because the template List query filters by org ID. When red hat repos are snapshotted nightly, an org ID of "-1" is used, so no templates would be returned. This adds a second query specifically to get all the templates containing the updated red hat repo.

## Testing steps
Testing this manually would be hard. The integration test covers it. I think the best you could do outside of stage is mimic the integration test manually


jlsherrill:
https://issues.redhat.com/browse/HMS-5477

test/integration/update_latest_snapshot_test.go
```diff
@@ -127,6 +131,75 @@ func (s *UpdateLatestSnapshotSuite) TestUpdateLatestSnapshot() {
...
+	snapUUID := s.snapshotAndWait(taskClient, *repoResp, uuidStr, true)
+
+	// Run update-latest-snapshot and verify template is using second snapshot
+	s.updateLatestSnapshotAndWait(config.RedHatOrg, repoResp.UUID)
+
+	tempResp, err = s.dao.Template.Fetch(ctx, orgID, tempResp.UUID, false)
+	assert.NoError(s.T(), err)
+	require.NotNil(s.T(), tempResp.Snapshots)
+	require.NotNil(s.T(), tempResp.Snapshots[0])
+	assert.Equal(s.T(), snapUUID, tempResp.Snapshots[0].UUID)
```
jlsherrill, line 200: Its probably worth adding a fetch of the actual rpm from the 2nd snapshot  (or listing of a letter directory that wouldn't be there in the first snapshot) to verify pulp was updated properly.

jlsherrill:
i actually tested this by:

* commenting the CA option out on remote creation
*  editing the red_hat.json file to only include 1repo and set the url to a repo I can control
* load the repos, snapshot the url
* create a template set to latest with this 'red hat' repo in it
* curl the template url and see the initial content
* update my yum repo to have more packages
* clear the last_snapshot_task_uuid on the repo config & re-snapshot the repo 
* curl the template repo url again see the new content
it worked great!

just one comment and it looks like maybe an issue with the integration tests

rverdile:
> i actually tested this by:

good idea!

jlsherrill (approved):

PR #984, opened 2025-02-18
HMS-5526: ignore canceled ctx for failed_snapshot_count

rverdile:
## Summary
There are two small bugs fixed here

- For canceled snapshots, the failed_snapshot_count was not being updated because the DB query was using context that was canceled. This changes it to use context.Background().
- GetContentPath() in the pulp client was logging an error when it failed due to canceled context, but an error is expected if context is canceled
## Testing steps
1. In the UI, add and remove the popular repositories (EPEL)
2. After removing them, there will no longer be an error logged "Failed to update failed_snapshot_count: context canceled"


jlsherrill:
https://issues.redhat.com/browse/HMS-5526

xbhouse (approved):
lgtm!

PR #986, opened 2025-02-18
HMS-5527: add request org_id to log parse

jlsherrill:
## Summary

Pulp added the request org id from the identity header, this adds it to our log gathering

also fixes a zero timestamp we were seeing

## Testing steps

just a code review


jlsherrill:
https://issues.redhat.com/browse/HMS-5527

xbhouse (approved):

PR #990, opened 2025-02-20
Test: Use currents for tests storage and dashboard

marusak:
"successful" run in https://github.com/content-services/content-sources-backend/actions/runs/13440630833 - posted https://app.currents.dev/run/6bf46c5b6f64d970

### DIFF
--- a/_playwright-tests/playwright.config.ts
+++ b/_playwright-tests/playwright.config.ts
@@ -16,12 +16,12 @@ export default defineConfig({
   workers: 1,
   reporter: !!process.env.CI
     ? [
-        ['list'],
+        ['html', { outputFolder: 'playwright-report' }],
         [
           'playwright-ctrf-json-reporter',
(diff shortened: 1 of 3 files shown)


jlsherrill (approved):

_playwright-tests/playwright.config.ts
```diff
@@ -16,12 +16,12 @@ export default defineConfig({
   workers: 1,
   reporter: !!process.env.CI
     ? [
-        ['list'],
```
Andrewgdewar, line 19: Any reason not to leave the list so when it's running we can watch it click through?

Andrewgdewar (approved):
Small comment

_playwright-tests/playwright.config.ts
marusak: I did the same that we have in the frontend. And I don't see it is necessary, you get his "list" view: (as it is also live in currents.dev)
Screenshot from 2025-02-24 16-13-28

PR #995, opened 2025-02-26
HMS-5536: enhance external-repos snapshot command

dominikvagner:
## Summary
This PR enhances the external-repos snapshot command in few ways:
- By adding a `--force` flag to it, which when used will enqueue those tasks even if they were recently snapshoted
- Allowing it to be run on even on non-RH repos
- Making the command clean the the inputted URLs, so that the ending slash isn't required
- Making it enqueue an `introspect` task along with the `snapshot` and `update-latest-snapshot` tasks

## Testing steps
1. Import the RH Ansible repo (`OPTIONS_REPOSITORY_IMPORT_FILTER=small go run cmd/external-repos/main.go import`)
2. Test that you can snasphot it with the `snapshot-command` command (`go run cmd/external-repos/main.go snapshot https://cdn.redhat.com/content/dist/layered/rhel8/x86_64/ansible/2/os/`). You should see all 3 tasks being enqueued.
3. Do the same with the `--force` flag and another 3 tasks should be enqueued. Try this with and without the slash at the end of the URL.
4. Create a custom repo and verify it also work with it's URL.


jlsherrill:
https://issues.redhat.com/browse/HMS-5536

xbhouse (approved):
lgtm! nice work!!

PR #1003, opened 2025-03-03
Test: move all secrets to github from codebuild

xbhouse:
## Summary

Switches secrets to be referenced from Github instead of Codebuild

## Testing steps

- PW tests pass (triggered manually on this branch and passed [here](https://github.com/content-services/content-sources-backend/actions/runs/13638994392))
- No secrets being pulled from Codebuild (these look like `$FOO` instead of `${{ secrets.FOO }}`)


Andrewgdewar (approved):
Looks good! Thanks for this @xbhouse!

PR #1005, opened 2025-03-04
Test: add test describes to playwright tests

dominikvagner:
## Summary
This adds test describes/suite wrappers to all current PW test files and removes the weird 'Content >' prefix that they all had, we know those tests test our API, since they live in out repo.

## Testing steps
- All tests should pass (nothing has changed there) but the PW reports and traces should be better/cleaner.


marusak (approved):
Thanks!

PR #1009, opened 2025-03-06
HMS-5530: add a helper for cleaning up test resources

dominikvagner:
## Summary
This adds new helpers and a fixture for easier cleaning up of test resources and better test idem-potency handling. And also "migrates" all existing tests that create test data to use this approach. 🧹😌 

## Testing steps
- All tests should pass repeatedly and even with existing data that would cause conflicts on resource creation. (Conflicts on repo name/URL or template name) 🔁


_playwright-tests/tests/API/Test.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+import { expect } from '@playwright/test';
+import { cleanupRepositories, poll } from './apiHelpers';
+
+test.describe('Test', () => {
+  // --------------- Approach no. 1 ---------------
+  const testConf: TestConfig = {
+    repoNames: ['test-2', 'reponame'],
+    urls: ['https://fedorapeople.org/groups/katello/fakerepos/zoo/'],
+    // templatesNames: [...]
+  };
```
jlsherrill, line 12: i wonder if we could name this something to make it a bit more obvious why its there.  `cleanupConf`  or something like that?

dominikvagner: defo, I am not the best with naming stuff, and `cleanupConf`|`CleanupConfig` for the type works 👍🏼😄

_playwright-tests/tests/API/apiHelpers.ts
```diff
@@ -0,0 +1,69 @@
...
+    result = await fn();
+    await timer(interval);
+  }
+  return result;
+};
+
+// timer in ms, must await e.g. await timer(num)
+export const timer = (ms: number) => new Promise((res) => setTimeout(res, ms));
+
+export const cleanupRepositories = async (
```
xbhouse, line 30: if things fail during cleanup, it might be difficult to understand where or why with approach 1 (without the test.step). what do you think about adding a try/catch in the cleanup functions and/or some logging?

dominikvagner: the cleanup functions should never fail / throw exceptions as they are I think 🤔 but if they do, then explicitly ignoring all errors with try/catch could be a helpful 😅 
_and I improved them a bit since you commented so have a look 😇_

xbhouse (approved):
looks great and tests are green 🤩 awesome work!!

PR #1013, opened 2025-03-11
Build: add PR head refs to github checkout actions

dominikvagner:
## Summary
This adds PR head refs to all GH checkout actions that currently don't have those, this is needed so that the actions run on the same code you are pushing and they don't have different results.

The same thing needed to be done in our FE repo [here](https://github.com/content-services/content-sources-frontend/pull/457).

### DIFF
--- a/.github/workflows/content-sources-actions.yml
+++ b/.github/workflows/content-sources-actions.yml
@@ -19,6 +19,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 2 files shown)


jlsherrill:
we discussed this in standup.  We can't actually make this change because pull_request_target doesn't honor the pull request settings, so anyone external could open a PR and the tests will run.  This creates a way for anyone to steal the secrets :/   So we will have to switch back to pull_request which checks out the PR automatically using the checkout action as you'd expect.  I'll go ahead and close this, but let me know if you have any questions!

PR #1018, opened 2025-03-11
HMS-5617: package groups test

xbhouse:


_playwright-tests/tests/API/PackageGroups.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+import { expect } from '@playwright/test';
+import { randomName } from './helpers/repoHelpers';
+import { poll } from './helpers/apiHelpers';
+import { randomUUID } from 'crypto';
+
+test.describe('Package groups', () => {
+  test('Search and list package groups', async ({ client }) => {
+    const repoUrl = 'https://content-services.github.io/fixtures/yum/comps-modules/v1/';
+    const repoName = randomName();
+    let repoUuid: string | undefined;
```
dominikvagner, line 18: nit: this could be declared just as a string and then expected and asserted as non-null right after repo creation, this would eliminate the need to use the bang (non-null assertion) several times after that 😅

dominikvagner (approved):
looks good, nice work on migrating and upgrading/extending the test! 👏🏼🎉 
_just one tiny nit 👀  if you wanted to address it feel free, but it's very minor so it's totally fine not to 😇_ 

approved! good job! ✅ ✨

PR #1024, opened 2025-03-12
HMS-5600: Add test for duplicated distro versions

dominikvagner:
## Summary
This adds a new PW test (migrates it from IQE) which tests that repository creation will ignore duplicate distribution versions.

## Testing steps
- All tests pass as expected.
- The new migrated test verifies the same functionality as the one in IQE.

### DIFF
--- a/_playwright-tests/tests/API/Repositories.spec.ts
+++ b/_playwright-tests/tests/API/Repositories.spec.ts
@@ -2,6 +2,7 @@ import { test } from './base_client';
 import {
   RepositoriesApi,
   GetRepositoryRequest,
+  ListRepositoriesRequest,
   ApiRepositoryResponse,
   ApiRepositoryValidationResponseFromJSON,
   type ValidateRepositoryParametersRequest,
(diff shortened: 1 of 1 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-5600

rverdile:
looks good but needs rebase!

rverdile (approved):
lgtm!

_playwright-tests/tests/API/Repositories.spec.ts
```diff
@@ -272,4 +272,55 @@ test.describe('Repositories', () => {
       expect(resp.raw.status).toBe(204);
     });
   });
+
+  test('Verify that repository creation ignores duplicate distro versions', async ({ client }) => {
+    const repoName = randomUUID();
+    const repoUrl = 'https://content-services.github.io/fixtures/yum/comps-modules/v2/';
```
swadeley, line 278: Hi, I see you are using a URL also used in another test and then you later delete all repos with this URL. at the start and end of the test. I think this might cause problems in future if tests run in parallel. How about using a two-part name so you just delete repos that match the prefix part of the name?

dominikvagner: good point, at the time of writing this test the repo helpers weren't there yet 😅 
rebased and made the test use a random URL and name, so this isn't an issue anymore 👍🏼

swadeley: Thank you

PR #1028, opened 2025-03-14
HMS-5395: add metrics for cert expiration

rverdile:
## Summary
- Adds gauge vector with "certificate_label" type to filter by 
- This replaces the existing red hat cdn metric, so eventually that can be removed. But not until the existing alert is removed.

## Testing steps
1. Configure all of the certs: cdn, candlepin, and feature service
2. `make run`
3. verify that `http:/localhost:9000/metrics` shows the CertificationExpiryDays metric, with each cert's days separated by the label

### DIFF
--- a/pkg/instrumentation/metrics.go
+++ b/pkg/instrumentation/metrics.go
@@ -33,6 +33,7 @@ const (
 	TemplatesUseDateCount                          = "templates_use_date_count"
 	TemplatesUpdatedInLast24HoursCount             = "templates_updated_in_last_24_hour_count"
 	TemplatesAgeAverage                            = "templates_age_average"
+	CertificateExpiryDays                          = "certificate_expiry_days"
 )
 
 type Metrics struct {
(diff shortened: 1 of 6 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-5395

pkg/instrumentation/metrics.go
```diff
@@ -132,7 +134,7 @@ func NewMetrics(reg *prometheus.Registry) *Metrics {
 			Namespace: NameSpace,
 			Name:      RHCertExpiryDays,
 			Help:      "Number of days until the Red Hat client certificate expires",
-		}),
+		}), // TODO remove in favor of CertificateExpiry Days
```
dominikvagner, line 137: are you planning on doing that as a later PR after setting up the alerts or is this something you forgot to do? 😅

dominikvagner (approved):
works as expected, will be nice to have this! 😌 
nice improvement to the certificates file by splitting it into smaller functions ✨ 

approved! good job 👏🏼
 _just one small question about the `TODO` comments in this commit 💭_

pkg/instrumentation/metrics.go
rverdile: yeah that will come later. the alert for this metric has to be removed first

dominikvagner: alrighty, thanks!

PR #1035, opened 2025-03-18
HMS-5397: add job for marking domains by labels 

dominikvagner:
## Summary
This adds a new job that sets a 'content-sources' label on all of our domains if they don't have it. Also makes the domain creation add this label as well. Will be useful for the pulp service team for differentiating our domains from others. 🍊 
For this an update of zest and pulp containers was needed, along with local pulp compose config change. 🔧 

## Testing steps
1. Create any repo with snapshots enabled.
2. List pulp domains and verify that the newly created domain has the `contentsources` label set.
3. Comment out the line **133**: `domain.SetPulpLabels(map[string]string{"contentsources": "true"})` in the file `pkg/clients/pulp_client/domains.go`, rebuild and restart the server.
4. Create any repo with snapshots enabled in a different org from the first one.
5. Check that it doesn't have the label.
6. Run the job `go run ./cmd/jobs/main.go set-domain-label`.
7. Check that it now has that label.


jlsherrill:
https://issues.redhat.com/browse/HMS-5397

deployments/jobs.yaml
```diff
@@ -53,3 +53,13 @@ objects:
...
+- apiVersion: cloud.redhat.com/v1alpha1
+  kind: ClowdJobInvocation
+  metadata:
+    labels:
+      app: content-sources-backend
+    name: set-domain-label-2025-03-18
+  spec:
+    appName: content-sources-backend
+    jobs:
+      - set-domain-label
```
jlsherrill, line 65: Lets copy this job to  jobs-stage.yaml and jobs-prod.yaml and then we can update app-interface to run a different stage and prod.  I'd just remove the rest of the jobs in this file for now

pkg/jobs/set_domain_label.go
```diff
@@ -0,0 +1,37 @@
...
+	for _, domain := range domains {
+		pulpHref, err := pulp_client.GetGlobalPulpClient().LookupDomain(ctx, domain.DomainName)
+		if err != nil || len(pulpHref) == 0 {
+			log.Error().Err(err).Msgf("failed to lookup pulp domain, for domain with name: %s", domain.DomainName)
+			continue
+		}
+
+		err = pulp_client.GetGlobalPulpClient().SetDomainLabel(ctx, pulpHref, "contentsources", "true")
+		if err != nil {
+			log.Error().Err(err).Msg("failed to update domain label")
```
jlsherrill, line 34: i'd probably include the domain name in the error here too

jlsherrill:
couple small things, but otherwise looks good!

pkg/jobs/set_domain_label.go
dominikvagner: added 🫡

deployments/jobs.yaml
dominikvagner: alrighty, moved it to `jobs-stage` and `jobs-prod` 😄

jlsherrill (approved):

PR #1038, opened 2025-03-19
HMS-5303: remove ansible repositories for snapshotting

jlsherrill:
## Summary

This adds support for removing Red Hat repos for syncing.  (Currently they stay for introspection).  

Some changes needed to support this

* within the delete-repository-snapshots task, move the candlepin processing prior to the pulp processing.  This is due to some models being deleted during the pulp processing (it also makes sense to do things in reverse upon deletion).
* Looking up the red hat content ids from their labels
* Looking up templates to remove the repo from across all orgs, not just templates with the same org as the repository

## Testing steps

setup, i'd recommend on a clean db.  On Main, run:

```
make compose-clean compose-up
make repos-import
```
start the server:
```
make run
```
snapshot all the repos:
```
go run cmd/external-repos/main.go process-repos
```

Create a template with RHEL 8, adding the ansible repo to the template.

Check the pulp db:
```
psql --host localhost --port 5432  -U pulp
select base_path from core_distribution;
```

look for:

```
 templates/341aaa4f-6050-4599-99d8-01e81bf4bf5c/content/dist/rhel8/8/x86_64/baseos/os
 templates/341aaa4f-6050-4599-99d8-01e81bf4bf5c/content/dist/rhel8/8/x86_64/appstream/os
 templates/341aaa4f-6050-4599-99d8-01e81bf4bf5c/content/dist/rhel8/8/x86_64/ansible/os
```

Check the candlepin db:
```
make db-cli-connect
\connect candlepin
 select * from cp_environment_content;
```
You should have 3 pieces of content.

Register RHEL 8 system and assign it to the template using https://github.com/content-services/content-sources-backend/blob/main/docs/register_client.md  and verify you can see the 3 repos on the client (via yum repolist).

Now checkout this pr, restart the server:
```
make run
```
Now reimport the red hat repos which will trigger the deletion:

```
make repos-import
```

re-run the pulp and candlepin queries and you should see the ansible repo distribution no longer there in pulp, and only 2 content objects in the environment in candlepin

On the client, run 'subscription-manager refresh', and verify that you only see 2 repos configured on the client

### DIFF
--- a/cmd/external-repos/main.go
+++ b/cmd/external-repos/main.go
@@ -184,7 +184,8 @@ func deleteNoLongerNeededRepos(ctx context.Context, daoReg *dao.DaoRegistry) err
 		"https://cdn.redhat.com/content/dist/layered/rhel8/aarch64/ansible/2/os/",
 	}
 	for _, url := range urls {
-		results, _, err := daoReg.RepositoryConfig.List(ctx, "0", api.PaginationData{Limit: 1}, api.FilterData{URL: url, Origin: config.OriginRedHat})
+		results, _, err := daoReg.RepositoryConfig.List(ctx, config.RedHatOrg, api.PaginationData{Limit: 1},
+			api.FilterData{URL: url, Origin: config.OriginRedHat})
 		if err != nil {
 			return fmt.Errorf("could not list repositories: %v", err)
 		}
@@ -206,6 +207,12 @@ func deleteNoLongerNeededRepos(ctx context.Context, daoReg *dao.DaoRegistry) err
 			_, err := c.Enqueue(task)
 			if err != nil {
 				return fmt.Errorf("could not enqueue task for repo deletion (%v): %v", url, err)
+			}
+
+			// Mark as not public, so orphan cleanup will clean it up later
+			err = daoReg.Repository.MarkAsNotPublic(ctx, url)
+			if err != nil {
+				return err
 			}
 		}
 	}
--- a/cmd/external-repos/main.go
+++ b/cmd/external-repos/main.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/content-services/content-sources-backend/pkg/api"
 	"github.com/content-services/content-sources-backend/pkg/clients/feature_service_client"
 	"github.com/content-services/content-sources-backend/pkg/clients/pulp_client"
 	"github.com/content-services/content-sources-backend/pkg/config"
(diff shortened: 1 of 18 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-5303

cmd/external-repos/main.go
```diff
@@ -158,16 +154,64 @@ func saveToDB(ctx context.Context, db *gorm.DB) error {
...
+	}
+	defer q.Close()
+	c := client.NewTaskClient(&q)
+
+	urls := []string{
+		"https://cdn.redhat.com/content/dist/layered/rhel8/x86_64/ansible/2/os/",
+		"https://cdn.redhat.com/content/dist/layered/rhel8/aarch64/ansible/2/os/",
+	}
+	for _, url := range urls {
+		results, _, err := daoReg.RepositoryConfig.List(ctx, "0", api.PaginationData{Limit: 1}, api.FilterData{URL: url, Origin: config.OriginRedHat})
```
xbhouse, line 187: this works so this is really just a curiosity -- why is org ID "0" here? because org doesn't matter in this case?

pkg/external_repos/redhat_repos.json
```diff
@@ -1,12 +1,4 @@
 [
-    {
```
xbhouse, line 2: should we remove the ansible repos from external_repos.json too?

xbhouse:
i couldn't test this fully with a registered client because apparently RHEL8 VMs [aren't compatible](https://github.com/lima-vm/lima/issues/841) with Mac M-series? 😞 @rverdile or @dominikvagner when either of you get a chance, could you verify the full flow works with a registered client?

this mostly looks good to me up to the point i could test though, just a couple questions 😄

cmd/external-repos/main.go
jlsherrill: yep, this is only for RH repos, so the org doesn't matter, but we have to pass something in, passing the RH org id might make more sense.  I'll do that

rverdile (commented):
I tested with a rhel8 VM, works!

jlsherrill:
Adding the removal of the introspected repo.  The way it works, is that we just mark it as not public.  Since there should be no repo configs, it should get cleaned up automatically.  You can test this by:

on main run:
``` make repos-import ```
```
make db-cli-connect

update repositories set created_at = '2024-01-01'; 
select url from repositories where url ilike '%ansible%'; 
```

should see the ansible repos

Switch to this PR, run:
```
make repos-import
make process-repos
```
that should trigger the cleanup.  Since the repo has no repo-configs, is not public, and older than a week, it should be deleted:

```
select url from repositories where url ilike '%ansible%'; 
```

xbhouse (approved):
lgtm! nice job!!

PR #1041, opened 2025-03-23
HMS-5570: delete upload entries on failed upload > artifact conversion

xbhouse:
## Summary

- Adds a DeleteUpload method
- If an upload fails to be converted to an artifact, the upload record is deleted in our db so the user can try to upload the rpm again

## Testing steps

1. Checkout the main branch
2. Create an upload, upload the same chunk twice, and finish the upload using these endpoints:
    - `POST /repositories/uploads/`
    - `POST /repositories/uploads/:upload_uuid/upload_chunk/`
    - `POST /repositories/:uuid/add_uploads/`
3. The add_uploads call should fail and you shouldn't be able to create another upload
4. Checkout this PR, and call add_uploads again. It should still fail but you should be able to create another upload and start over


jlsherrill:
https://issues.redhat.com/browse/HMS-5570

rverdile (approved):
nice job!!

PR #1046, opened 2025-03-28
HMS-5565: fix admin task fetch parsing error

rverdile:
## Summary
Fixes bug where fetching a snapshot task resulted in a parsing error. The snapshot payload was incorrectly saving the distribution href instead of the distribution task href.

Also fixes issue where we were not polling the update distribution task from pulp

## Testing steps
1. Create a snapshot
2. Fetch that snapshot task from the admin tasks API: `/admin/tasks/<task_uuid>`
3. Previously, there would be a 500 because of a missing "logging_cid"


jlsherrill:
https://issues.redhat.com/browse/HMS-5565

xbhouse (approved):
nice job!!

PR #1056, opened 2025-04-02
HMS-5798: move helpers and fixtures to shared dir

dominikvagner:
## Summary
This moves and in some cases slightly polishes helpers, fixtures and the API client to a new directory which is going to contain any shared test utilities between our projects (through git submodules and sparse checkout).
Along with this a new `package.json` file was added to the to be shared directory which should contain any dependencies required by these utilities. This shared utilities library should then be included in `package.json` files of repos consuming it and also to their TS configs for compilation.

## Testing steps
All tests should pass as they did before and helpers/fixtures work in the same way.


jlsherrill:
https://issues.redhat.com/browse/HMS-5798

xbhouse (approved):
this looks good! tests pass locally (except for snapshot cleanup but that also fails for me on main so not related to this 😃)

mk/swag.mk
```diff
@@ -31,7 +31,7 @@ openapi-doc: install-swag ## Regenerate openapi json document and lint
 
 .PHONY: openapi-js
 openapi-js: 
-	$(DOCKER) run -v .:/backend openapitools/openapi-generator-cli:latest-release generate -i backend/api/openapi.json  -g typescript-fetch -o backend/_playwright-tests/tests/API/client
+	$(DOCKER) run -v .:/backend:z openapitools/openapi-generator-cli:latest-release generate -i backend/api/openapi.json  -g typescript-fetch -o backend/_playwright-tests/test-utils/src/client
```
marusak, line 34: why `:z`? :thinking:

_playwright-tests/yarn.lock
```diff
@@ -879,6 +879,13 @@ consola@^3.2.3, consola@^3.4.0:
   resolved "https://registry.yarnpkg.com/consola/-/consola-3.4.0.tgz#4cfc9348fd85ed16a17940b3032765e31061ab88"
   integrity sha512-EiPU8G6dQG0GFHNR8ljnZFki/8a+cQwEQ+7wpxdChl02Q8HXlwEZWD5lqAF8vC2sEC3Tehr8hy7vErz88LHyUA==
 
+"content-sources-test-utils@file:./test-utils":
```
marusak, line 882: I though this would also drop `pg` since you moved it, but I guess not?

marusak (approved):
Thank you! I have only two questions, but should be all good :)

mk/swag.mk
dominikvagner: this is needed to overcome a SELinux restriction where it would prevent the containers from acccessing host files without proper labeling/permissions 😅 
and `:z` is a mount/label that we use in our makefiles for mounting host files 😄

_playwright-tests/yarn.lock
dominikvagner: well the `pg` package was removed from the main package json, but it isn't removed because now it is being picked up as a dependency from the test utils 😄

PR #1057, opened 2025-04-03
HMS-5925: bypass feature service api for red hat org

xbhouse:
## Summary

Task workers in stage are crashlooping due to calling the feature service status API with an invalid org ID (-1). This adds a check to bypass that API if using an org ID of -1 (which is only used to delete unneeded repos). 

## Testing steps

1. Run `make repos-import` when the feature service is enabled. Command should succeed and repos should be imported as normal
2. Layered repos (OCP and HA) should be present when listing RH repos or viewing RH repos in the UI if using an account that has access to these repos
3. Previously `make repos-import` failed when the feature service is enabled

### DIFF
--- a/pkg/dao/repository_configs.go
+++ b/pkg/dao/repository_configs.go
@@ -385,15 +385,25 @@ func (r repositoryConfigDaoImpl) List(
 	var totalRepos int64
 	repoConfigs := make([]models.RepositoryConfiguration, 0)
 	var contentPath string
+	var filteredDB *gorm.DB
 
-	accessibleFeatures, err := r.fsClient.GetEntitledFeatures(ctx, OrgID)
-	if err != nil {
-		return api.RepositoryCollectionResponse{}, totalRepos, err
-	}
+	if OrgID == config.RedHatOrg {
+		// Bypass the feature service status API if listing repos with the RH org
+		var err error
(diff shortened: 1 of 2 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-5925

pkg/dao/repository_configs.go
```diff
@@ -385,15 +385,25 @@ func (r repositoryConfigDaoImpl) List(
...
 	repoConfigs := make([]models.RepositoryConfiguration, 0)
 	var contentPath string
+	var filteredDB *gorm.DB
 
-	accessibleFeatures, err := r.fsClient.GetEntitledFeatures(ctx, OrgID)
-	if err != nil {
-		return api.RepositoryCollectionResponse{}, totalRepos, err
-	}
+	if OrgID == config.RedHatOrg {
+		// Bypass the feature service status API if listing repos with the RH org
```
rverdile, line 391: would do you think about doing this check from the `GetEntitledFeatures` call function itself? This way we don't have to remember to check this in other places if we're using the red hat org

xbhouse: ah yep that's a much better idea 😄 just pushed that up

rverdile (approved):
looks good!

xbhouse:
merging to try and fix stage, playwright tests are failing due to a timeout and were not affected by this change

PR #1058, opened 2025-04-04
Test: re-add caching to buildspec

xbhouse:
## Summary

The build-playwright-test job has been taking a bit longer than 30 minutes (about 40 minutes) so this adds caching back to the buildspec to see if this speeds it up a bit

## Testing steps

build-playwright-test finishes, codebuild doesn't time out

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -23,13 +23,13 @@ phases:
     commands:
       - echo Entered the post_build phase...
       - echo Build completed on `date`
-# cache:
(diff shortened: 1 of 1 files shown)


marusak:
What was the time difference with and without this change?

xbhouse:
> What was the time difference with and without this change?

@marusak this doesn't seem to have helped :/ both with and without this change the job took about 39 minutes

PR #1059, opened 2025-04-04
Test: fix name creation in repo tests

xbhouse:
## Summary

We are seeing 403s in the integration repo [test runs](https://github.com/content-services/content-sources-playwright/actions/runs/14269364245/job/39998792719#step:10:119) from these 2 tests when they call the cleanup helper, likely because we were passing a function instead of a string. This fixes the name creation to call the function before passing it to the helper.


swadeley (approved):
ACK

xbhouse:
merging to try to fix integration suite

PR #1068, opened 2025-04-08
HMS-5802: add lifecycle dates to package sources

rverdile:
## Summary
Adds the lifecycle dates from the roadmap API to the package sources in RPM search.

Fixes issue where we were returning older versions of module streams in package sources. Only the latest version is needed.

## Testing steps
1. run `make repos-import` and `go run cmd/external-repos/main.go process-repos` to make sure you have the RHEL 8 and RHEL 9 appstreams repos snapshotted
2. In your config.yaml, set the new fields under client
```
  roadmap:
    server: https://console.stage.redhat.com/api/roadmap/v1
    username:
    password:
    proxy:
```
4. Search RPMs (/rpms/names) and RPM snapshots (/snapshots/rpms/names)
```
RPMs:
{
  "search": "maven",
  "urls": ["https://cdn.redhat.com/content/dist/rhel9/9/x86_64/appstream/os/"],
  "include_package_sources": true
}

Snapshot:
{
  "search": "maven",
  "uuids": ["snapshot-uuid"],
  "include_package_sources": true
}

```
5. The response should include the "start_date" and "end_date" of the module streams listed in package_sources


jlsherrill:
https://issues.redhat.com/browse/HMS-5802

pkg/clients/roadmap_client/roadmap.go
```diff
@@ -0,0 +1,111 @@
...
+	var err error
+	var appStreamResp ModulesResponse
+	var body []byte
+
+	appstreams, err := rc.cache.GetRoadmapAppstreams(ctx)
+	if err != nil && !errors.Is(err, cache.NotFound) {
+		return ModulesResponse{}, statusCode, fmt.Errorf("error reading cache: %w", err)
+	}
+	if appstreams != nil {
+		err = json.Unmarshal(body, &appStreamResp)
```
xbhouse, line 62: should this be `err = json.Unmarshal(appstreams, &appStreamResp)` instead?

seeing an unmarshal error when hitting the search endpoint more than once, i think maybe because the first call doesn't retrieve anything the cache

xbhouse: also if we do retrieve something from the cache, i think we could probably return early here

rverdile: oh whoops, yeah. I also need to log an error when reading from the cache instead of returning

xbhouse (commented):
similar to Stephen's output, i'm seeing empty strings for start_date and end_date. maybe i'm testing with the wrong repos/rpms, did you see those populated?

rverdile:
It's only going to be populated for RHEL 8 and RHEL 9 because that is the scope of the roadmap api

xbhouse:
> It's only going to be populated for RHEL 8 and RHEL 9 because that is the scope of the roadmap api. I guess I can omit them if they're empty

hrmmm that's what i was testing with and they're empty 😕 

```
{
  "search": "maven",
  "urls": ["https://cdn.redhat.com/content/dist/rhel9/9/x86_64/appstream/os/"],
  "include_package_sources": true
}

[
    {
        "package_name": "maven",
        "summary": "Java project management and project comprehension tool",
        "package_sources": [
            {
                "type": "module",
                "name": "maven",
                "stream": "3.8",
                "context": "470dcefd",
                "arch": "x86_64",
                "version": "9040020240210002822",
                "description": "Maven is a software project management and comprehension tool. Based on the concept of a project object model (POM), Maven can manage a project's build, reporting and documentation from a central piece of information.",
                "start_date": "",
                "end_date": ""
            }
        ]
    },
    {
        "package_name": "maven-lib",
        "summary": "Core part of Maven",
        "package_sources": [
            {
                "type": "module",
                "name": "maven",
                "stream": "3.8",
                "context": "470dcefd",
                "arch": "x86_64",
                "version": "9040020240210002822",
                "description": "Maven is a software project management and comprehension tool. Based on the concept of a project object model (POM), Maven can manage a project's build, reporting and documentation from a central piece of information.",
                "start_date": "",
                "end_date": ""
            }
        ]
    },
    ...
]
```

pkg/api/rpms.go
```diff
@@ -105,6 +105,8 @@ type ModuleInfoResponse struct {
 	Version      string         `json:"version,omitempty"`     // Version of the module
 	Description  string         `json:"description,omitempty"` // Description of the module
 	PackageNames pq.StringArray `json:"-" gorm:"type:text"`    // Packages included in the module
+	StartDate    string         `json:"start_date"`
```
xbhouse, line 108: omitting might be a good idea, also i can't remember from the diagram.. should we be returning the dates if it's not a module? i guess omitting if empty will handle this if we shouldn't be

```
{
        "package_name": "389-ds-base",
        "summary": "389 Directory Server (base)",
        "package_sources": [
            {
                "type": "package",
                "start_date": "",
                "end_date": ""
            }
        ]
    },
```

rverdile: I believe we would also return it for packages, although that is for the next task

rverdile:
> hrmmm that's what i was testing with and they're empty 😕

It seems the output from the roadmap API has changed significantly since yesterday :) so I'll have to back and fix this. Going to put back into draft temporarily

rverdile:
Should be working now

xbhouse (approved):
works well and lgtm! nice job 😃

swadeley:
Hi

no more redundant keys with empty values:
```
In [2]: app.content_sources.rest_client.rpms_api.search_snapshot_rpms(dict(uuids=['3b84362c-4a0a-4078-85ef-efe5287fc4ae'], search='kangaroo',  include_package_sources=True))
<snip>
Out[2]: 
[{'package_name': 'kangaroo',
  'package_sources': [{'arch': 'noarch',
                       'context': 'deadbeef',
                       'description': 'A module for the kangaroo 0.3 package',
                       'name': 'kangaroo',
                       'stream': '0',
                       'type': 'module',
                       'version': '20180730223407'}],
  'summary': 'A dummy package of kangaroo'}]
````

Testing RHEL repo will have to be in stage.

Thank you.

PR #1069, opened 2025-04-09
Test: Change to centirepos

swadeley:
## Summary

I made a new batch of 100 repos without a number in the URL
The multirepos path with a rank number in URL causes problems when that number corresponds to a release version.

## Testing steps

tests pass

### DIFF
--- a/_playwright-tests/test-utils/src/helpers/repositories.ts
+++ b/_playwright-tests/test-utils/src/helpers/repositories.ts
@@ -1,13 +1,12 @@
 export const randomName = () => (Math.random() + 1).toString(36).substring(2, 6);
 
-export const rank = () => Math.floor(Math.random() * 10 + 1).toString();
 export const randomNum = () =>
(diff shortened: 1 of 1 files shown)


rverdile:
looks good but there's a linting issue

rverdile:
Doesn't break anything, but wanted to point out that some the repos are named with a leading 0 e.g. `repo03`. I'm not sure if you meant to include those

swadeley:
> Doesn't break anything, but wanted to point out that some the repos are named with a leading 0 e.g. `repo03`. I'm not sure if you meant to include those.

Hi, yes, I prefer to use leading zeros so that where the repo numbers are used in the names it helps with sorting sorting (looks better, easier to read, in pagination tests). I have now deleted repos 2 to 9 without leading zeros which were created by mistake.

Thank you

rverdile:
@swadeley once you update the URL here, should be good to go

rverdile (approved):
lgtm!

PR #1088, opened 2025-04-23
Build: upgrade to Go 1.23.6

rverdile:
## Summary
Go 1.23.6 is available in the ubi 9 container, so we can start using it


jlsherrill (approved):

jlsherrill:
@rverdile i think we can merge this 🥳

PR #1105, opened 2025-05-07
Build: separate delete-snapshots from process-repos

xbhouse:
## Summary

Separates the snapshot-cleanup job from process-repos so it can be run on its own schedule (this sets it to run daily). This may also help narrow down and reduce the errors being reported by glitchtip (https://issues.redhat.com/browse/HMS-5947)

## Testing steps

1. Create a repo, let it snapshot, then edit the repo URL to get another snapshot
2. Adjust the date of one of the snapshots to be a year old: `UPDATE snapshots SET created_at = (NOW() - CAST('365 days' AS INTERVAL)) WHERE uuid = '<snapshot_uuid>';`
3. Run `go run cmd/external-repos/main.go process-repos`. You shouldn't see any delete-snapshots tasks enqueued
4. Run `go run cmd/external-repos/main.go snapshot-cleanup`. This should start the delete-snapshots task

### DIFF
--- a/deployments/deployment.yaml
+++ b/deployments/deployment.yaml
@@ -875,6 +875,87 @@ objects:
                     name: content-sources-candlepin
                     key: key
                     optional: true
+        - name: snapshot-cleanup
+          schedule: "0 1 * * *"
+          suspend: ${{SUSPEND_CRON_JOB}}
+          concurrencyPolicy: "Forbid"
+          podSpec:
+            securityContext:
+              runAsNonRoot: true
+              runAsUser: 1001
+            image: ${IMAGE}:${IMAGE_TAG}
+            inheritEnv: true
+            command:
+              - /external-repos
+              - snapshot-cleanup
+            env:
+              - name: CLOWDER_ENABLED
+                value: ${CLOWDER_ENABLED}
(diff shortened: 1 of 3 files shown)


deployments/deployment.yaml
```diff
@@ -1058,4 +1139,8 @@ parameters:
   - name: CLIENTS_PULP_LOG_PARSER_S3_FILE_PREFIX
     description: path prefix to store pulp logs at in s3
   - name: CLIENTS_ROADMAP_SERVER
-    description: URL to the roadmap service (i.e. https://console.stage.redhat.com/api/roadmap/v1)
\ No newline at end of file
+    description: URL to the roadmap service (i.e. https://console.stage.redhat.com/api/roadmap/v1)
+  - name: SUSPEND_SNAPSHOT_CLEANUP
+    description: whether to not run the daily job to delete outdated snapshots
+    required: false
+    value: "false"
```
rverdile, line 1146: this is unused, and looks like the new job is using `SUSPEND_CRON_JOB` instead. I'm guessing this can just be removed

xbhouse: oops :) meant to update that value but yes just using the existing param is better!

PR #1109, opened 2025-05-14
HMS-5807: improve snapshots not found error for template create

rverdile:
## Summary
When creating a template, if the repository used does not have snapshots, the repo will be listed in the error response

## Testing steps
1. Create a repository for introspection only
2. Try to create a template using that repository. You will get a 404 with the error message "No snapshots found for the following repositories: ... "


jlsherrill:
https://issues.redhat.com/browse/HMS-5807

pkg/dao/templates.go
```diff
@@ -139,17 +139,23 @@ func (t templateDaoImpl) validateRepositoryUUIDs(ctx context.Context, orgId stri
...
-		Count(&snapshotCount)
+		Find(&repos)
 	if resp.Error != nil {
 		return fmt.Errorf("could not query repository uuids: %w", resp.Error)
 	}
-	if snapshotCount != int64(len(uuids)) {
-		return &ce.DaoError{NotFound: true, Message: "One or more repositories does not have a snapshot."}
+	for _, repo := range repos {
+		if repo.LastSnapshotUUID == "" {
+			missingSnapshotRepos = append(missingSnapshotRepos, repo.UUID)
```
xbhouse, line 153: i wonder if we should show the repo name(s) instead of uuid(s) in the error message? the uuids are helpful when using the API, but names would be more helpful in the UI. what do you think?

rverdile: sure!

xbhouse (approved):
lgtm!

rverdile:
@coderabbitai review

PR #1110, opened 2025-05-15
Build: debug snapshot deletion

xbhouse:

### DIFF
--- a/pkg/tasks/delete_snapshots.go
+++ b/pkg/tasks/delete_snapshots.go
@@ -137,6 +137,8 @@ func (ds *DeleteSnapshots) configurePulpClient() error {
 }
 
 func (ds *DeleteSnapshots) deleteOrUpdatePulpContent(snap models.Snapshot, repo api.RepositoryResponse, templateUpdateMap map[string]models.Snapshot) error {
+	logger := LogForTask(ds.task.Id.String(), ds.task.Typename, ds.task.RequestID)
+
 	for templateUUID, snaps := range templateUpdateMap {
 		distPath, distName, err := getDistPathAndName(repo, templateUUID)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/tasks/delete_snapshots.go
```diff
@@ -178,9 +180,34 @@ func (ds *DeleteSnapshots) deleteOrUpdatePulpContent(snap models.Snapshot, repo
...
+		distributions, err := ds.getPulpClient().ListDistributions(ds.ctx, ds.domainName)
+		if err != nil {
+			return err
+		}
+		publicationDistMap := make(map[string][]string)
+		logger.Debug().Msgf("Distribution count: %d", len(*distributions))
+		for i, dist := range *distributions {
+			logger.Debug().Msgf("Distribution #%d", i+1)
+			logger.Debug().Msgf("Distribution href: %v", *dist.PulpHref)
+			logger.Debug().Msgf("Distribution base path: %v", dist.BasePath)
```
jlsherrill, line 192: I think we only want to print this stuff for distributions that are including the publication we're deleting (via the version delete).  Otherwise it would print a LOT of output!

xbhouse: yep was thinking the same, it was a lot even locally :laughing: so only log if the dist.publication matches the snapshot's publication href or version href? the terms are confusing me

xbhouse: does this make more sense?

```
logger.Debug().Msgf("Checking distributions for publication: %v", snap.PublicationHref)
for _, dist := range *distributions {
	if dist.Publication.IsSet() {
		pub := dist.Publication.Get()
		if *pub == snap.PublicationHref {
			logger.Debug().Msg("Found distribution using deleted publication")
			logger.Debug().Msgf("  - pulp href: %v", *dist.PulpHref)
			logger.Debug().Msgf("  - base path: %v", dist.BasePath)
			logger.Debug().Msgf("  - publication: %v", *pub)
		}
	}
}
```

jlsherrill: I think thats right.  You might ensure pub isn't nil there, but otherwise looks right

pkg/tasks/delete_snapshots.go
```diff
@@ -178,9 +180,26 @@ func (ds *DeleteSnapshots) deleteOrUpdatePulpContent(snap models.Snapshot, repo
...
 	if err != nil {
+		distributions, err := ds.getPulpClient().ListDistributions(ds.ctx, ds.domainName)
+		if err != nil {
+			return err
+		}
+		logger.Debug().Msgf("Checking distributions for publication: %v", snap.PublicationHref)
+		for _, dist := range *distributions {
+			if dist.Publication.IsSet() {
+				pub := dist.Publication.Get()
+				if dist.PulpHref != nil && pub != nil && *pub == snap.PublicationHref {
```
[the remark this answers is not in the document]
rverdile: oh I missed that you checked for that, nevermind!

rverdile (approved):
lgtm!

PR #1112, opened 2025-05-16
HMS-8556: import different RHEL 10 gpg keys

jlsherrill:
## Summary

While RHEL 10 uses the same GPG keys as rhel 9, the format is different for one of the keys and image builder is failing on a step

## Testing steps

on main:
# make repos-import

switch to this PR:
# make repos-import
# make db-cli-connect

you should see the new gpg keys with this query:
```
select gpg_key from repository_configurations where name ilike '%10%' limit 1;
```

and the old ones with this query:
```
select gpg_key from repository_configurations where name ilike '%9%' limit 1;
```


rverdile (approved):
works!

jlsherrill:
merging, will verify in stage

PR #1114, opened 2025-05-19
HMS-8563: change to testing proxy for PW tests in actions

dominikvagner:
## Summary
This PR changes the playwright action to use a new custom testing proxy, which should make tests faster and more stable.
Also increases the instance size and parallelizes the different long
running setups.

## Testing steps
Playwright test should pass without loading errors.


marusak (approved):
Thank you!

PR #1115, opened 2025-05-21
HMS-5948: refactor external repos command

rverdile:
## Summary
Introduces cli library for external-repos commands, and moves the command implementations to a separate package.

Consolidates the cleanup jobs into the same cron. Pulp orphan cleanup is left on a separate weekly cadence, so pulp workers are not overwhelmed.

## Testing steps
1. Test all the external repos commands
2. `make build`
3. `./release/external-repos cleanup` to test the cleanup command. Use `cleanup --type snapshot,task` to specify a specific type. Or use `cleanup --exclude upload,repository` to exclude specific types.
4. Any commands that had arguments now use flags, e.g. `./external-repos snapshot --url <url>` instead of `./external-repos snapshot <url>`


jlsherrill:
https://issues.redhat.com/browse/HMS-5948

deployments/deployment.yaml
```diff
@@ -932,6 +951,8 @@ parameters:
     value: "0 8 * * 4"
   - name: NIGHTLY_CRON_JOB
     value: "0 0/1 * * *"
+  - name: DAILY_CRON_JOB
+    value: "0 1 * * *"
```
jlsherrill, line 955: would it make sense to call this something like  CLEANUP_CRON_EXP  ? .  I say this as the previous 'nightly' cron is no longer nightly in the different environments

rverdile: yeah, that's a good point

xbhouse (commented):
it looks like the playwright api test for snapshot cleanup is failing due to the command format change, do you mind updating that test too? i think you'd just need to change [this line](https://github.com/content-services/content-sources-backend/blob/main/_playwright-tests/tests/API/Snapshots.spec.ts#L141)

xbhouse (commented):
do you mind also updating the commands in the [introspection](https://github.com/content-services/content-sources-backend/blob/main/docs/workflows/introspection.md) and [snapshotting](https://github.com/content-services/content-sources-backend/blob/main/docs/workflows/snapshotting.md) docs? otherwise this looks really nice! great job :smile:

rverdile:
done!

xbhouse (approved):
awesome job!

PR #1117, opened 2025-05-29
HMS-5827: add community EPEL

rverdile:


jlsherrill:
https://issues.redhat.com/browse/HMS-5827

pkg/external_repos/community_repos.json
```diff
@@ -0,0 +1,23 @@
+[
+  {
+    "name": "EPEL 10 Everything x86_64",
+    "url": "https://dl.fedoraproject.org/pub/epel/10/Everything/x86_64/",
+    "distribution_version": "10",
+    "distribution_arch": "x86_64",
```
xbhouse, line 6: i'm seeing the `distribution_arch` in the api responses for community repos show up as `any`. i think this field needs to be `arch` to map them properly. but i also think using `distribution_arch` everywhere would be good too and cause less confusion!

xbhouse: oh also i think we'll need to import the epel repos for the playwright ui tests, do you mind adding a selector to each epel repo here so we can import them similarly to the small RH repo?

xbhouse:
the UI playwright tests are failing across multiple PRs :( seems like an issue reaching the backend? tests pass locally so could be related to CI setup, but the failures are unrelated to this PR

rverdile:
I think we may need to do a feature flag or something like that for the unified EPEL work, so that we are able to merge things over time. I'm going to put this back into draft for now while I play around with that.

rverdile:
Updated it so list/fetch won't show community repos without enabling the option in the config

PR #1122, opened 2025-06-04
Build: Adjust job name for transform-pulp-logs-fix

jlsherrill:
## Summary

The transform-pulp-logs-fix-2025-05 job didn't run in stage and I'm not sure why, a couple theories:

* the job name in deployment.yaml and jobs-stage.yaml were the same, maybe this isn't supported?
* The job is very similar to the other job (and cron jobs get created with a similar name), so make the start different

## Testing steps

none


jlsherrill:
oh yes, in this case its fixing it for may.... which isn't really the intention of those dates :)   will keep that in mind and adjust if i have to do a 2nd round

PR #1127, opened 2025-06-20
HMS-8709: update architecture docs

rverdile:
## Summary
Replaces old architecture diagram with an updated one. Removes the plantuml diagram because the diagram became too complicated to create via code.


jlsherrill:
https://issues.redhat.com/browse/HMS-8709

PR #1145, opened 2025-07-18
HMS-8810: add indexes and query improvements for dequeue

jlsherrill:
## Summary

Improves query by:

* Adding Partial indexes specifically for ready tasks filtering
* Adding Composite indexes for priority-based ordering
* Dependency indexes to optimize the expensive NOT IN subquery
* Optimized view using EXISTS instead of NOT IN

After merge, i'll monitor stage & production for a drop in cpu load as well as re-run the explain plan after its merged to stage.

## Testing steps

* tests pass

Explain plan on local dev env before the change:

```
explain UPDATE tasks
                SET token = 'ea6c185d-0770-4d2f-8309-e9558385b362', started_at = clock_timestamp(), status = 'running'
                WHERE id = (
                  SELECT id
                  FROM ready_tasks
                          -- use ANY here, because "type in ()" doesn't work with bound parameters
                          -- literal syntax for this is '{"a", "b"}': https://www.postgresql.org/docs/13/arrays.html
                  WHERE type = 'snapshot'
                  LIMIT 1
                  FOR UPDATE SKIP LOCKED
                )
                RETURNING  id, type, payload, queued_at, started_at, finished_at, status, error, org_id, object_uuid, object_type, token;
```

```
 Update on tasks  (cost=46.71..54.73 rows=1 width=62)
   InitPlan 2 (returns $3)
     ->  Limit  (cost=46.53..46.57 rows=1 width=1352)
           ->  LockRows  (cost=46.53..46.57 rows=1 width=1352)
                 ->  Subquery Scan on ready_tasks  (cost=46.53..46.56 rows=1 width=1352)
                       ->  LockRows  (cost=46.53..46.55 rows=1 width=1318)
                             ->  Sort  (cost=46.53..46.54 rows=1 width=1318)
                                   Sort Key: tasks_2.priority DESC, tasks_2.queued_at
                                   ->  Index Scan using tasks_multi_idx on tasks tasks_2  (cost=38.06..46.52 rows=1 width=1318)
                                         Index Cond: ((type)::text = 'snapshot'::text)
                                         Filter: ((started_at IS NULL) AND (((status)::text <> 'canceled'::text) OR (status IS NULL)) AND (NOT (hashed SubPlan 1)))
                                         SubPlan 1
                                           ->  Hash Join  (cost=10.61..37.86 rows=23 width=16)
                                                 Hash Cond: (task_dependencies.dependency_id = tasks_1.id)
                                                 ->  Seq Scan on task_dependencies  (cost=0.00..23.60 rows=1360 width=32)
                                                 ->  Hash  (cost=10.60..10.60 rows=1 width=16)
                                                       ->  Seq Scan on tasks tasks_1  (cost=0.00..10.60 rows=1 width=16)
                                                             Filter: (finished_at IS NULL)
   ->  Index Scan using tasks_pkey on tasks  (cost=0.14..8.16 rows=1 width=62)
         Index Cond: (id = $3)
(20 rows)
```

and after:

```
 Update on tasks  (cost=23.97..31.99 rows=1 width=62)
   InitPlan 1 (returns $4)
     ->  Limit  (cost=4.46..23.83 rows=1 width=1353)
           ->  LockRows  (cost=4.46..23.83 rows=1 width=1353)
                 ->  Subquery Scan on ready_tasks  (cost=4.46..23.82 rows=1 width=1353)
                       ->  LockRows  (cost=4.46..23.81 rows=1 width=1331)
                             ->  Nested Loop Anti Join  (cost=4.46..23.80 rows=1 width=1331)
                                   ->  Index Scan using idx_tasks_priority_queued on tasks t  (cost=0.12..8.14 rows=1 width=1319)
                                         Filter: ((started_at IS NULL) AND (((status)::text <> 'canceled'::text) OR (status IS NULL)) AND ((type)::text = 'snapshot'::text))
                                   ->  Nested Loop  (cost=4.33..15.65 rows=1 width=28)
                                         ->  Bitmap Heap Scan on task_dependencies td  (cost=4.21..14.35 rows=7 width=38)
                                               Recheck Cond: (task_id = t.id)
                                               ->  Bitmap Index Scan on idx_task_dependencies_task_id  (cost=0.00..4.21 rows=7 width=0)
                                                     Index Cond: (task_id = t.id)
                                         ->  Index Scan using idx_tasks_dependency_check on tasks dep  (cost=0.12..0.18 rows=1 width=22)
                                               Index Cond: ((id = td.dependency_id) AND (finished_at IS NULL))
   ->  Index Scan using tasks_pkey on tasks  (cost=0.14..8.16 rows=1 width=62)
         Index Cond: (id = $4)
(18 rows)
```

shows no seq scans and a much lower cost


jlsherrill:
https://issues.redhat.com/browse/HMS-8810

db/migrations/20241104115955_add_cancel_attempted_to_tasks.down.sql
```diff
@@ -1,5 +1,19 @@
...
+WHERE started_at IS NULL
+  AND (status != 'canceled' OR status is null)
+  AND id NOT IN (
+    SELECT task_id
+    FROM task_dependencies JOIN tasks ON dependency_id = id
+    WHERE finished_at IS NULL
+)
+ORDER BY priority DESC, queued_at ASC;
+
 COMMIT;
```
jlsherrill, line 19: This was required as the cancel_attempted column is part of the ready_tasks view.  It hadn't been hit before because teh ready_tasks view wasn't ever recreated  after this column was added.  

This means the view didn't actually include this column until this migration was added. 

I wonder if we should explicitly list the columns on the ready_task view to avoid this situation.  Thoughts?

jlsherrill: So the ready_tasks view creation would become:

```
CREATE OR REPLACE VIEW ready_tasks AS
SELECT id, org_id, token,type, payload, status, error, queued_at, started_at, finished_at, request_id, account_id, retries, next_retry_time, priority, object_uuid, object_type
FROM tasks t
```

we could probably even reduce the amount of columns in the list i would think

jlsherrill: sounds good to me!

jlsherrill: this will require two migrations, one switching to this query, and then once this is in production, one dropping the ready_tasks view.  I'll go ahead and open the 2nd PR and we'll just hold it until this one is promoted to prod

rverdile (approved):

PR #1157, opened 2025-07-25
Build: generate deployment.yaml from template

jlsherrill:


rverdile (commented):
I haven't tested it, but I like the idea. I say let's do it.

xbhouse:
+1 to this idea!

i did a small test to generate the deployment with a new common env var and noticed that it wasn't added to the parameters section, that's something we'll probably want to add. maybe automate it if possible or a validation step that every env var has a matching parameter definition?

PR #1165, opened 2025-08-04
HMS-8888: add repository partial update playwright test

rverdile:
## Summary
Migrates the IQE test for using PATCH with repositories

## Testing steps
Tests pass

### DIFF
--- a/_playwright-tests/tests/API/Repositories.spec.ts
+++ b/_playwright-tests/tests/API/Repositories.spec.ts
@@ -667,4 +667,71 @@ test.describe('Repositories', () => {
       expect(taskCreatedAt.getTime()).toBeGreaterThanOrEqual(currentTime.getTime());
     });
   });
+
+  test('Partial update repository', async ({ client, cleanup }) => {
+    const repoName = `patch-repository-${randomName()}`;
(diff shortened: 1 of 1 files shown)


_playwright-tests/tests/API/Repositories.spec.ts
```diff
@@ -530,4 +530,69 @@ test.describe('Repositories', () => {
...
+        apiRepositoryUpdateRequest: {
+          name: updatedName,
+          url: updatedUrl,
+        },
+      });
+
+      expect(updatedRepo.uuid).toBe(repo.uuid);
+      expect(updatedRepo.name).toBe(updatedName);
+      expect(updatedRepo.url).toBe(updatedUrl);
+      expect(updatedRepo.distributionArch).toBe('x86_64');
```
xbhouse, line 572: i would add assertions for the module_hotfixes and gpg_key here too just since you're updating them in the following test case

rverdile: they're not set during creation, so they'd just be the empty values, but yeah might as well

xbhouse (approved):
nice job! lgtm :)

PR #1175, opened 2025-08-14
HMS-8900: Test for Snapshotting of Red Hat Repositories

mayurilahane:
Adds a Playwright API test that verifies public and Red Hat repo's introspection.

### DIFF
--- a/_playwright-tests/tests/API/Repositories.spec.ts
+++ b/_playwright-tests/tests/API/Repositories.spec.ts
@@ -15,6 +15,8 @@ import {
   CreateSnapshotRequest,
   GetTaskRequest,
   ApiTaskInfoResponse,
+  PublicRepositoriesApi,
+  ApiPublicRepositoryResponse,
 } from 'test-utils/client';
 import {
   cleanupRepositories,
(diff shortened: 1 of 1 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-8900

xbhouse:
i may not be too clear on the goal of this test, but i think this might make more sense as an integration test to verify the stage repos. 

right now  we're only introspecting / snapshotting the small RH repo and the EPEL 10 repo in CI, so this test would really only verify that small RH repo. 

also the public repos' statuses aren't `Invalid` but they aren't `Valid` either because they haven't been introspected, and i'm not sure that we should introspect all of them in CI :thinking:

mayurilahane:
> i may not be too clear on the goal of this test, but i think this might make more sense as an integration test to verify the stage repos.
> 
> right now we're only introspecting / snapshotting the small RH repo and the EPEL 10 repo in CI, so this test would really only verify that small RH repo.
> 
> also the public repos' statuses aren't `Invalid` but they aren't `Valid` either because they haven't been introspected, and i'm not sure that we should introspect all of them in CI 🤔

privious card for this test 
https://issues.redhat.com/browse/HMS-2487

swadeley:
> > i may not be too clear on the goal of this test, but i think this might make more sense as an integration test to verify the stage repos.
> > right now we're only introspecting / snapshotting the small RH repo and the EPEL 10 repo in CI, so this test would really only verify that small RH repo.
> > also the public repos' statuses aren't `Invalid` but they aren't `Valid` either because they haven't been introspected, and i'm not sure that we should introspect all of them in CI 🤔
> 
> previous card for this test has this acceptance criteria
> 
> `A user with the snapshotting feature can list red hat repositories that are being snapshotted in the api & UI `
> 
> `A user with the snapshotting feature can list red hat repository snapshots in the api & UI`
> 
> https://issues.redhat.com/browse/HMS-2487
> 
> Hey @swadeley any thoughts ?

Hi, the Jira says "A user with the snapshotting feature can list", so we are testing that the user has permissions and tools to do things rather than testing the repos themselves. 
We can test against stage as we don't mind where the repos are, we just need to test the user has the ability to use the API and UI created in HMS-2243 to list repos. 
We only need to test one RHEL and one custom repo as we are not testing the validity of the repos or the backend code.
It makes sense to have the test in the Integration project so it can run nightly and just test repos that already exist in stage without per-test setup.

mayurilahane:
closing this because raised it here - https://github.com/content-services/content-sources-frontend/pull/633

PR #1182, opened 2025-08-19
HMS-9081: make db log level configurable

rverdile:
## Summary
- Makes the DB log level optionally configurable because the number of logs is freezing up my laptop
- Gorm now ignores "context canceled" error logs 

## Testing steps
1. Use this to test the context canceling. It lists repositories and the cancels it quickly: https://gist.github.com/rverdile/646a8576d3919f18ab1998f7cb2682cd
2. Without this PR, gorm will log the context is canceled
3. Change the db_level option under logging to change it from the global logging level

### DIFF
--- a/configs/config.yaml.example
+++ b/configs/config.yaml.example
@@ -45,6 +45,7 @@ tasking:
 logging:
   level: debug
   metrics_level: debug
+  db_level: debug
   console: True
   color: True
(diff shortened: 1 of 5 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-9081

PR #1204, opened 2025-09-17
Test: Fix template cleanup support suffix

swadeley:
## Summary

Make template cleanup accept prefix name

## Testing steps

check that template tests cleanup afterwards


dominikvagner (approved):
ahh, oops **my bad** 🤦🏼‍♂️ 
don't know why I used `in` here and the correct one in the other cleanup func 🫣

_the fact both of these exist, do something different and `for ... in ...` is used by several other languages is imho a language design flaw of typescript 🙄😅_

acked! thanks for fixing this! 🙇🏼‍♂️ ✅

PR #1205, opened 2025-09-17
HMS-9371: add debug logs for kessel api calls

rverdile:
## Summary
Getting a 401 but we're not logging any additional information


swadeley (approved):
ACK

PR #1206, opened 2025-09-17
HMS-8908: Add repository filtering test

mayurilahane:
## Summary

- Test filtering by architecture, version, UUID, URL and text search
- Validates single-value and multi-value filters

### DIFF
--- a/_playwright-tests/tests/API/RepositoryFilteringTest.spec.ts
+++ b/_playwright-tests/tests/API/RepositoryFilteringTest.spec.ts
@@ -1,0 +1,203 @@
+import { expect, test } from 'test-utils';
+import { RepositoriesApi, ApiRepositoryResponse, GetRepositoryRequest } from 'test-utils/client';
+import { poll, randomName, cleanupRepositories } from 'test-utils/helpers';
+
+/**
+ * Tests that repository listings can be filtered by various parameters including
+ * architecture, version, UUID, URL, and text search.
+ */
+
+test.describe('Repository Filtering Test', () => {
+  const createdRepos: ApiRepositoryResponse[] = [];
+
+  test('should filter repository listings with various parameters', async ({ client, cleanup }) => {
+    const baseRepoName = `Filter-Test-${randomName()}`;
+
+    await cleanup.runAndAdd(() =>
+      cleanupRepositories(
+        client,
+        baseRepoName,
+        'https://dl.fedoraproject.org/pub/epel/8/Everything/ppc64le/',
+        'https://dl.fedoraproject.org/pub/epel/9/Everything/x86_64/',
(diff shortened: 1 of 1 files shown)


_playwright-tests/tests/API/RepositoryFilteringTest.spec.ts
```diff
@@ -0,0 +1,264 @@
...
+      expect(version8RepoNames).toContain(`${baseRepoName}-3`);
+      expect(version8RepoNames).not.toContain(`${baseRepoName}-2`);
+
+      const availableVersion9Response = await new RepositoriesApi(client).listRepositories({
+        availableForVersion: '9',
+        origin: 'external',
+      });
+
+      const availableVersion9Names = availableVersion9Response.data?.map((repo) => repo.name) || [];
+      expect(availableVersion9Names).toContain(`${baseRepoName}-2`);
```
xbhouse, line 109: i'd suggest adding a check to each test to verify the number of repos in the response is equal to the count you expect, so it's clear that the other repos are filtered out

_playwright-tests/tests/API/RepositoryFilteringTest.spec.ts
```diff
@@ -0,0 +1,264 @@
...
+
+test.describe('Repository Filtering Test', () => {
+  const createdRepos: ApiRepositoryResponse[] = [];
+
+  test('should filter repository listings with various parameters including status', async ({
+    client,
+    cleanup,
+  }) => {
+    const baseRepoName = `Filter-Test-${randomName()}`;
+    const repositoriesApi = new RepositoriesApi(client);
```
xbhouse, line 18: this is used only in a few steps, for consistency i'd suggest either using it everywhere or removing it

_playwright-tests/tests/API/RepositoryFilteringTest.spec.ts
```diff
@@ -0,0 +1,264 @@
...
+      }
+    });
+
+    await test.step('Test single-value filters', async () => {
+      const x86Response = await repositoriesApi.listRepositories({
+        origin: 'external',
+        arch: 'x86_64',
+      });
+
+      const x86RepoNames = getRepoNames(x86Response);
```
xbhouse, line 88: this relies on keeping track of all the repo names and what the repo's properties are. is there a reason you've done it this way? we could instead (or also?) check the response only includes repos with whatever the requested filter is

mayurilahane: Yeah we should be consistant with other tests.
_It was an AI suggestion for simplified assertions and reusability so i kinda like the idea but now i think its not going to give an error if there is no value return for repo.name or other properties its just going to give an empty array and we want the test to fail if nothing is returned_ 

so changing it.

PR #1208, opened 2025-09-18
HMS-9066: Change go version to 1.24

mayurilahane:
## Summary
update: Change go version to 1.24

### DIFF
--- a/.github/workflows/content-sources-actions.yml
+++ b/.github/workflows/content-sources-actions.yml
@@ -21,7 +21,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v4
         with:
-          go-version: "1.23"
+          go-version: "1.24"
       - run: |
           make openapi
       - run: |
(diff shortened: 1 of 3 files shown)


rverdile (commented):
I would do a dependency bump here: `go get -u ./...`

mayurilahane:
> I would do a dependency bump here: `go get -u ./...`

Yeah i did that at very first but after rebasing with the main branch i think it changed the go.sum and go.mod file 
thanks for the reminder to do it again.

xbhouse:
do we need to also update the build image in build/Dockerfile to `quay.io/projectquay/golang:1.24`? or is `registry.redhat.io/ubi9/go-toolset:latest` ok?

marusak:
> i was thinking to update that too but docker file says it uses latest quay image so i did not bother it
https://github.com/content-services/content-sources-backend/blob/main/build/Dockerfile i think now its using 1.24.7
and i have mentioned 1.24 so it will be able to handle all the subversion /?

I believe @jlsherrill mentioned the other day in standup that we should change it to 1.24 to make sure it is locked to the same version. Isn't that so? :thinking:

jlsherrill:
yep!  all it is is changing the 'latest' tag here:

https://github.com/content-services/content-sources-backend/blob/main/build/Dockerfile#L7

to `1.24`

PR #1210, opened 2025-09-19
HMS-9072: fix update calls to pulp to handle 204 with no task

Starle21:
## Summary

**Pulp update calls now return 200 with no task**

Previously pulp always created an internal pulp task and returned the task href,
when there had been triggered an update call to domains or remote, irrespective if
the update was changing anything or not. 

From now on, when an update call to
- domain (UpdateDomainIfNeeded) or 
- remote (UpdateRmpRemote) does not change anything,

pulp does not create a task. 
This is indicated by 200 status code. 

If there is an actual change, pulp sends 202 with taskHref back. 

This commit updates the parts of code
that deal with these pulp functions to take this pulp api change into account.

Note that Pulp team has been waiting on content-sources to be updated first,
before they update Pulp. This means that zest client is still expecting to receive
a task href always, failing when it is empty. There is a temporary workaround
in the code, which will be removed after Pulp will get updated.

## Testing steps
Run the integration tests.

### DIFF
--- a/pkg/clients/pulp_client/domains.go
+++ b/pkg/clients/pulp_client/domains.go
@@ -48,9 +48,9 @@ func (r *pulpDaoImpl) LookupOrCreateDomain(ctx context.Context, name string) (st
 	}
 }
 
-// Updates a domain if that domain is using s3 and its storage configuration has changed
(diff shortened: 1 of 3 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-9072

PR #1215, opened 2025-09-23
Build: fix instructions for running go integration tests

xbhouse:
## Summary

Updates the IP address for the pulp.content host entry needed to run the go integration tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,11 +25,14 @@ Create a config file from the example:
 cp ./configs/config.yaml.example ./configs/config.yaml
 ```
 
-### Add pulp.content to /etc/hosts for integration tests and client access
+### Add pulp.content to /etc/hosts for Go integration tests and client access
 
 ```sh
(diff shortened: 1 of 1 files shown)


swadeley:
Hi @xbhouse ,

 is this to get Pulp content from stage?
 If so, would I not need the IP address from the VPN tunnel endpoint?

xbhouse:
@swadeley these instructions are for running the go integration tests (not playwright). the readme was wrong and was causing some confusion when tests would fail with connection errors

swadeley:
> @swadeley these instructions are for running the go integration tests (not playwright). the readme was wrong and was causing some confusion when tests would fail with connection errors.
> 
> i'll clarify this in the readme too!

OK, it is in the "Developing" section , but good to make it clear. Thank you.

swadeley (approved):
ACK

PR #1219, opened 2025-09-29
Build: Only recompile the main binary on make run

jlsherrill:
## Summary

After updating some code and re-running 'make run', the makefile would recompile every binary, which wasn't really necessary.  This limits it to only rebuilding content-sources.

## Testing steps

Modify some dao file or model
run 'make run'

look for:
```
go build  -o "/home/jlsherri/git/content-sources-backend/release/content-sources" "cmd/content-sources/main.go"
```

previously you'd see about ~5 of these.


xbhouse (approved):
lgtm!

PR #1224, opened 2025-10-01
HMS-8896: Verify same repo URL can be created in different orgs.

mayurilahane:
## Summary

Add test to ensure different user from different org can create repositories with identical URL

### DIFF
--- a/_playwright-tests/test-utils/src/helpers/poll.ts
+++ b/_playwright-tests/test-utils/src/helpers/poll.ts
@@ -1,3 +1,11 @@
+import { expect } from '@playwright/test';
+import {
+  RepositoriesApi,
+  GetRepositoryRequest,
+  ApiRepositoryResponse,
+  Configuration,
+} from '../client';
+
 // while condition is true, calls fn, waits interval (ms) between calls.
 // condition's parameter should be the result of the function call.
 /* eslint-disable @typescript-eslint/no-explicit-any */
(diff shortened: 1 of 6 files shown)


jlsherrill:
https://issues.redhat.com/browse/HMS-8896

_playwright-tests/tests/API/Repositories.spec.ts
```diff
@@ -742,4 +742,45 @@ test.describe('Repositories', () => {
       expect(updatedRepo.gpgKey).toBe('test-gpg-key');
     });
   });
+
+  test('Second user in other org can create same repo', async ({ client, cleanup }) => {
+    /**
+     * This test focuses on creating same repo in different accounts within different org.
+     * stable_sam_stage has this repo "https://dl.fedoraproject.org/pub/epel/10/Everything/x86_64/"
```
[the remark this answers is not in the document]
mayurilahane: Sure, lets do that !

_playwright-tests/test-utils/src/helpers/poll.ts
```diff
@@ -16,3 +24,23 @@ export const poll = async (
...
 export const sleep = (ms: number) => new Promise((res) => setTimeout(res, ms));
+
+/**
+ * Waits for a repository introspection to complete by polling the repository status.
+ * @param client - Configuration object for API client (optional, uses default if not provided)
+ * @param repoUuid - UUID of the repository to check
+ * @param interval - Polling interval in seconds (default: 10)
+ */
+export const waitForRepositoryIntrospection = async (
+  client: Configuration | undefined,
```
dominikvagner, line 35: this shouldn't be optional, not sure what you mean by it actually using `default` if not provided 🤔 
the client is being used in this helper so it is required, please change that 🙏🏼

_playwright-tests/test-utils/src/helpers/poll.ts
```diff
@@ -16,3 +24,23 @@ export const poll = async (
...
 // sleep in ms, must await e.g. await sleep(num)
 export const sleep = (ms: number) => new Promise((res) => setTimeout(res, ms));
+
+/**
+ * Waits for a repository introspection to complete by polling the repository status.
+ * @param client - Configuration object for API client (optional, uses default if not provided)
+ * @param repoUuid - UUID of the repository to check
+ * @param interval - Polling interval in seconds (default: 10)
+ */
+export const waitForRepositoryIntrospection = async (
```
dominikvagner, line 34: the name here is kinda incorrect, it doesn't wait for only the introspection, but while the status is pending, that can also mean snapshoting, so `waitWhileRepositoryIsPending` would be a better name 😅

dominikvagner (changes requested):
some more comments regarding the new changes 💭 

and also, the second commit seems to be addressing feedback from @marusak on the first one and it changes the same test, or fixes it rather, so those 2 should squashed into one 🧹😅

PR #1453, opened 2026-04-08
Test: make generated client use the PW requests 

dominikvagner:
## Summary
This changes the way how the generated typescript client, we use for
testing, sends the requests. Previously we used the default or undici
for this, but that meant we couldn't see those requests in the PW
traces, making it hard to debug.
We can utilize the built-in Playwright request fixture and API context
to do this, by providing a custom "FetchAPI" to the generated client.
Also makes the client dynamically pick-up fresh credentials per request
by default.

And makes a few smaller tweaks in the first 2 commits:
> This fixes an issue with a wrong describe used in `
> auth.setup.ts`,
> changes the project name of API tests to something appropriate and
> quiets unnecessary logs.
> 
> This fixes incorrect docs and makes the interval longer as 10 ms was way
> too spammy.

## Testing steps
All tests pass and the network logs in Playwright traces display the requests made by the client.

### DIFF
--- a/_playwright-tests/test-utils/src/fixtures/client.ts
+++ b/_playwright-tests/test-utils/src/fixtures/client.ts
@@ -1,38 +1,77 @@
 // Define a fixture to hold the API client
-import { test as oldTest, expect } from '@playwright/test';
-import { Configuration, ResponseContext, ResponseError } from '../client';
-import { setGlobalDispatcher, ProxyAgent, Agent } from 'undici';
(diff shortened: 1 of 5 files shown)


katarinazaprazna (approved):
Thanks for the changes! This is super helpful. Nice work 🙌

rverdile:
@dominikvagner did you wanna merge this? :)

PR #1462, opened 2026-04-14
HMS-10537: add none filter for extended release version

rverdile:
## Summary
Adds a "none" filter to extended release version so it is possible to filter by major version only

## Testing steps
1. Create a template with an extended release version and import an extended support repository
2. Filtering by `/templates?version=9&extended_release_version=none`. Should return only templates with major version 9 that are not extended support templates.
3. Filtering by `/repositories?distribution_version=9&extended_release_version=none` should return only repositories with major version 9 that are not extended release repositories.

### DIFF
--- a/api/docs.go
+++ b/api/docs.go
@@ -478,7 +478,7 @@ const docTemplate = `{
                     },
                     {
                         "type": "string",
-                        "description": "A comma separated list of extended release versions to filter on (e.g. 9.4,9.6)",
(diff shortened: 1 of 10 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-10537

xbhouse (approved):
thank you! tested and looks good :) ack!

PR #1468, opened 2026-04-17
HMS-10346: Add support CSV list of repo features

swadeley:
## Summary
HMS-10346 support feature filter with overlapping product IDs
 Feature filter can now handle comma separated list.
## Testing steps

On a running backend:

Set both of RHEL-EEUS-x86_64,RHEL-E4S-x86_64 in configs/config.yaml

run `make repos-import`

Expect to see two rows with:

```
    ---------------------|           feature_name           |                                                 
------------------------ | RHEL-EEUS-x86_64,RHEL-E4S-x86_64 | 
```

Set one of RHEL-EEUS-x86_64,RHEL-E4S-x86_64 in  in configs/config.yaml

Expect to see the same two rows.

### DIFF
--- a/pkg/clients/pulp_client/content_guard.go
+++ b/pkg/clients/pulp_client/content_guard.go
@@ -3,7 +3,8 @@ package pulp_client
 import (
 	"context"
 	"fmt"
-	"reflect"
+	"slices"
+	"strings"
 
 	"github.com/content-services/content-sources-backend/pkg/api"
 	"github.com/content-services/content-sources-backend/pkg/config"
@@ -19,8 +20,12 @@ const TURNPIKE_JQ_FILTER = ".identity.x509.subject_dn"
 
 const COMPOSITE_GUARD_NAME = "composite_guard"
 
-func rhelCompositeGuardName(featureName string) string {
-	return fmt.Sprintf("rhel_composite_%s", featureName)
+func rhelCompositeGuardName(features []string) string {
+	return "rhel_composite_" + strings.Join(features, "_")
+}
+
+func featureGuardPulpName(features []string) string {
+	return "feature_" + strings.Join(features, "_")
 }
 
 func (r pulpDaoImpl) CreateOrUpdateGuardsForOrg(ctx context.Context, orgId string) (string, error) {
@@ -41,20 +46,24 @@ func (r pulpDaoImpl) CreateOrUpdateGuardsForOrg(ctx context.Context, orgId strin
 }
 
 func (r pulpDaoImpl) CreateOrUpdateGuardsForRhelRepo(ctx context.Context, featureName string) (string, error) {
-	// First create/update/fetch the Feature Guard
-	FeatureHref, err := r.CreateOrUpdateFeatureGuard(ctx, featureName)
+	features := utils.ParseFeatures(featureName)
+	if len(features) == 0 {
+		return "", fmt.Errorf("feature name required for RHEL composite content guard")
+	}
+	guardHrefs := make([]string, 0, len(features)+1)
+	for _, f := range features {
+		href, err := r.CreateOrUpdateFeatureGuard(ctx, f)
+		if err != nil {
+			return "", err
+		}
+		guardHrefs = append(guardHrefs, href)
+	}
+	turnpikeHref, err := r.CreateOrUpdateTurnpikeGuard(ctx)
 	if err != nil {
 		return "", err
 	}
-	// Second create/update/fetch the guard for turnpike
-	TurnpikeHref, err := r.CreateOrUpdateTurnpikeGuard(ctx)
-	if err != nil {
-		return "", err
-	}
-
-	// lastly join them together with the RHEL composite guard
-	CompositeHref, err := r.createOrUpdateRhelCompositeGuard(ctx, FeatureHref, TurnpikeHref, featureName)
-	return CompositeHref, err
+	guardHrefs = append(guardHrefs, turnpikeHref)
+	return r.createOrUpdateRhelCompositeGuard(ctx, guardHrefs, features)
 }
 
 func (r pulpDaoImpl) CreateOrUpdateOrgIdGuard(ctx context.Context, orgId string) (string, error) {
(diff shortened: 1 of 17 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-10346

pkg/clients/pulp_client/content_guard.go
rverdile, line 1: I'd forgotten about feature guards, good find. In this case, we actually want to create a composite feature guard for both features. Assigning both features to one feature guard uses AND logic, meaning all features must match. We want a user with E4S or EEUS to be able to access the content and composite feature guards use OR logic.

And using a composite would probably also make it so you don't have to change all the function signatures to use a slice of strings. You would just need to pass a list of guards to the composite function.

pkg/utils/feature_names.go
rverdile, line 1: Imo we can combine some of the functions here to reduce some of the abstraction. How does this sound?

I think `SplitFeatureNames`, `NormalizeUniqueSorted`, and `NormalizeUniqueSortedFeatureNamesFromCSV` could all be condensed into one function: `ParseFeatures(featureNames string) []string`.

And `EntitledToAnyFeatureInCSV` and `ImportFeatureMatches` could be merged into: `AnyFeatureMatch(featureNames string, list []string) bool`.

And this is more of a nitpick but `AllStringsIn` could be moved to `utils.go` since it's generically useful and not specific to features.

pkg/clients/pulp_client/content_guard.go
```diff
@@ -19,8 +20,12 @@ const TURNPIKE_JQ_FILTER = ".identity.x509.subject_dn"
 
 const COMPOSITE_GUARD_NAME = "composite_guard"
 
-func rhelCompositeGuardName(featureName string) string {
-	return fmt.Sprintf("rhel_composite_%s", featureName)
+func rhelCompositeGuardName(features []string) string {
+	return "rhel_composite_" + strings.Join(features, "_")
```
rverdile, line 24: Small thing but the features names can also have an underscore, so for readability you might want to test a `+` or `,`. I don't know if the pulp API limits the characters.

pkg/external_repos/snapshotted_repos_importer.go
rverdile, line 1: Could you add a test to `pkg/external_repos/snapshotted_repos_importer_test.go`, that tests the import behavior? It would do something like set override FeatureFilter to only ["RHEL-E4S-x86_64"] (without RHEL-EEUS-x86_64), call loadFromFiles(), and assert that the EEUS x86_64 repos are still present in the result.

rverdile (commented):
A few comments, but looks good overall!

pkg/clients/pulp_client/content_guard.go
swadeley: Hi, it works, test added.

pkg/external_repos/snapshotted_repos_importer.go
swadeley: OK, test added, thank you.

pkg/utils/feature_names.go
swadeley: Hi, changes made, please review.

pkg/clients/pulp_client/content_guard.go
swadeley: Hi, changes made, please review.

rverdile (commented):
~~When multiple repos with the same feature name are processed concurrently (e.g. during a batch job or parallel snapshot tasks), two goroutines can both call `fetchFeatureGuard`, both get nil, and then both try to create the same guard — the second fails with 400~~

ignore this :)

pkg/clients/pulp_client/content_guard.go
rverdile, line 60: I think we simplify the flow in content_guard.go a little more by creating two feature guards with `CreateOrUpdateFeatureGuard` in a loop, then creating a composite guard `createOrUpdateRhelCompositeGuard` that includes all the features rather. 

It would simplify the logic a little, so for example: ` Feature 1 || Feature 2 || TurnpikeGuard ` rather than ` (Feature 1 || Feature 2) || TurnpikeGuard`. 

I think it would also remove the need to introduce as many new "create feature guard methods" and could mostly likely just re-use the existing one that creates one feature guard.

How does that sound?

pkg/tasks/update_template_content.go
```diff
@@ -56,6 +57,16 @@ func UpdateTemplateContentHandler(ctx context.Context, task *models.TaskInfo, qu
...
 	}
+	// Task payload is JSONB round-tripped through the queue; in some environments RepoConfigUUIDs
+	// can end up nil/empty while the template row still has associations. Unfiltered
+	// RepositoryConfig.List (empty UUID) then pulls every org / Red Hat / community repo and
+	// overwrites the Candlepin environment with the wrong set of content overrides.
+	if len(opts.RepoConfigUUIDs) == 0 && len(template.RepositoryUUIDS) > 0 {
+		opts.RepoConfigUUIDs = sanitizeRepoConfigUUIDs(template.RepositoryUUIDS)
+	}
+	if len(opts.RepoConfigUUIDs) == 0 {
+		return fmt.Errorf("update template content: no repository config uuids in task payload and template has none (template %s)", opts.TemplateUUID)
```
rverdile: were you seeing an error without this?

rverdile (commented):
Thanks for the updates, looking really good. Just a couple of comments. I was testing it and seems like it's working. Although we can't properly verify pulp until this is in stage.

pkg/clients/pulp_client/content_guard.go
swadeley: Hi, OK, will do.

pkg/tasks/update_template_content.go
swadeley: Hi @rverdile ,
git blame says this is from `2026-04-17 19:34:19 +0100` which is the first day I started working on this. I don't have logs going back that far. I will remove it and run tests.

pkg/clients/pulp_client/content_guard.go
rverdile: @swadeley did you make these changes yet or were you still planning to do that?

pkg/tasks/update_template_content.go
swadeley: Hi @rverdile 

I did remove the  empty RepoConfigUUID handling and have now fixed the docs in line with that change.

If you cannot think of any path that can enqueue update-template-content with empty RepoConfigUUIDs then we should be OK.

pkg/clients/pulp_client/content_guard.go
swadeley: Hi @rverdile , yes, there is now one composite whose Guards list is [feature₁, feature₂, …, turnpike]

pkg/tasks/update_template_content.go
```diff
@@ -610,6 +618,23 @@ func difference(a, b []string) []string {
 	return diff
 }
 
+func sanitizeRepoConfigUUIDs(repoConfigUUIDs []string) []string {
```
rverdile: The UUIDs in the payload come directly from our database, which is already doing syntax checks on the uuid. I think this was added a part of a possible fix to a test failure you saw? But i think it's likely not needed

rverdile:
one more comment, otherwise looks good

pkg/tasks/update_template_content.go
swadeley: Hi @rverdile , I have removed that sanitizeRepoConfigUUIDs helper.  Thank you.

rverdile (approved):
nice job! looks good

swadeley:
> nice job! looks good

Thank you

PR #1470, opened 2026-04-19
HMS-10532: sort repos and templates by minor version

katarinazaprazna:
## Summary

- Enhanced version sorting for repos and templates to handle minor and major versions. The order follows RHEL versioning: 8, 8.0, 8.1, ..., 9, 9.0, 9.1, ...
- Uses `extended_release_version` for minor version repos/templates, with this formula `major * 10000 + minor (or major * 10000 - 1 for major-only)` to compute an integer sort key

Screenshot From 2026-04-20 00-19-48

## Testing steps

- Existing tests pass
- Repositories sorted by version return "", "9.4", "9.4", "9.6" in ascending order
- Templates sorted by version return "8.6", "", "9.4" in ascending order

### DIFF
--- a/pkg/dao/repository_configs.go
+++ b/pkg/dao/repository_configs.go
@@ -541,11 +541,13 @@ func (r repositoryConfigDaoImpl) List(
 		return api.RepositoryCollectionResponse{}, totalRepos, err
 	}
 
+	versionSortExpr := "(CASE WHEN extended_release_version IS NOT NULL AND extended_release_version != '' THEN split_part(extended_release_version, '.', 1)::int * 10000 + split_part(extended_release_version, '.', 2)::int ELSE (SELECT min(v::int * 10000 - 1) FROM unnest(versions) AS v WHERE v <> 'any') END, cardinality(versions))"
+
 	sortMap := map[string]string{
 		"name":                      "name",
 		"url":                       "url",
 		"distribution_arch":         "arch",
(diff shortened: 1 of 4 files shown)


katarinazaprazna:
Hi @rverdile and @xbhouse, I’m concerned about the performance of my sorting logic for repos, as the sort expression appears to run for every row on every request. This overhead could become a bottleneck in larger repo datasets. I’d like to get your thoughts on whether we should optimize this. Thanks!

xbhouse:
https://issues.redhat.com/browse/HMS-10532

xbhouse:
@katarinazaprazna i took a look at this and ran `EXPLAIN ANALYZE` comparing the new sort vs the original.

for repositories, with the full set of RH and EPEL repos (65), the new sort actually looks faster :) 
- ~0.317ms (new)
- ~0.505ms (old)

i think since your new query uses `extended_release_version` to sort when it's present, the number of times we have to account for multiple versions in the version array is reduced :star_struck: 

for templates, i checked the current max number of templates an org has in prod, which is 37. with a set of 50 templates, the new sort is _slightly_ slower, which makes sense:
- ~0.202ms (new)
- ~0.135ms (old)

based on this i'd say we don't need to optimize right now, but we can revisit if these numbers grow or if we start seeing an impact on performance. @rverdile @jlsherrill - curious what you think?

rverdile:
@xbhouse i agree!

xbhouse (approved):
looks great to me! thank you! 🫶

PR #1474, opened 2026-04-20
HMS-9613: new api endpoint bulkRemoveRpms

Starle21:
## Summary
Clients can now remove RPM packages from the latest snapshot of an upload repository.
A new snapshot is automatically created with the remaining packages.
Templates that use the latest snapshot are also updated.

To prevent race conditions, there is logic that blocks enqueuing bulkRemoveRpms task or addUploads task  if any of the relevant tasks are running.

## Testing steps
Create an upload repository and add uploads there.
Get the `repository_config uuid` and `[]rpm uuid` to be removed.
Do a post request to `/repositories/:uuid/rpms/bulk_remove/` with the RPMs to be removed.
Check that a new snapshot is created without the removed RPMs. Also the `down arrow` in the repository table on the frontend should show the number of removed RPMs.


xbhouse:
https://issues.redhat.com/browse/HMS-9613

pkg/handler/repositories.go
```diff
@@ -901,6 +902,79 @@ func (rh *RepositoryHandler) bulkImportRepositories(c echo.Context) error {
...
+
+	repository, err := rh.DaoRegistry.RepositoryConfig.Fetch(c.Request().Context(), orgID, repositoryUuid)
+	if err != nil {
+		return ce.NewErrorResponse(ce.HttpCodeForDaoError(err), "Error fetching repository", err.Error())
+	}
+
+	if repository.Origin != config.OriginUpload {
+		return ce.NewErrorResponse(http.StatusBadRequest, "Cannot remove RPMs from this repository", "Only upload repositories support removing RPMs")
+	}
+
```
Starle21, line 952: Should I check if introspect or snapshot are active and cancel them first?

xbhouse: good question! upload repos aren't introspected so we don't need to worry about that. for snapshots, we wouldn't want a running snapshot task on the repo cancelled in the case new rpms are added right before something else was removed, otherwise the content in the latest snapshot might not be what's expected.

what we _might_ want to do, is block removal of rpms if there are any active `bulk-remove-rpms` tasks for the repo. this is probably not a common scenario, but right now if a request to remove the same rpm occurs twice in a row, the second request fails with `error listing rpm publications: 400 Bad Request: ["No value supplied for repository version filter"]`

pkg/handler/repositories.go
```diff
@@ -901,6 +902,79 @@ func (rh *RepositoryHandler) bulkImportRepositories(c echo.Context) error {
...
+// @Accept       json
+// @Produce      json
+// @Param  		 uuid  path    string                     true  "Repository ID."
+// @Param        body  body    api.BulkRemoveRpmsRequest  true  "request body"
+// @Success      201 {object} api.TaskInfoResponse
+// @Failure      400 {object} ce.ErrorResponse
+// @Failure      404 {object} ce.ErrorResponse
+// @Failure      500 {object} ce.ErrorResponse
+// @Router		 /repositories/{uuid}/rpms/bulk_remove/ [post]
+func (rh *RepositoryHandler) bulkRemoveRpms(c echo.Context) error {
```
xbhouse, line 930: could you add a unit test in `pkg/handler/repositories_test.go` for bulkRemoveRpms?

xbhouse:
this might be a good candidate for an integration test too :) could you add a test (or update an existing test) in `test/integration/pulp_upload_test.go` to cover removing an rpm from an upload repo?

pkg/config/tasks.go
xbhouse, line 45: BulkRemoveRpmsTask types should also be cleaned up, adding it to this list will let the cleanup job look for this task type too 🧹

pkg/handler/repositories.go
```diff
@@ -901,6 +902,79 @@ func (rh *RepositoryHandler) bulkImportRepositories(c echo.Context) error {
...
+	}
+
+	if len(req.RpmUuids) == 0 {
+		return ce.NewErrorResponse(http.StatusBadRequest, "No rpm_uuids provided", "rpm_uuids cannot be empty")
+	}
+
+	repository, err := rh.DaoRegistry.RepositoryConfig.Fetch(c.Request().Context(), orgID, repositoryUuid)
+	if err != nil {
+		return ce.NewErrorResponse(ce.HttpCodeForDaoError(err), "Error fetching repository", err.Error())
+	}
```
xbhouse, line 947: we probably want to also try fetching the rpms requested for a repo to make sure they're valid before enqueuing the task

pkg/handler/repositories.go
Starle21: > what we might want to do, is block removal of rpms if there are any active bulk-remove-rpms tasks for the repo

Good catch! I will add that check. :rocket: 

I also talked with Justin that I should not need to cancel a running snapshot task, but there could be race conditions so I might want to test if everything works in that scenario.

Now I wonder that there might be a race condition - if one worker is creating a new last snapshot and another is doing the removal, the one doing the removal might read the second-to-last snapshot before the first worker finishes and the second worker then starts creating a new snapshot based on that. Do you think this is a realistic scenario?

pkg/handler/repositories.go
Starle21: Sure I'm on it! Also on the other integration test :blush:

pkg/handler/repositories.go
Starle21: Ok, I'll add that, I did not realize that RPMs might not be valid when the whole repository is "valid".

Also I noticed that when I accidentally wanted to remove RPMs that are not in that repository, the whole repo becomes "unavailable" instead of not doing anything. It is the same check.

Starle21:
I also wonder if we might want to call the endpoint something different? Since it is not truly deleting any RPMs from Pulp, it just actually creates a new snapshot with those RPMs removed. So `bulkRemoveRpms` might not be such a good name. :thinking: 
Maybe `removeRpmsAndCreateSnapshot` or `removeRpmsFromLatestSnapshot`? What do you think?

pkg/handler/repositories.go
xbhouse: the `update-latest-snapshot` task is done sequentially for both `add-uploads` and `bulk-remove-rpms` tasks, so after removing there would be a new snapshot and after adding there would be a different new snapshot. i think you're right though, there could be a stale view of what's in the latest snapshot if these tasks overlap. this could probably happen if more than 1 `add-uploads` tasks are running concurrently too. 

i think this would be rare, but we could maybe block removal of rpms if there are any active `update-latest-snapshot` tasks or `add-uploads` tasks for the repo? and similarly block adding rpms/uploads if any active `update-latest-snapshot` or `bulk-remove-rpms`. 

@rverdile i'm curious what you think about these scenarios and how we should handle them? :thinking:

xbhouse:
> I also wonder if we might want to call the endpoint something different? Since it is not truly deleting any RPMs from Pulp, it just actually creates a new snapshot with those RPMs removed. So `bulkRemoveRpms` might not be such a good name. 🤔 Maybe `removeRpmsAndCreateSnapshot` or `removeRpmsFromLatestSnapshot`? What do you think?

hmm i think `bulkRemoveRpms` makes sense, the end result is a new snapshot without those rpms and this is probably what users would care about. the description you've added in the API spec explains how that's done, i think that's good enough :) `addUploads` also creates a new snapshot with added rpms, so i'd keep the naming consistent with that

pkg/handler/repositories.go
rverdile: Yes, I would block these tasks on each other. These tasks could maybe even run at the same time if all the workers are busy when the tasks are first queued, which could be an issue because they both do things like update rpm associations. A slightly better solution would probably be to cancel the active tasks, but I think adding the cleanup for that adds too much complexity here.

rverdile: Another good option might be to look up the active task and then make the soon-to-be-queued task dependent on it. I think that would ensure they run sequentially. But I haven't actually tried that to see how it would work, so I recommend just blocking here for now.

Starle21: I will go with the blocking version then. Thank you for your inputs! :heart:

pkg/handler/repositories.go
xbhouse: oh by "valid" i just meant that the rpm is actually in that repository :)

test/integration/pulp_upload_test.go
```diff
@@ -225,6 +225,50 @@ func (s *UploadSuite) TestUploadResumableBehavior() {
...
+	orgId := fmt.Sprintf("UploadandAddRpm-%v", rand.Int())
+
+	s.identity = test_handler.MockIdentity
+	s.identity.Identity.OrgID = orgId
+
+	// prepare repository and upload one rpm into it
+	repo := s.createUploadRepository()
+
+	t := s.T()
+	rpm := "./fixtures/giraffe/giraffe-0.67-2.noarch.rpm"
```
xbhouse: the integration test failures might be related to using the same rpm for both this test and `TestUploadAndAddRpmPublic`. i think since `resumable` is set to true in `TestUploadAndAddRpmPublic`, it tries to find the upload uuid for an upload with a matching sha256 and chunk size but since it was created by a different org, it's not found. i'd either switch the new test to use an rpm that isn't used by any other tests, or update the other test so `resumable` is `false` there too

test/integration/pulp_upload_test.go
```diff
@@ -225,6 +225,50 @@ func (s *UploadSuite) TestUploadResumableBehavior() {
 	assert.NotEqual(t, uuidSecond, uuidThird)
 }
 
+func (s *UploadSuite) TestBulkRemoveRpms() {
+	orgId := fmt.Sprintf("UploadandAddRpm-%v", rand.Int())
```
xbhouse: i'd update this to `fmt.Sprintf("UploadandRemoveRpm-%v", rand.Int())` to reflect the test name (and avoid any rare collisions, rand.Int() is only "pseudo-random" 😆)

test/integration/pulp_upload_test.go
Starle21: I went with taking the other available `.rpm` file for the `TestBulkRemoveRpms`, so no changes in the other tests. Now it passes! :tada: 
Thanks for your awesome analysis :heart:

mk/go-rules.mk
```diff
@@ -58,7 +58,9 @@ test-integration: ## Run tests for ci
 	CLIENTS_PULP_CLIENT_CERT_PATH="$(PROJECT_DIR)/compose_files/pulp/assets/certs/dev_certs/client.crt" \
 	CLIENTS_PULP_CLIENT_KEY_PATH="$(PROJECT_DIR)/compose_files/pulp/assets/certs/dev_certs/client.key" \
 	CLIENTS_PULP_CA_CERT_PATH="$(PROJECT_DIR)/compose_files/pulp/assets/certs/dev_certs/ca.crt" \
-	go test -timeout 15m $(MOD_VENDOR) ./test/integration/...
+	go test -timeout 15m $(MOD_VENDOR) ./test/integration/... \
+	$(if $(SUITE),-run '$(SUITE)',) \
```
xbhouse, line 62: curious what these additions are for?

Starle21: I needed to have a way to run just one integration suite or just one integration test, with all the setup around it.
If no SUITE or no TEST is there, all integration tests run.

xbhouse (approved):
this looks great to me, awesome work!!

PR #1484, opened 2026-04-28
Build: add fmt to make lint

rverdile:
## Summary
To match the workflow action


swadeley (approved):
ACK

rverdile:
After looking into it further, this actually isn't needed. Run will do both. The lint error was manually being ignored.

PR #1489, opened 2026-05-04
HMS-10637: Include EUS in template filter by major version

rverdile:
## Summary

Template filtering on the major version only needs to include EUS templates to compatible with image builder. But this adds an option to exclude it so it is still compatible with our UI.

## Testing steps
1. Import and snapshot the extended release repos
2. Create templates of varying versions (9, 9.4, 9.6, 10, 10.0, etc)
3. Checkout the frontend PR as well: https://github.com/content-services/content-sources-frontend/pull/993
4. Use the version filtering in various combinations and make sure it works as you'd expect


xbhouse:
https://issues.redhat.com/browse/HMS-10637

xbhouse (approved):
looks good! tested with IB too, templates are showing up :) thank you!!

PR #1490, opened 2026-05-05
Test: Reduce patch polling rate and logging

swadeley:
## Summary
Reduce patch polling rate
Avoids undue load and makes logs less noisy.

### DIFF
--- a/_playwright-tests/test-utils/src/helpers/systemHelpers.ts
+++ b/_playwright-tests/test-utils/src/helpers/systemHelpers.ts
@@ -1,7 +1,13 @@
-import { Page, expect } from '@playwright/test';
+import { expect, Page } from '@playwright/test';
 
 /** Poll timeout for system propagation to inventory and patch (3 minutes) */
 export const INVENTORY_PATCH_POLL_TIMEOUT_MS = 180_000;
+
+/**
+ * Delay between Inventory and Patch polling attempts. Playwright can emit a network line per
+ * `page.request` call; a modest interval keeps logs readable without materially slowing success.
+ */
+const INVENTORY_PATCH_POLL_INTERVAL_MS = 10_000;
 
 /**
  * Count matching systems in Patch.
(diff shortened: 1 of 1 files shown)


_playwright-tests/test-utils/src/helpers/systemHelpers.ts
```diff
@@ -109,17 +134,15 @@ export const waitInPatch = async (
   expectedAttachment: boolean = true,
   timeoutMs: number = INVENTORY_PATCH_POLL_TIMEOUT_MS,
 ): Promise<void> => {
-  await expect
-    .poll(async () => await isInInventory(page, hostname), {
-      message: 'System did not appear in inventory in time',
-      timeout: timeoutMs,
-    })
```
katarinazaprazna, line 116: I believe we could pass `intervals: [INVENTORY_PATCH_POLL_INTERVAL_MS]` directly to Playwright's poll function in order to enforce the 3-second gap between checks. Then we wouldn't need to maintain the custom `waitForPollResult`

_playwright-tests/test-utils/src/helpers/systemHelpers.ts
```diff
@@ -109,17 +134,15 @@ export const waitInPatch = async (
...
-    .toBe(1);
-
-  await expect
-    .poll(async () => await isInPatch(page, hostname, expectedAttachment), {
-      message: 'System did not appear in patch in time',
-      timeout: timeoutMs,
```
katarinazaprazna, line 122: Same as above

_playwright-tests/test-utils/src/helpers/systemHelpers.ts
swadeley: OK, done that. I also increased the interval to 10 secs as I want to reduce the amount of logging not just the load in patch in stage. I think 10 secs in relation to the timeout of 3 mins should be acceptable.

_playwright-tests/test-utils/src/helpers/systemHelpers.ts
swadeley: Change made, thank you.

katarinazaprazna (approved):
Thank you!

PR #1495, opened 2026-05-14
HMS-10580: Set limit on pulp orphan cleanup tasks

swadeley:
## Summary
Set limit on pulp orphan cleanup tasks
Sets a global cap and adds new constant:
 maxGetTaskAttempts (defaults to 5 when unset) 

HMS-10580 API - pulp orphan cleanup runs too many tasks
## Testing steps

tests pass

### DIFF
--- a/pkg/external_repos/commands/cleanup.go
+++ b/pkg/external_repos/commands/cleanup.go
@@ -6,6 +6,7 @@ import (
 	"slices"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/content-services/content-sources-backend/pkg/clients/pulp_client"
@@ -116,7 +117,8 @@ func enqueueSnapshotsCleanup(ctx context.Context, olderThanDays int, batchSize i
 		return fmt.Errorf("error getting repository configurations: %v", err)
 	}
 	if batchSize > 0 {
-		repoConfigs = repoConfigs[:batchSize] // Limit to batch size
+		limit := min(batchSize, len(repoConfigs))
+		repoConfigs = repoConfigs[:limit]
 	}
 	log.Info().Msgf("Snapshot cleanup: processing %d repositories with outdated snapshots", len(repoConfigs))
 
@@ -143,12 +145,15 @@ func enqueueSnapshotCleanupForRepoConfig(ctx context.Context, taskClient client.
 	slices.SortFunc(snaps, func(s1, s2 models.Snapshot) int {
 		return s1.CreatedAt.Compare(s2.CreatedAt)
 	})
+
+	keepIndex := len(snaps) - 1
+	cutoffTime := time.Now().Add(-time.Duration(olderThanDays) * 24 * time.Hour)
 	toBeDeletedSnapUUIDs := make([]string, 0, len(snaps))
 	for i, snap := range snaps {
-		if i == len(snaps)-1 && len(toBeDeletedSnapUUIDs) == len(snaps)-1 {
+		if i == keepIndex && len(toBeDeletedSnapUUIDs) == len(snaps)-1 {
 			break
 		}
-		if snap.CreatedAt.Before(time.Now().Add(-time.Duration(olderThanDays) * 24 * time.Hour)) {
+		if snap.CreatedAt.Before(cutoffTime) {
 			toBeDeletedSnapUUIDs = append(toBeDeletedSnapUUIDs, snap.UUID)
 		}
 	}
@@ -194,7 +199,6 @@ func enqueueSnapshotCleanupForRepoConfig(ctx context.Context, taskClient client.
 }
 
 func pulpOrphanCleanup(ctx context.Context, db *gorm.DB, batchSize int) error {
-	var err error
 	daoReg := dao.GetDaoRegistry(db)
 
 	domains, err := daoReg.Domain.List(ctx)
@@ -202,7 +206,16 @@ func pulpOrphanCleanup(ctx context.Context, db *gorm.DB, batchSize int) error {
 		log.Panic().Err(err).Msg("orphan cleanup error: error listing orgs")
 	}
 
-	for i := 0; i < len(domains); i += batchSize {
+	return runPulpOrphanCleanupForDomains(ctx, domains, batchSize, pulp_client.GetPulpClientWithDomain)
+}
+
+// runPulpOrphanCleanupForDomains runs orphan cleanup in batches and stops if any PollTask fails
+// (PollTask already retries transient GetTask errors).
+func runPulpOrphanCleanupForDomains(ctx context.Context, domains []models.Domain, batchSize int, getPulpClient func(domainName string) pulp_client.PulpClient) error {
+	var abort atomic.Bool
+	var firstErr error
+
+	for i := 0; i < len(domains) && !abort.Load(); i += batchSize {
 		end := i + batchSize
 		if end > len(domains) {
 			end = len(domains)
@@ -218,8 +231,10 @@ func pulpOrphanCleanup(ctx context.Context, db *gorm.DB, batchSize int) error {
 			wg.Add(1)
 			go func() {
 				defer wg.Done()
-
-				pulpClient := pulp_client.GetPulpClientWithDomain(domainName)
+				if abort.Load() {
+					return
+				}
+				pulpClient := getPulpClient(domainName)
 				cleanupTask, err := pulpClient.OrphanCleanup(ctx)
 				if err != nil {
 					logger.Error().Err(err).Msgf("error starting orphan cleanup")
@@ -227,14 +242,21 @@ func pulpOrphanCleanup(ctx context.Context, db *gorm.DB, batchSize int) error {
 				}
 				logger.Info().Str("task_href", cleanupTask).Msgf("running orphan cleanup for org: %v", org)
 
-				_, err = pulp_client.GetGlobalPulpClient().PollTask(ctx, cleanupTask)
+				_, err = pulpClient.PollTask(ctx, cleanupTask)
 				if err != nil {
-					logger.Error().Err(err).Msgf("error polling pulp task for orphan cleanup")
+					logger.Error().Err(err).
+						Msg("error polling pulp task for orphan cleanup; remaining domains will be skipped")
+					if abort.CompareAndSwap(false, true) {
+						firstErr = err
+					}
 					return
 				}
 			}()
 		}
 		wg.Wait()
+	}
+	if abort.Load() {
+		return fmt.Errorf("pulp orphan cleanup aborted after failed task poll: %w", firstErr)
 	}
 	return nil
 }
--- a/pkg/external_repos/commands/cleanup_pulp_orphan_test.go
+++ b/pkg/external_repos/commands/cleanup_pulp_orphan_test.go
@@ -1,0 +1,58 @@
+package commands
+
+import (
+	"context"
+	"errors"
+	"sync/atomic"
+	"testing"
+
+	"github.com/content-services/content-sources-backend/pkg/clients/pulp_client"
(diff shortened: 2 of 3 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-10580

rverdile (commented):
Let's just use a constant here. I don't think this is a value that needs to be configurable.

Also this is a behavior that can be tested locally by manually writing an error after the return for the polling task

swadeley:
Testing with "maxPollFailures = 2":

```
:~/content-sources-backend$ go test -count=1 -v ./pkg/external_repos/commands/ -run TestRunPulpOrphanCleanupForDomains
=== RUN   TestRunPulpOrphanCleanupForDomains_StopsAfterMaxPollFailures
=== PAUSE TestRunPulpOrphanCleanupForDomains_StopsAfterMaxPollFailures
=== CONT  TestRunPulpOrphanCleanupForDomains_StopsAfterMaxPollFailures
{"level":"info","org_id":"1","pulp_domain_name":"d1","task_href":"https://pulp.example/api/v3/tasks/fake-href/","time":"2026-05-14T18:44:21+01:00","message":"running orphan cleanup for org: 1"}
{"level":"error","org_id":"1","pulp_domain_name":"d1","error":"injected poll failure","time":"2026-05-14T18:44:21+01:00","message":"error polling pulp task for orphan cleanup"}
{"level":"info","org_id":"2","pulp_domain_name":"d2","task_href":"https://pulp.example/api/v3/tasks/fake-href/","time":"2026-05-14T18:44:21+01:00","message":"running orphan cleanup for org: 2"}
{"level":"error","org_id":"2","pulp_domain_name":"d2","error":"injected poll failure","time":"2026-05-14T18:44:21+01:00","message":"error polling pulp task for orphan cleanup"}
{"level":"error","max_poll_failures":2,"poll_failures":2,"time":"2026-05-14T18:44:21+01:00","message":"stopping pulp orphan cleanup: too many failed task polls; remaining domains skipped (orphan cleanup tasks might still be running on Pulp)"}
--- PASS: TestRunPulpOrphanCleanupForDomains_StopsAfterMaxPollFailures (0.00s)
PASS
ok  	github.com/content-services/content-sources-backend/pkg/external_repos/commands	0.006s
```

pkg/external_repos/commands/cleanup.go
```diff
@@ -218,24 +248,33 @@ func pulpOrphanCleanup(ctx context.Context, db *gorm.DB, batchSize int) error {
...
 
-				_, err = pulp_client.GetGlobalPulpClient().PollTask(ctx, cleanupTask)
+				err = pulp.pollOrphanTask(ctx, cleanupTask)
 				if err != nil {
 					logger.Error().Err(err).Msgf("error polling pulp task for orphan cleanup")
+					n := pollFailures.Add(1)
+					if n >= int64(maxPollFailures) && abort.CompareAndSwap(false, true) {
+						log.Error().Int("max_poll_failures", maxPollFailures).Int64("poll_failures", n).
+							Msg("stopping pulp orphan cleanup: too many failed task polls; remaining domains skipped (orphan cleanup tasks might still be running on Pulp)")
+					}
```
rverdile: I think it would make sense to move, let's call it the "circuit breaker" logic, into the [pulp_client.PollTask function](https://github.com/content-services/content-sources-backend/blob/main/pkg/clients/pulp_client/task.go#L68) itself. 

I think it would be pretty simple. Let's say, if the call to `GetTask` errors, rather than immediately return, let's have it retry (with the same backoff) 5 times. And if it fails 5 times, then return the error. This way other callers of PollTask would benefit from this change. 

Then in the cleanup command, when it errors once, we just exit.

What do you think?

pkg/external_repos/commands/cleanup.go
```diff
@@ -24,6 +25,27 @@ import (
...
 
+// pulpOrphanMaxPollFailures caps failed PollTask calls before checking further domains. Transient
+// errors (e.g. 504) do not advance batches while orphan cleanup tasks continue to run on Pulp.
+const pulpOrphanMaxPollFailures = 10
+
+// pulpOrphanPulp abstracts Pulp orphan cleanup for tests.
+type pulpOrphanPulp interface {
+	startOrphanCleanup(ctx context.Context, domainName string) (taskHref string, err error)
+	pollOrphanTask(ctx context.Context, taskHref string) error
+}
```
rverdile: Defining an interface is a good approach to unit testing this! But PulpClient is already an interface. What I would suggest is instead using the PulpClient interface and mocking the result of the PollTask function call. There are examples in other tests for how to mock it.

pkg/external_repos/commands/cleanup.go
swadeley: Hi @rverdile , I have implemented that "GetTask" suggestion and set maxGetTaskAttempts to 5.

pkg/external_repos/commands/cleanup.go
swadeley: OK, changes made.

pkg/external_repos/commands/cleanup.go
```diff
@@ -143,12 +145,15 @@ func enqueueSnapshotCleanupForRepoConfig(ctx context.Context, taskClient client.
 	slices.SortFunc(snaps, func(s1, s2 models.Snapshot) int {
 		return s1.CreatedAt.Compare(s2.CreatedAt)
 	})
-	toBeDeletedSnapUUIDs := make([]string, 0, len(snaps))
+
+	keepIndex := len(snaps) - 1
+	cutoffTime := time.Now().Add(-time.Duration(olderThanDays) * 24 * time.Hour)
+	toBeDeletedSnapUUIDs := make([]string, 0, len(snaps)-1)
```
TenSt:
```suggestion
	toBeDeletedSnapUUIDs := make([]string, 0, len(snaps))
```

swadeley: Hi, I was thinking we don't want to delete the most recent snapshot. 
But, if we are sure they are all orphans, then OK, I'll remove that.

pkg/external_repos/commands/cleanup_pulp_orphan_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	mockPulp := pulp_client.NewMockPulpClient(t)
+	mockPulp.On("OrphanCleanup", mock.Anything).Return(taskHref, nil).Once()
+	mockPulp.On("PollTask", mock.Anything, taskHref).Return(nil, errors.New("injected poll failure")).Once()
+
+	getPulpClient := func(_ string) pulp_client.PulpClient {
+		return mockPulp
+	}
+
+	err := runPulpOrphanCleanupForDomains(ctx, domains, 1, getPulpClient)
```
rverdile: it would be good to be the batchSize greater than one so it tests the parallel go routines. Maybe 9 domains, batch size 3 would be good

pkg/external_repos/commands/cleanup.go
```diff
@@ -143,12 +145,15 @@ func enqueueSnapshotCleanupForRepoConfig(ctx context.Context, taskClient client.
...
 	})
+
+	keepIndex := len(snaps) - 1
+	cutoffTime := time.Now().Add(-time.Duration(olderThanDays) * 24 * time.Hour)
 	toBeDeletedSnapUUIDs := make([]string, 0, len(snaps))
 	for i, snap := range snaps {
-		if i == len(snaps)-1 && len(toBeDeletedSnapUUIDs) == len(snaps)-1 {
-			break
+		if i == keepIndex {
+			continue
```
rverdile: I'd like to leave in the `len(toBeDeletedSnapUUIDs) == len(snaps)-1`. There should always be 1 snapshot left. The keepIndex logic does enforce that, but checking the length more explicit

pkg/external_repos/commands/cleanup.go
rverdile: `toBeDeletedSnapUUIDs := make([]string, 0, len(snaps))` is how it was previously, so I think it's fine to leave it. It's true that we don't want to delete the most recent snapshot (there should always be one snapshot), but the logic in the for loop is handling that

rverdile: also this section of the code is unrelated to orphan cleanup. This is for snapshot cleanup, which deletes expired snapshots

pkg/external_repos/commands/cleanup.go
swadeley: Hi @rverdile  OK, change made. Thank you.

pkg/external_repos/commands/cleanup_pulp_orphan_test.go
swadeley: Hi @rverdile  OK, change made. Thank you.

pkg/clients/pulp_client/uploads.go
rverdile, line 1: what are these changes to uploads.go for? if they're unrelated, maybe we can leave them out. I worry about the changed return value causing a bug

pkg/external_repos/commands/cleanup.go
```diff
@@ -218,24 +235,35 @@ func pulpOrphanCleanup(ctx context.Context, db *gorm.DB, batchSize int) error {
...
+				_, err = pulpClient.PollTask(ctx, cleanupTask)
 				if err != nil {
-					logger.Error().Err(err).Msgf("error polling pulp task for orphan cleanup")
+					logger.Error().Err(err).
+						Msg("error polling pulp task for orphan cleanup; remaining domains will be skipped")
+					if abort.CompareAndSwap(false, true) {
+						firstErrMu.Lock()
+						firstErr = err
+						firstErrMu.Unlock()
+					}
```
rverdile, line 251: I believe the mutex here is redundant. CompareAndSwap will only return true for the first GoRoutine that changes the value (when abort is false). Once abort is true, it will return false, so only the first goroutine it hit an error writes here.

rverdile: I would remove the mutex for simplification

pkg/external_repos/commands/cleanup.go
rverdile: the keepIndex line would keep the check from running. I would either remove the keepIndex conditional, or I would put them in the same statement, i.e. 

```
if keepIndex && len(toBeDeletedSnapUUIDs) == len(snaps)-1 {
			break
		}
```
Either way, let's remove one of these conditionals

swadeley: Hi, OK, like so then:

```
if i == keepIndex && len(toBeDeletedSnapUUIDs) == len(snaps)-1 {
    break
}
```

pkg/clients/pulp_client/uploads.go
swadeley: They are small refactors, mostly because cleanup.go touches upload cleanup too. I will reverse the change but just for info, here is a summary of them:

| Change | Purpose |
|--------|--------|
| **`UploadHref(domainName, uploadUUID)`** | Centralizes building `/api/pulp/{domain}/api/v3/uploads/{uuid}/`. `uploadCleanup` in `cleanup.go` now calls this instead of an inline string. |
| **`pulpUploadNotFoundMessage`** | Pulls the `"No Upload matches"` substring into a named constant used by `DeleteUpload` when treating 404 as “already gone” (same behavior as before). |
| **`UploadChunk` returns `nil` on error** | On failure it used to return `&zest.UploadResponse{}`; it now returns `nil`, which is normal Go style and matches other methods in that file (`CreateUpload`, `FinishUpload`). |

pkg/external_repos/commands/cleanup.go
swadeley: OK, will do.

rverdile (approved):
looks good, nice job!

swadeley:
> looks good, nice job!

Thank you.

PR #1496, opened 2026-05-14
Test: exclude EPEL, RH repos from BulkCreateCleanupURL test

xbhouse:
## Summary

Excludes EPEL and RH repos from the query in BulkCreateCleanupURL so the test does not intermittently fail if one of the seeded repos is found first

## Testing steps

TestBulkCreateCleanupURL passes consistently

### DIFF
--- a/pkg/dao/repository_configs_test.go
+++ b/pkg/dao/repository_configs_test.go
@@ -273,7 +273,7 @@ func (suite *RepositoryConfigSuite) TestCreateAlreadyExists() {
 	orgID := seeds.RandomOrgId()
 	var err error
 
-	err = seeds.SeedRepository(tx, 1, seeds.SeedOptions{})
+	_, err = seeds.SeedRepository(tx, 1, seeds.SeedOptions{})
 	assert.NoError(t, err)
 	var repo []models.Repository
 	err = tx.Limit(1).Find(&repo).Error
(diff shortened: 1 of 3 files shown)


pkg/dao/repository_configs_test.go
```diff
@@ -532,7 +532,7 @@ func (suite *RepositoryConfigSuite) TestBulkCreateCleanupURL() {
 	err := seeds.SeedRepository(tx, 1, seeds.SeedOptions{})
 	require.NoError(t, err)
 
-	err = tx.Where("origin != ?", config.OriginUpload).Where("url IS NOT NULL").First(&repository).Error
+	err = tx.Where("origin != ?", config.OriginUpload).Where("url IS NOT NULL").Where("url not in ?", config.EPELUrls).First(&repository).Error
```
rverdile: would it make sense to filter out the community origin? and maybe even the red hat origin? rather than specific URLs

xbhouse: it would, but looking at the db, epel 8 and 9 repos are imported somewhere with the `external` origin. i actually don't know where they're coming from :eyes:

xbhouse: oh.. i wonder if it's a db migration?

xbhouse: i'm not sure about removing those, but i could update this test to query by the URL of the repo just created instead

TenSt (dismissed):

pkg/dao/repository_configs_test.go
rverdile: yeah, let's do that. It seems like an error could happen for community or red hat repos too, even if it hasn't yet.

rverdile: @TenSt for awareness, there's a change that needs to be made here

xbhouse: thank you! just pushed up those changes :)

TenSt (approved):

PR #1509, opened 2026-06-01
HMS-10467: Making deletions tasks more resilient

Starle21:
## Summary
Previously, if any of the 3 deletion tasks:
`delete_snapshots`
`delete_templates`
`delete_repository_snapshots`
failed to delete a resource, it was not exactly clear where and why it happened.

To make it more resilient, this was changed:
- detailed failure messages were added to identify easily which op caused a failure
- logging warnings if a resource was not found or nil
- loops now return errors at the end, so all the resources in a loop get a chance to be deleted

## Testing steps

Test manually by creating templates, repositories with snapshots and without snapshots and then deleting them.
Everything works as before, but you get much better idea what is happening if anything fails.
Automated tests pass.

### DIFF
--- a/pkg/tasks/delete_repository_snapshots.go
+++ b/pkg/tasks/delete_repository_snapshots.go
@@ -33,7 +33,9 @@ type DeleteRepositorySnapshots struct {
 
 // This org may or may not have a domain created in pulp, so make sure the domain exists and if not, return a nil pulpClient
 func lookupOptionalPulpClient(ctx context.Context, globalClient pulp_client.PulpGlobalClient, task *models.TaskInfo, daoReg *dao.DaoRegistry) (*pulp_client.PulpClient, error) {
+	logger := LogForTask(task.Id.String(), task.Typename, task.RequestID)
 	if !config.PulpConfigured() {
+		logger.Debug().Msg("pulp not configured, skipping pulp client setup")
 		return nil, nil
 	}
 	domainName, err := daoReg.Domain.FetchOrCreateDomain(ctx, task.OrgId)
@@ -51,6 +53,10 @@ func lookupOptionalPulpClient(ctx context.Context, globalClient pulp_client.Pulp
 		client := pulp_client.GetPulpClientWithDomain(domainName)
 		return &client, nil
 	}
+	logger.Debug().
+		Str("org_id", task.OrgId).
+		Str("domain_name", domainName).
+		Msg("no pulp domain found for org, org has never snapshotted, skipping pulp cleanup")
 	return nil, nil
 }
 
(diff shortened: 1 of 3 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-10467

pkg/tasks/delete_repository_snapshots.go
```diff
@@ -115,20 +121,24 @@ func (d *DeleteRepositorySnapshots) Run() error {
...
-			err = d.deleteSnapshot(snap.UUID)
-			if err != nil {
-				return err
+			if err = d.deleteSnapshot(snap.UUID); err != nil {
+				snapErrs = append(snapErrs, err)
 			}
 		}
+		if err = errors.Join(snapErrs...); err != nil {
+			return err
+		}
```
rverdile, line 141: If deleteTemplateSnapshot errors, let's also continue. I think if you try to delete the snapshot before deleting the template snapshot, I think that would cause an error or at least leave things in an odd state.

pkg/tasks/delete_repository_snapshots.go
```diff
@@ -160,60 +171,80 @@ func (d *DeleteRepositorySnapshots) fetchSnapshots() ([]models.Snapshot, error)
...
 	return task, nil
 }
 
 func (d *DeleteRepositorySnapshots) deleteRpmRepoAndRemote() (taskRepo, taskRemote *zest.TaskResponse, err error) {
+	logger := LogForTask(d.task.Id.String(), d.task.Typename, d.task.RequestID)
+
 	remoteResp, err := d.getPulpClient().GetRpmRemoteByName(d.ctx, d.payload.RepoConfigUUID)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, fmt.Errorf("failed to look up rpm remote %v: %w", d.payload.RepoConfigUUID, err)
```
rverdile: In this error message, and in a few others for this task `delete_repository_snapshots`, let's remove the inclusion of the RepoConfigUUID. This task is scoped to a particular repo config, as it deletes that repo config and its snapshots, so we'll know the UUID without seeing it in the error message.

pkg/tasks/delete_templates.go
```diff
@@ -90,7 +90,7 @@ func (d *DeleteTemplates) Run() error {
 	if config.CandlepinConfigured() {
 		err = d.cpClient.DeleteEnvironment(d.ctx, d.payload.TemplateUUID)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to delete candlepin environment for template %v: %w", d.payload.TemplateUUID, err)
```
rverdile: Similarly, this task is scoped to a particular template, so let's remove the template uuid from the error messages

rverdile (commented):
To small comments and then it's good to go!

pkg/tasks/delete_repository_snapshots.go
Starle21: True, it's an overkill.
I removed it from the error messages, but kept it in the log ones, as I think it might be easier to read them.
But if you think it should be removed also from those, let me know :)

rverdile (approved):
nice job!!

PR #1511, opened 2026-06-01
Build: Add agents files

rverdile:
## Summary
Adds a couple of agents files for use by Claude and Cursor. Should speed up some answers a little.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,47 @@
+# AGENTS.md
+
+Project: Backend for Red Hat's content sources service (Go + PostgreSQL)
+
+Frontend: https://github.com/content-services/content-sources-frontend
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -0,0 +1,47 @@
...
+
+Frontend: https://github.com/content-services/content-sources-frontend
+
+## Overview
+
+See [docs/architecture.md](docs/architecture.md) for a description of the service and its architecture.
+
+## Dev environment setup
+
+- Requires podman/docker with compose
```
swadeley, line 13: Try to avoid slash in text and rather use "or" or "and" as appropriate.

swadeley (approved):
ACK

PR #1513, opened 2026-06-03
HMS-10486: increase introspect timeout, enqueue once per job

xbhouse:
## Summary

- Increases CDN request timeout during introspection from 6 to 10 minutes. In stage, some larger repos are getting close to hitting the 6 minute limit:
```
{"level":"debug","task_type":"introspect","task_id":"c965a698-1d90-483d-b632-e1dcfeb85b4b","request_id":"","url":"https://cdn.redhat.com/content/eus/rhel9/9.8/aarch64/baseos/os/","phase":"parse_packages","duration":331.864419797,"time":"2026-06-01T14:05:39Z","severity":"debug","message":"introspection phase parse_packages completed"}
```
-  Adjusts the `process-repos` job and snapshot command to only enqueue one introspect task per repo and prevent new introspect or snapshot tasks if any are already active for a repo
- Adds an optional `--introspect` flag to the snapshot command and updates `make repos-minimal` to use this
- Fixes a couple methods used during introspection to return proper values

## Testing steps

1. Run `make repos-import && make process-repos`. All introspect tasks should succeed and there should only be one per repo
2. Running `make process-repos` twice in a row shouldn't add multiple introspects for a repo
2. `make repos-minimal` should still introspect and snapshot the "small" repo (RHEL9 ARM codeready)

### DIFF
--- a/mk/repos.mk
+++ b/mk/repos.mk
@@ -22,6 +22,6 @@ repos-import-rhel10: ## Import only rhel 10 repos
 .PHONY: repos-minimal
 repos-minimal: ## Import and snapshot repos needed for a minimal setup, usefull for Playwright testing, currently: SMALL + EPEL10
 	OPTIONS_REPOSITORY_IMPORT_FILTER=small go run ./cmd/external-repos/main.go import
-	go run cmd/external-repos/main.go snapshot --url https://cdn.redhat.com/content/dist/rhel9/9/aarch64/codeready-builder/os/ --force
+	go run cmd/external-repos/main.go snapshot --url https://cdn.redhat.com/content/dist/rhel9/9/aarch64/codeready-builder/os/ --force --introspect
 	OPTIONS_REPOSITORY_IMPORT_FILTER=epel10 go run ./cmd/external-repos/main.go import
 	go run cmd/external-repos/main.go introspect --url https://dl.fedoraproject.org/pub/epel/10/Everything/x86_64/ --force
--- a/pkg/external_repos/commands/snapshot.go
+++ b/pkg/external_repos/commands/snapshot.go
@@ -25,6 +25,7 @@ func SnapshotAction(c *cli.Context) error {
 	ctx := c.Context
 	urlsParam := c.StringSlice("url")
 	force := c.Bool("force")
+	shouldIntrospect := c.Bool("introspect")
 
 	var urls []string
 	for _, url := range urlsParam {
@@ -33,7 +34,7 @@ func SnapshotAction(c *cli.Context) error {
 
 	if config.Get().Features.Snapshots.Enabled {
 		waitForPulp(ctx)
-		err := enqueueSnapshotRepos(ctx, &urls, nil, force)
+		err := enqueueSnapshotRepos(ctx, &urls, nil, force, shouldIntrospect)
 		if err != nil {
 			log.Warn().Msgf("Error enqueuing snapshot tasks: %v", err)
 		}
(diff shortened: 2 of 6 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-10486

jlsherrill:
out of curiosity, do you know which files are taking so long to download?  Is it the larger metadata files?

xbhouse:
> out of curiosity, do you know which files are taking so long to download? Is it the larger metadata files?

@jlsherrill yea it's mostly the primary, processing the packages for RHEL baseos repos takes the longest usually. for RHEL8, the module streams seem to take quite a bit to process too, the longest i've seen so far for those is about 90 seconds. i'm guessing it takes longer when multiple baseos repos get updates at the same time

rverdile:
> Adjusts the process-repos job to only enqueue one introspect task per repo and prevent new ones if any are active

wanna change this for snapshots too? i.e. don't enqueue a new snapshot task for a repo when one is already active

mk/repos.mk
```diff
@@ -22,6 +22,6 @@ repos-import-rhel10: ## Import only rhel 10 repos
 .PHONY: repos-minimal
 repos-minimal: ## Import and snapshot repos needed for a minimal setup, usefull for Playwright testing, currently: SMALL + EPEL10
 	OPTIONS_REPOSITORY_IMPORT_FILTER=small go run ./cmd/external-repos/main.go import
-	go run cmd/external-repos/main.go snapshot --url https://cdn.redhat.com/content/dist/rhel9/9/aarch64/codeready-builder/os/ --force
+	go run cmd/external-repos/main.go snapshot --url https://cdn.redhat.com/content/dist/rhel9/9/aarch64/codeready-builder/os/ --force --introspect
```
Starle21, line 25: Is this `--introspect` needed here? The previous version without the flag also introspects the repo.

Starle21 (commented):
Testing locally, I confirm that `make process-repos` does not add subsequent tasks if there are still active tasks 👍

mk/repos.mk
xbhouse: yep it's needed :) without this flag, the repo is snapshotted, but isn't introspected. so the introspection metadata, packages, groups, etc are never stored for the repo. and the UI stays in the default state (no packages or last introspection time and a pending status).

we don't use the commands like this in the service, and triggering a snapshot in the UI start both introspection and snapshot tasks. but maybe this is something we should account for explicitly. meaning if a repo hasn't been introspected yet, it shouldn't try to snapshot

swadeley: Hi, re "if a repo hasn't been introspected yet, it shouldn't try to snapshot"

[RFE: don't attempt a snapshot on introspection failure](https://redhat.atlassian.net/browse/HMS-9098)

Not same but similar

Starle21: Ok, did not dig into the introspect deep enough, I did not realize the introspect is not happening properly, even though the task for introspection was happening.
Thanks for clarifying :100:

pkg/external_repos/commands/snapshot.go
```diff
@@ -85,19 +86,31 @@ func enqueueSnapshotRepos(ctx context.Context, urls *[]string, interval *int, fo
...
+		taskIDs, err := dao.GetTaskInfoDao(db.DB).FetchActiveTasks(ctx, repo.OrgID, repo.RepositoryUUID, config.RepositorySnapshotTask)
 		if err != nil {
-			log.Err(err).Msgf("error enqueueing introspection for repository %v", repo.Name)
+			log.Err(err).Msgf("error checking active snapshot tasks for %v", repo.Repository.URL)
+			continue
+		}
+		if len(taskIDs) > 0 {
+			log.Debug().Msgf("skipping snapshot for %v, task already exists", repo.Repository.URL)
+			continue
+		}
```
rverdile, line 122: I think if it fails to check for an active snapshot task we'd still want to enqueue a snapshot. Also, I think the continue here (in both cases) would also skip introspect? I think we should let introspect be decided separately. What do you think?

xbhouse: > I think if it fails to check for an active snapshot task we'd still want to enqueue a snapshot

oh oops :) yes, maybe we should just log an error there and keep going

> Also, I think the continue here (in both cases) would also skip introspect? I think we should let introspect be decided separately.

good catch! you're right, let me fix that

rverdile (approved):
nice job!!

PR #1536, opened 2026-06-10
HMS-10723: add endpoint to return template advisory IDs

xbhouse:
## Summary

- Adds an endpoint that returns a template's advisory IDs with no pagination
- Sends template update events whenever a template is updated

## Testing steps

1. Create a template with snapshots that have advisories.
2. Send a GET request to  `/templates/:uuid/advisories/ids`. This should return all the advisory IDs in the template.
3. Run `make kafka-topic-consume KAFKA_TOPIC=platform.content-sources.template`
4. Update the template. There should be 2 messages in the consumer output, one when the update request is sent and one during the `update-template-content` task.
5. Add a custom repo to a template that's using the latest snapshots. Triggering a new snapshot for that repo should enqueue the `update-latest-snapshot` task, and a message should be sent in the consumer output during that task.
6. Change the template to a date before the latest custom repo snapshot. Delete the custom repo snapshot used by the template, this should enqueue a `delete-snapshots` task and a message should be sent in the consumer output.
7. Delete the custom repository used by the template. This should enqueue a `delete-repository-snapshots` task and a message should be sent during this task too.

### DIFF
--- a/pkg/dao/rpms.go
+++ b/pkg/dao/rpms.go
@@ -3,6 +3,7 @@ package dao
 import (
 	"context"
 	"fmt"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -21,6 +22,8 @@ import (
 )
 
 var DbInClauseLimit = 60000
+
+const TemplateErrataIDsPageLimit = 10000
 
 type rpmDaoImpl struct {
 	db            *gorm.DB
@@ -1034,4 +1037,35 @@ func (r *rpmDaoImpl) FetchForRepository(ctx context.Context, orgID string, repos
 	}
 
 	return rpms, nil
+}
+
+// FetchTemplateErrataIDs returns deduplicated and sorted errata IDs for the given template
+func (r *rpmDaoImpl) FetchTemplateErrataIDs(ctx context.Context, orgID string, templateUUID string) ([]string, error) {
+	limit := TemplateErrataIDsPageLimit
+	var offset int
+	seen := make(map[string]struct{})
+	var errataIDs []string
+
+	// paginate through ListTemplateErrata until all errata are fetched
+	for {
+		page, total, err := r.ListTemplateErrata(ctx, orgID, templateUUID, tangy.ErrataListFilters{}, api.PaginationData{
+			Limit: limit, Offset: offset,
+		})
+		if err != nil {
+			return nil, err
+		}
+		for _, errata := range page {
+			if _, ok := seen[errata.ErrataId]; !ok {
+				seen[errata.ErrataId] = struct{}{}
+				errataIDs = append(errataIDs, errata.ErrataId)
+			}
+		}
+		offset += len(page)
+		if offset >= total || len(page) == 0 {
+			break
+		}
+	}
+
+	slices.Sort(errataIDs)
+	return errataIDs, nil
 }
--- a/pkg/dao/templates.go
+++ b/pkg/dao/templates.go
@@ -124,7 +124,7 @@ func (t templateDaoImpl) create(ctx context.Context, tx *gorm.DB, reqTemplate ap
 	templatesModelToApi(modelTemplate, &respTemplate)
 	respTemplate.RepositoryUUIDS = reqTemplate.RepositoryUUIDS
 
-	event.SendTemplateEvent(*reqTemplate.OrgID, event.TemplateCreated, []event.TemplateEvent{event.MapTemplateResponse(respTemplate, nil, nil)})
+	event.SendTemplateEvent(*reqTemplate.OrgID, event.TemplateCreated, []event.TemplateEvent{event.MapTemplateResponse(respTemplate)})
 
 	return respTemplate, nil
 }
@@ -272,7 +272,7 @@ func (t templateDaoImpl) Update(ctx context.Context, orgID string, uuid string, 
 		return resp, err
 	}
 
-	event.SendTemplateEvent(orgID, event.TemplateUpdated, []event.TemplateEvent{event.MapTemplateResponse(resp, nil, nil)})
+	event.SendTemplateEvent(orgID, event.TemplateUpdated, []event.TemplateEvent{event.MapTemplateResponse(resp)})
 
 	return resp, err
 }
@@ -460,7 +460,7 @@ func (t templateDaoImpl) SoftDelete(ctx context.Context, orgID string, uuid stri
 
 	var resp api.TemplateResponse
 	templatesModelToApi(modelTemplate, &resp)
-	event.SendTemplateEvent(orgID, event.TemplateDeleted, []event.TemplateEvent{event.MapTemplateResponse(resp, nil, nil)})
+	event.SendTemplateEvent(orgID, event.TemplateDeleted, []event.TemplateEvent{event.MapTemplateResponse(resp)})
 
 	return nil
 }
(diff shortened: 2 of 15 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-10723

pkg/dao/rpms.go
```diff
@@ -1035,3 +1036,35 @@ func (r *rpmDaoImpl) FetchForRepository(ctx context.Context, orgID string, repos
...
+	for {
+		page, total, err := r.ListSnapshotErrata(ctx, orgID, snapshotUUIDs, tangy.ErrataListFilters{}, api.PaginationData{
+			Limit: limit, Offset: offset,
+		})
+		if err != nil {
+			return nil, err
+		}
+		for _, errata := range page {
+			seen[errata.ErrataId] = struct{}{}
+		}
```
TenSt, line 1062: I think it would be better to append to the final slice here instead of ranging over the map later.
Something like:
```
for _, errata := range page {
	if !seen[errata.ErrataId] {
        errataIDs = append(errataIDs, errata.ErrataId)
        seen[errata.ErrataId] = struct{}{}
    }
}
```

pkg/tasks/template_event_helper.go
```diff
@@ -0,0 +1,85 @@
...
+// Events with no updated advisories include only the template data.
+func sendTemplateEvent(orgID string, template api.TemplateResponse, isCreate bool, addedAdvisories []string, removedAdvisories []string) {
+	var eventName event.EventName
+	if isCreate {
+		eventName = event.TemplateCreated
+	} else {
+		eventName = event.TemplateUpdated
+	}
+
+	addedIndex, removedIndex, eventCount := 0, 0, 0
```
TenSt: It is better to do like this as you defining it at type zero value anyway:
```
var addedIndex, removedIndex, eventCount int
```

TenSt (dismissed):
lgtm! 
There are a couple comments, but feel free to merge as is or I'll re-approve if you decide to do the changes.

pkg/dao/templates.go
```diff
@@ -124,8 +124,6 @@ func (t templateDaoImpl) create(ctx context.Context, tx *gorm.DB, reqTemplate ap
 	templatesModelToApi(modelTemplate, &respTemplate)
 	respTemplate.RepositoryUUIDS = reqTemplate.RepositoryUUIDS
 
-	event.SendTemplateEvent(*reqTemplate.OrgID, event.TemplateCreated, []event.TemplateEvent{event.MapTemplateResponse(respTemplate, nil, nil)})
-
```
rverdile, line 128: I'm concerned about removing these events from here because they keep the patch database in sync with our database. If we move them only to the task (which can take some time to start, or fail), I think we might start seeing situations where the databases get out of sync.

What if we left these here, and changed the task to just send TemplateUpdate events for the advisories? 

What do you guys think? @xbhouse @TenSt

pkg/tasks/payloads/update_template_content.go
```diff
@@ -1,7 +1,9 @@
 package payloads
 
 type UpdateTemplateContentPayload struct {
-	TemplateUUID    string
-	RepoConfigUUIDs []string
-	PoolID          *string // Add during task runtime
+	TemplateUUID          string
+	RepoConfigUUIDs       []string
+	PreviousSnapshotUUIDs []string
```
rverdile: Should previousSnapshotUUIDs be found at runtime within the task? If the tasks fail or run in a different order than expected, I think this might cause some incorrect errata diffs. Thoughts?

pkg/dao/rpms.go
xbhouse: yep that's cleaner :) thanks, fixing that

pkg/dao/templates.go
xbhouse: hmm good point :) keeping the create here makes sense so template data gets sent asap instead of waiting on the task. and i can change the event name for events emitted from the task for new templates to `TemplateUpdated`

i removed the update event from the dao too, would it make sense to restore that as well for consistency?

pkg/tasks/payloads/update_template_content.go
xbhouse: i see what you mean. if 2 update tasks are running for the same template, and one finishes first or fails halfway through, the other would have the wrong prev snapshot uuids. 

in cases of out of order tasks or retries, wouldn't checking the prev snapshots inside the task also result in wrong errata diffs? i'll try to test these scenarios with prev snapshots in the payload vs the task to understand more clearly

pkg/dao/templates.go
rverdile: yeah, let's restore both

pkg/tasks/payloads/update_template_content.go
rverdile: Hmm so the task takes the repositories that should be in the template in the payload, and when the task is complete the template is in that state. So I think the diff would be okay? But maybe there's still a race condition in there somewhere

Not sure what other options would be except to do this in the handler, and not use a task (maybe too slow?). Or send the whole list to patch and let patch calculate the changes against the system (maybe too many events?).

pkg/dao/templates.go
TenSt: I'm okay with both approaches. We're usually going for the "eventual consistency", so some delay is okay. Depends on the delay. But if we can keep updating right away without loosing any performance or spending too much resources then let's do it.

pkg/tasks/payloads/update_template_content.go
TenSt: Can we run the tasks sequentially for the same template? This will create some delay, but will solve these issues. As we're having here async tasks, some delay is inevitable and we can use it to make sure consistency of the data.

xbhouse: the previous snapshots are sent in the payload because these are updated in the db via `insertTemplateRepoConfigsAndSnapshots` and lost before the task starts. but you're right that this could produce wrong diffs when tasks finish out of order :/ i don't see a way to retrieve or recompute this previous state in the task without storing it somewhere else first, i might have overlooked something though.

doing this in the handler would be slower and maybe too optimistic, meaning we send a diff to patch even if a step afterward fails. sending everything to patch would mean more events, and maybe too many, but it might make sense to have the diff logic where the template-advisory relationships are stored.

if storing the previous snapshots for a template somewhere makes sense or if there's another way to fetch the previous snapshot in the task, i can switch to that. or if we can ensure tasks are run sequentially (maybe this can be done with task dependencies?), that's another option :) otherwise, sending the full set to patch seems like a reasonable approach, but that would be a big design shift so we should talk through it with the team.

xbhouse: discussed a different approach in slack, we're going to add an endpoint patch can use to fetch the current advisories in a template. in patch, whenever there's an event that a template has been updated, we'll query this endpoint and calculate the diff. moving this back to draft while i update this :) 

thank you both for your input @rverdile @TenSt!

TenSt (dismissed):
lgtm! great work 👍

rverdile (commented):
delete-snapshots and delete-repository-snapshots also modify templates, so those need to send an update event too

xbhouse:
> delete-snapshots and delete-repository-snapshots also modify templates, so those need to send an update event too

thank you! was about to ask which other 2 tasks needed to send an event :) will add that in

TenSt (dismissed):

TenSt (approved):

PR #1541, opened 2026-06-17
Test: Apply kafka backoff and retry to sarama

swadeley:
## Summary
 Apply kafka backoff and retry to sarama

Kafka retry, backoff, and required-acks from config were defined but never applied in GetSaramaConfig(). This applies them so template and notification producers retry transient broker failures.

SendTemplateEvent now returns an error on send failure so callers can detect and report failed events instead of silently continuing.

## Testing steps

tests pass

### DIFF
--- a/pkg/config/event.go
+++ b/pkg/config/event.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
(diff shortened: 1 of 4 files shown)


swadeley:
Hi, closing this until some Kafka producer experiences overloading.

PR #1543, opened 2026-06-18
HMS-10544: Add lookup for orphaned distributions

swadeley:
## Summary
Add lookup for orphaned distributions
To work with delete-snapshots task
## Testing steps

start new clean backend without any repos, i.e. `make compose-clean`

Automated
```
go test ./pkg/tasks/ -run TestDeleteRepositoryVersionCleansUntrackedBlockingDistributions -v
go test ./test/integration/ -run TestDeleteSnapshotCleansUntrackedBlockingDistribution -v (with local compose stack)
```

Manual (local)

Setup

make compose-up and make run
`Org header: HDR="$(./scripts/header.sh $ORG_ID snapUser)"`

Repo with snapshotting enabled using https://fixtures.pulpproject.org/rpm-unsigned/
Simulate untracked Pulp distribution

Enable a temporary failure before GetRpmRepositoryVersion in snapshot_helper.go 

I added this to `pkg/tasks/snapshot_helper.go`:
```
	 if os.Getenv("SIMULATE_GET_RPM_REPOSITORY_VERSION_TIMEOUT") == "1" {
	 	return fmt.Errorf("simulated timeout fetching repository version: %w", context.DeadlineExceeded)
	 }
```
and then reran the backend with `SIMULATE_GET_RPM_REPOSITORY_VERSION_TIMEOUT=1` 

Create the repo and let the first snapshot task fail after creating Pulp resources but before writing the snapshot row.
Disable the simulation (temporary failure code) and restart the backend.
Change the repo URL (e.g. to rpm-with-sha-512) and trigger a second snapshot so a valid snapshot exists in the DB.

Exercise delete-snapshots cleanup

Confirm two snapshots exist:
`curl -s ".../repositories/$REPO_CONFIG_UUID/snapshots/" -H "$HDR" | jq '.meta.count'`
On the older snapshot, set a bogus distribution_href so delete skips the real Pulp distribution:
```
UPDATE snapshots
SET distribution_href = 'bogus-untracked-distribution-href'
WHERE uuid = '<older_snapshot_uuid>';
Delete the older snapshot via API (no trailing slash after $REPO_CONFIG_UUID):
curl -s -X DELETE \
  ".../repositories/$REPO_CONFIG_UUID/snapshots/<older_snapshot_uuid>" \
  -H "$HDR"
```
Expected results

DELETE returns 204
delete-snapshots task status is completed with no error
Snapshot count drops from 2 to 1
Worker logs include: Deleting untracked distribution blocking repository version deletion
Pulp distribution at the snapshot’s distribution_path and its repository version are removed despite the bogus DB distribution_href

Look for a line like this in the terminal where backend is running:
`1:19PM INF Deleting untracked distribution blocking repository version deletion base_path=`

### DIFF
--- a/pkg/tasks/delete_snapshots.go
+++ b/pkg/tasks/delete_snapshots.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/content-services/content-sources-backend/pkg/api"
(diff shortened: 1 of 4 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-10544

pkg/clients/pulp_client/rpm_distributions.go
```diff
@@ -121,3 +121,19 @@ func (r *pulpDaoImpl) ListDistributions(ctx context.Context) (*[]zest.RpmRpmDist
 	res := resp.GetResults()
 	return &res, nil
 }
+
+func (r *pulpDaoImpl) ListDistributionsByBasePath(ctx context.Context, basePath string) (*[]zest.RpmRpmDistributionResponse, error) {
```
rverdile: This is the right approach, but there is an existing method, FindDistributionByPath, that we should be able to use here

pkg/tasks/delete_snapshots.go
```diff
@@ -273,77 +273,100 @@ func (ds *DeleteSnapshots) deleteOrUpdatePulpContent(snap models.Snapshot, repo
...
-					if dist.PulpHref != nil && pub != nil && *pub == snap.PublicationHref {
-						logger.Debug().
-							Str("href", *dist.PulpHref).
-							Str("base_path", dist.BasePath).
-							Str("publication", *pub).
-							Msg("Found distribution using deleted publication")
-					}
-				}
-			}
+		if cleanupErr := ds.deleteDistributionsBlockingPublication(snap, snap.PublicationHref); cleanupErr != nil {
```
rverdile: We'd only want to do this if the poll task fails. The `DeleteRpmRepositoryVersion` method only starts a task to delete the repository version

pkg/tasks/delete_snapshots.go
```diff
@@ -273,77 +273,100 @@ func (ds *DeleteSnapshots) deleteOrUpdatePulpContent(snap models.Snapshot, repo
...
-	if deleteVersionHref != nil {
+
+	_, err = ds.getPulpClient().PollTask(ds.ctx, *deleteVersionHref)
+	if err != nil {
+		if cleanupErr := ds.deleteDistributionsBlockingPublication(snap, snap.PublicationHref); cleanupErr != nil {
+			return fmt.Errorf(
+				"error polling repository version deletion task for snapshot %v: %w (also failed to clean up blocking distributions: %v)",
+				snap.UUID, err, cleanupErr,
+			)
+		}
```
rverdile: We should also check that the error message contains the error that we're expecting. We don't want to do this if the task fails for a different reason

pkg/tasks/delete_snapshots.go
```diff
@@ -273,77 +273,100 @@ func (ds *DeleteSnapshots) deleteOrUpdatePulpContent(snap models.Snapshot, repo
...
 		}
-		return fmt.Errorf("failed to delete repository version %v for snapshot %v: %w", snap.VersionHref, snap.UUID, err)
+		deleteVersionHref, err = ds.getPulpClient().DeleteRpmRepositoryVersion(ds.ctx, snap.VersionHref)
+		if err != nil {
+			return fmt.Errorf("failed to delete repository version %v for snapshot %v after cleanup: %w", snap.VersionHref, snap.UUID, err)
+		}
+	}
+
+	if deleteVersionHref == nil {
+		return nil
```
rverdile: This value is only nil if there was an error, so we probably shouldn't return nil here. What was it returning previously?

rverdile:
A few comments, looks good overall :)

pkg/tasks/delete_snapshots.go
swadeley: Hi, it returned nil (success) without polling if there was no task "href".

pkg/clients/pulp_client/rpm_distributions.go
swadeley: Ok, changes made, thank you.

pkg/tasks/delete_snapshots.go
swadeley: Ok, changes made (cleanup only on poll failure), thank you.

pkg/tasks/delete_snapshots.go
swadeley: Ok, changes made, thank you.

pkg/tasks/delete_snapshots.go
swadeley: Ok, check what the error was first, thank you.

rverdile (commented):
Can you add an integration test for this in test/integration/delete_test.go

It would have to do something like:

1. Create repo, sync, snapshot
2. Delete the distribution href for the snapshot directly from pulp
3. Create a new distribution for that repository version at the same path
4. Run delete snapshots

pkg/tasks/delete_snapshots.go
```diff
@@ -274,77 +275,95 @@ func (ds *DeleteSnapshots) deleteOrUpdatePulpContent(snap models.Snapshot, repo
...
-			return fmt.Errorf("error polling repository version deletion task for snapshot %v: %w", snap.UUID, err)
+			if !isRepositoryVersionBlockedByDistribution(err) {
+				return fmt.Errorf("error polling repository version deletion task for snapshot %v: %w", snap.UUID, err)
+			}
+			if cleanupErr := ds.deleteDistributionsBlockingPublication(snap, snap.PublicationHref); cleanupErr != nil {
+				return fmt.Errorf(
+					"error polling repository version deletion task for snapshot %v: %w (also failed to clean up blocking distributions: %v)",
+					snap.UUID, err, cleanupErr,
+				)
+			}
```
rverdile, line 462: For all the error messages in this function, can we remove the snapshotUUID and versionHref from the message itself? Instead, let's add them as fields to the logger, `snapshot_uuid` and `version_href`. We do something simlar with RequestID and TaskID as an example.

rverdile: so like logger.Str("snapshot_uuid", snap.UUID)

pkg/tasks/delete_snapshots.go
```diff
@@ -274,77 +275,95 @@ func (ds *DeleteSnapshots) deleteOrUpdatePulpContent(snap models.Snapshot, repo
...
+func (ds *DeleteSnapshots) deleteDistributionsBlockingPublication(snap models.Snapshot, publicationHref string) error {
+	logger := LogForTask(ds.task.Id.String(), ds.task.Typename, ds.task.RequestID)
+
+	dist, err := ds.getPulpClient().FindDistributionByPath(ds.ctx, snap.DistributionPath)
+	if err != nil {
+		return fmt.Errorf("failed to find distribution at base path %v: %w", snap.DistributionPath, err)
+	}
+	if dist == nil || dist.PulpHref == nil {
+		return nil
+	}
```
rverdile, line 343: let's add a warning log here because if we're in this function we likely expect to find a distribution, so we'd want to know if we don't

rverdile (commented):
There are some repositories in stage and prod with hundreds of old snapshots. I am concerned about how long it would take an individual task to run in this case. Can you add a limit to the number of snapshots added to the delete task here? https://github.com/content-services/content-sources-backend/blob/main/pkg/external_repos/commands/cleanup.go#L152-L159

Let's say 100. And we can monitor from there if we need to increase it.

swadeley:
> There are some repositories in stage and prod with hundreds of old snapshots. I am concerned about how long it would take an individual task to run in this case. Can you add a limit to the number of snapshots added to the delete task here? https://github.com/content-services/content-sources-backend/blob/main/pkg/external_repos/commands/cleanup.go#L152-L159
> 
> Let's say 100. And we can monitor from there if we need to increase it.

Hi, I am adding:
`maxSnapshotsPerDeleteTask = 10`
to `pkg/external_repos/commands/cleanup.go`

pkg/tasks/delete_snapshots.go
swadeley: OK, changes made, thank you.

pkg/tasks/delete_snapshots.go
swadeley: OK, changes made, thank you.

pkg/tasks/delete_snapshots.go
```diff
@@ -274,78 +275,132 @@ func (ds *DeleteSnapshots) deleteOrUpdatePulpContent(snap models.Snapshot, repo
...
+		logger.Error().Err(err).Msg("failed to delete repository version")
+		return fmt.Errorf("failed to delete repository version: %w", err)
 	}
+
 	if deleteVersionHref != nil {
 		_, err = ds.getPulpClient().PollTask(ds.ctx, *deleteVersionHref)
 		if err != nil {
-			return fmt.Errorf("error polling repository version deletion task for snapshot %v: %w", snap.UUID, err)
+			if !isRepositoryVersionBlockedByDistribution(err) {
+				logger.Error().Err(err).Msg("error polling repository version deletion task")
```
rverdile: let's remove this error log `logger.Error().Err(err).Msg("error polling repository version deletion task")`. this returns an error so we'll see it in the return

pkg/tasks/delete_snapshots.go
rverdile: I think you missed adding a log here
```
if dist == nil || dist.PulpHref == nil {
		return nil
	}
```
We want to make sure this function doesn't fail silently. In this case we're expecting a distribution to exist since the repository version deletion is failing. If we don't find a distribution, we could be looking it up wrong.

pkg/tasks/delete_snapshots.go
```diff
@@ -44,7 +45,7 @@ func DeleteSnapshotsHandler(ctx context.Context, task *models.TaskInfo, _ *queue
 	ctxWithLogger := logger.WithContext(ctx)
 
 	daoReg := dao.GetDaoRegistry(db.DB)
-	domainName, err := daoReg.Domain.Fetch(ctxWithLogger, task.OrgId)
+	domainName, err := daoReg.Domain.FetchOrCreateDomain(ctxWithLogger, task.OrgId)
```
rverdile: we should keep this as Fetch. We don't expect to create here. It should already exist since it's deleting existing resources.

pkg/tasks/delete_snapshots.go
swadeley: Hi @rverdile , I fixed that and I found some other cases where an error will be returned thereby making the prior log message redundant. I changed that in a separate commit for transparency, your ease of review.

pkg/tasks/delete_snapshots.go
swadeley: OK, thanks

pkg/tasks/delete_snapshots.go
swadeley: oh, sorry. Now fixed on line 335

rverdile:
Looks like the integration test is failing because an untracked snapshot isn't being cleaned up

```
2026-07-15T12:45:31.6880604Z --- FAIL: TestDeleteTest (134.14s)
2026-07-15T12:45:31.6880981Z     --- FAIL: TestDeleteTest/TestDeleteSnapshotCleansUntrackedBlockingDistribution (27.95s)
2026-07-15T12:45:31.6881150Z         suite_test.go:275: 
2026-07-15T12:45:31.6881880Z             	Error Trace:	/home/runner/work/content-sources-backend/content-sources-backend/test/integration/suite_test.go:275
2026-07-15T12:45:31.6882934Z             	            				/home/runner/work/content-sources-backend/content-sources-backend/test/integration/delete_test.go:521
2026-07-15T12:45:31.6885191Z             	Error:      	Should be empty, but was failed to delete pulp content for snapshot 50984240-db87-4e86-a7a3-1e8812eda09b: failed to delete repository version: error deleting rpm repository versions: 400 Bad Request: ["The repository version cannot be deleted because it (or its publications) are currently being used to distribute content. Please update the necessary distributions first."]
2026-07-15T12:45:31.6885800Z             	Test:       	TestDeleteTest/TestDeleteSnapshotCleansUntrackedBlockingDistribution
2026-07-15T12:45:31.6885960Z         suite_test.go:277: 
2026-07-15T12:45:31.6886679Z             	Error Trace:	/home/runner/work/content-sources-backend/content-sources-backend/test/integration/suite_test.go:277
2026-07-15T12:45:31.6887726Z             	            				/home/runner/work/content-sources-backend/content-sources-backend/test/integration/delete_test.go:521
2026-07-15T12:45:31.6887912Z             	Error:      	Not equal: 

```

swadeley:
> Looks like the integration test is failing because an untracked snapshot isn't being cleaned up
> 
OK, changes made, thank you

rverdile (commented):
Thanks Stephen, all the changes look good. I did however realize that template distributions aren't covered. When we lookup the distribution by `snapshot.dist_path`, we're only getting the distribution created for the snapshot. But we could be missing some orphaned distributions that were created for templates.

So for templates we need to do something similar, but have after computing the distribution path for the template. For example:

1. Query the DAO for templates that contain snap.RepositoryConfigurationUUID                                                                       
 2. For each template, compute the template distribution path via getDistPathAndName 
 3. Call FindDistributionByPath for each to check if the distribution's publication matches snap.PublicationHref

swadeley:
Hi @rverdile 

deleteDistributionsBlockingPublication now searches for Templates with relevant repo UUID and then checks those Templates and deletes any distribution whose publication matches snap.PublicationHref

Thank you

rverdile (approved):
looks good! let's monitor the delete-snapshots tasks in stage and see if they start passing

PR #1544, opened 2026-06-18
HMS-10835: Hardcode pulp content path

rverdile:
## Summary
The pulp content path should never change. This makes is configurable, but hardcodes it with the expected default.

## Testing steps
1. Integration tests are probably enough
2. You could also create a snapshot and then make sure the content URL is correct

### DIFF
--- a/pkg/clients/pulp_client/status.go
+++ b/pkg/clients/pulp_client/status.go
@@ -28,19 +28,12 @@ func (r *pulpDaoImpl) Status(ctx context.Context) (*zest.StatusResponse, error) 
 	return status, nil
 }
 
-func (r *pulpDaoImpl) GetContentPath(ctx context.Context) (string, error) {
-	resp, err := r.Status(ctx)
+func (r *pulpDaoImpl) GetContentPath() (string, error) {
+	c := config.Get()
+	pulpContentPath, err := url.JoinPath(c.Clients.Pulp.ContentOrigin, c.Clients.Pulp.ContentPathPrefix)
 	if err != nil {
 		return "", err
 	}
-
-	contentPathPrefix := resp.ContentSettings.ContentPathPrefix
-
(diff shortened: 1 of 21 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-10835

pkg/config/config.go
```diff
@@ -105,7 +105,8 @@ type Pulp struct {
 	ClientCertPath    string       `mapstructure:"client_cert_path"`
 	ClientKeyPath     string       `mapstructure:"client_key_path"`
 	CACertPath        string       `mapstructure:"ca_cert_path"`
-	ContentOrigin     string       `mapstructure:"content_origin"` // hostname of the location of pulp content
+	ContentOrigin     string       `mapstructure:"content_origin"`      // hostname of the location of pulp content
+	ContentPathPrefix string       `mapstructure:"content_path_prefix"` // path prefix for pulp content (e.g. /api/pulp-content/)
```
swadeley:
```suggestion
	ContentPathPrefix string       `mapstructure:"content_path_prefix"` // path prefix for pulp content (e.g. /pulp/content/)
```

pkg/clients/pulp_client/status.go
```diff
@@ -29,18 +29,11 @@ func (r *pulpDaoImpl) Status(ctx context.Context) (*zest.StatusResponse, error)
 }
 
 func (r *pulpDaoImpl) GetContentPath(ctx context.Context) (string, error) {
```
swadeley: GetContentPath(ctx) no longer uses ctx. Lint did not flag it; you could rename to _ context.Context if you want to be explicit.

rverdile:
thanks @swadeley, updated!

swadeley (approved):
ACK

PR #1549, opened 2026-06-24
HMS-10835: fix pulp content path

rverdile:
## Summary
The pulp content path is different in stage/prod than local. Sets the content path to the stage/prod one and also updates the local one to match

### DIFF
--- a/deployments/build/env-variables.yaml
+++ b/deployments/build/env-variables.yaml
@@ -11,6 +11,8 @@ env:
     value: ${{CLIENTS_PULP_SERVER}}
   - name: CLIENTS_PULP_CONTENT_ORIGIN
     value: ${{CLIENTS_PULP_CONTENT_ORIGIN}}
+  - name: CLIENTS_PULP_CONTENT_PATH_PREFIX
+    value: ${{CLIENTS_PULP_CONTENT_PATH_PREFIX}}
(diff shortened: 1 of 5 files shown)


dominikvagner (dismissed):
ack ✅

xbhouse:
https://issues.redhat.com/browse/HMS-10835

deployments/build/env-variables.yaml
```diff
@@ -11,6 +11,9 @@ env:
     value: ${{CLIENTS_PULP_SERVER}}
   - name: CLIENTS_PULP_CONTENT_ORIGIN
     value: ${{CLIENTS_PULP_CONTENT_ORIGIN}}
+  - name: CLIENTS_PULP_CONTENT_PATH_PREFIX
+    value: ${{CLIENTS_PULP_CONTENT_PATH_PREFIX}}
+    default: "/api/pulp-content/"
```
xbhouse: i think `default` might not be a recognized parameter? maybe we need to set `/api/pulp-content/` as a value in the [parameters](https://github.com/rverdile/content-sources-backend/blob/db00c695a3d24b87ae07395af6d46c2b52787221/deployments/build/env-variables.yaml#L225) section of env-variables.yaml, and then re-run `make deployment-generate`

rverdile: good catch, thanks!

xbhouse (approved):

PR #1559, opened 2026-07-01
HMS-10936,HMS-10938: add import for lightwell repositories

rverdile:
## Summary
Adds importing of the lightwell repos

I intentionally left out the lightwell repos from list/fetch methods because I wanted to think about the right way to handle that. Similarly, I left out the feature name from the import. This intention here is really just to focus on the importing and the database changes.

I also might need to add some tests

## Testing steps
1. Point your local pulp server to stage pulp
2. Run dbmigrate up
3. Run external-repos import
4. Two repos should be created and re-running import should not make any new additions


db/migrations/20260701093355_add_lightwell_columns.down.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+
+ALTER TABLE repositories
+    DROP CONSTRAINT IF EXISTS repositories_unique_published_dist_url_origin,
+    DROP COLUMN IF EXISTS published_distribution_url,
+    DROP COLUMN IF EXISTS security_level,
+    DROP COLUMN IF EXISTS ecosystem;
```
jlsherrill: we might could derive ecosystem from the content type in the frontend? since content_type is already in the api response.

xbhouse:
https://issues.redhat.com/browse/HMS-10936

db/migrations/20260701093355_add_lightwell_columns.down.sql
rverdile: good point!

pkg/dao/uploads_test.go
```diff
@@ -88,6 +88,8 @@ func (s *UploadsSuite) TestListUploadsForCleanup() {
 	uploadDao := s.uploadsDao()
 	ctx := context.Background()
 
+	s.tx.Exec("DELETE FROM uploads")
```
jlsherrill, line 91: was this fixing something?

pkg/models/repository_configuration.go
```diff
@@ -140,6 +140,10 @@ func (rc *RepositoryConfiguration) IsCommunity() bool {
 	return rc.OrgID == config.CommunityOrg
 }
 
+func (rc *RepositoryConfiguration) IsLightwell() bool {
+	return rc.OrgID == config.LightwellOrg
+}
+
```
jlsherrill: i don't actually see this used.

jlsherrill:
we need to adjust the response checker to handle lightwell orgs, listing repos gets:

```
{
  "errors": [
    {
      "status": 500,
      "title": "Organization ID mismatch",
      "detail": "Response organization ID does not match user organization ID"
    }
  ]
}
```

here: ./pkg/middleware/enforce_consistent_org_id.go

pkg/external_repos/lightwell_repos.json
```diff
@@ -0,0 +1,4 @@
+[
+  {"name": "lightwell/java/remediated", "type": "maven", "base_path": "java/remediated"},
+  {"name": "lightwell/java/validated", "type": "maven", "base_path": "java/validated"}
```
jlsherrill: ohh, should the feature be imported into repo_configs here?

jlsherrill: or we could hardcode it in the lightwell importer

jlsherrill: and the feature name is `lightwell-network`

pkg/dao/uploads_test.go
rverdile: yeah this test can fail locally if there are uploads

pkg/models/repository_configuration.go
rverdile: I just added it because the other orgs are there, ill remove it for now. we can add it back if we need it

pkg/dao/repository_configs.go
```diff
@@ -1709,6 +1712,52 @@ func (r repositoryConfigDaoImpl) InternalOnly_RefreshPredefinedSnapshotRepo(ctx
...
+		return nil, result.Error
+	}
+
+	newRepo = models.Repository{}
+	result = r.db.WithContext(ctx).Where("published_distribution_url = ? AND origin = ?", models.CleanupURL(publishedDistURL), config.OriginLightwell).First(&newRepo)
+	if result.Error != nil {
+		return nil, result.Error
+	}
+
+	newRepoConfig := models.RepositoryConfiguration{
```
xbhouse, line 1739: i think we'd need to set the feature_name here too

rverdile: thanks!

xbhouse:
https://issues.redhat.com/browse/HMS-10938

jlsherrill (approved):

PR #1560, opened 2026-07-03
HMS-10956: Add lightwell package api

jlsherrill:
## Summary

This adds an api to support the tang changes here: https://github.com/content-services/tang/pull/25

The idea being that these two apis are agnostic to what content type it is. (although for something like python, the group field may be blank)


jlsherrill:
need to release tang so this will run and pass

pkg/handler/packages.go
```diff
@@ -0,0 +1,124 @@
...
+// @Failure      404 {object} ce.ErrorResponse
+// @Failure      500 {object} ce.ErrorResponse
+// @Router       /repositories/{uuid}/packages [get]
+func (ph *PackageHandler) listPackages(c echo.Context) error {
+	uuid := c.Param("uuid")
+	// _, orgID := getAccountIdOrgId(c)
+	pageData := ParsePagination(c)
+
+	// Fetch repository config to get content type and distribution URL
+	repo, err := ph.DaoRegistry.RepositoryConfig.Fetch(c.Request().Context(), config.LightwellOrg, uuid) // TODO, don't hardcode lightwell org
```
jlsherrill, line 52: This is hard coded to lightwell, because i realized that Fetch() doesn't honor entitlements and hasn't been updated to work with lightwell, working on a followup pr

xbhouse:
https://issues.redhat.com/browse/HMS-10956

rverdile (approved):
looks good! merging

PR #1563, opened 2026-07-04
HMS-10958: Workaround maven lightwell repos without repository link

jlsherrill:
## Summary

This workarounds a maven repository issue with lightwell stage/prod issues, where by one maven repo has a distribution without a link to the reposiotry.  Instead it links to a publication.  pulp-maven provides no publication apis so it is a dead end, however for lightwell the repo name and distro name are the same, so we can 'abuse' that to lookup the right repository.  

## Testing steps

🙏🏻


xbhouse:
https://issues.redhat.com/browse/HMS-10958

TenSt (approved):

PR #1567, opened 2026-07-04
HMS-10956: add resolveRepositoryHref to maven details

rverdile:
## Summary

The details API was missing this so remediated packages were not returning build details


xbhouse:
https://issues.redhat.com/browse/HMS-10956

xbhouse (approved):

pkg/handler/packages.go
jlsherrill, line 242: we should probably rename this to  getMavenPackageDetail?   but we can do it in a followup.

jlsherrill (approved):

PR #1588, opened 2026-07-14
HMS-10468: API: Keep re-trying failed deletion tasks

swadeley:
## Summary
Add retry-failed-deletion-tasks job
To keep re-trying failed deletion tasks

### DIFF
--- a/pkg/config/tasks.go
+++ b/pkg/config/tasks.go
@@ -36,6 +36,8 @@ var TaskTypes = []string{
 
 var RequeueableTasks = []string{DeleteTemplatesTask, DeleteRepositorySnapshotsTask, UpdateTemplateContentTask, DeleteSnapshotsTask}
 
+var DeletionTasks = []string{DeleteRepositorySnapshotsTask, DeleteTemplatesTask}
+
 var CancellableTasks = []string{IntrospectTask, RepositorySnapshotTask, UpdateTemplateContentTask}
 
 const ObjectTypeRepository = "repository"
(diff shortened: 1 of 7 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-10468

swadeley:
Hi,

Note that:

1.  “once per day” can mean up to ~48h to first daily retry
A task that exhausts retries at 07:00 UTC is not yet 24h old at the next 06:00 run, so it waits until the following day. That’s reasonable, but worth knowing if you expected “next morning after permanent failure.”

2. overlap with existing retry-failed-tasks
The older job retries recent delete-repository-snapshots failures (last 72h, no retry-count or limit checks) and currently has no schedule in the template (requires manual operation). This new job covers all deletion types, only when retries >= MaxTaskRetries, older than 24h, limited to 100.

3. DeletionTasks duplicates TasksToCleanupIfCompleted

4. no upper bound
Failed deletion tasks are never cleaned up (TasksToCleanupIfCompleted only drops completed ones). This job will keep trying the oldest 100/day forever, which matches the orphan-cleanup goal. If some failures are permanent (bad payload or missing pulp objects), we might want a max-age or max-lifetime-retry cap value?

pkg/config/tasks.go
```diff
@@ -36,6 +36,8 @@ var TaskTypes = []string{
 
 var RequeueableTasks = []string{DeleteTemplatesTask, DeleteRepositorySnapshotsTask, UpdateTemplateContentTask, DeleteSnapshotsTask}
 
+var DeletionTasks = []string{DeleteRepositorySnapshotsTask, DeleteSnapshotsTask, DeleteTemplatesTask}
```
rverdile: I think we can leave DeleteSnapshotsTask out for now. It is already sorta being re-run by the job that queues them (it make a new task, but the result is the same). And I don't wanna start retrying while we know that they will likely fail because of the other bug.

swadeley: Hi @rverdile, OK, I have removed that.

Thank you.

rverdile (approved):
nice job!

PR #1589, opened 2026-07-16
HMS-10705: add the ability to (un)publish snapshots

dominikvagner:
## Summary
This adds a new endpoint that enables (un)publishing of snapshots from partner repositories. 
Publishing a snapshot removes its Pulp content guard, making it accessible to other organizations. 
Unpublishing restores the org-scoped guard. 
The repository's latest distribution guard is kept in sync based on whether any published snapshots remain.

## Testing steps
1. Create a repo with at least two snapshots and set it to partner. 
    `UPDATE repository_configurations SET partner = true WHERE uuid = '...';`
2. Hit the new endpoint and publish a snapshot(s)
    `PATCH :8000/api/content-sources/v1.0/repositories/<repo_config_uuid>/snapshots/<snap_uuid>/published  published:=true`
3. Verify the behavior matches the requirements in the ticket (tasks, content guards, status syncing, /latest pointing to the latest published snapshot), by using our APIs, manually checking the database, and in the Pulp API.
	ex.:
    - `SELECT r.public FROM repositories r JOIN repository_configurations rc ON rc.repository_uuid = r.uuid WHERE rc.uuid = '<repo_config_uuid>';`
    - `GET http://pulp.content:8080/api/pulp/<your_domain>/api/v3/distributions/`

_Associated Ticket:_ [[HMS-10705](https://redhat.atlassian.net/browse/HMS-10705)]

[HMS-10705]: https://redhat.atlassian.net/browse/HMS-10705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


xbhouse:
https://issues.redhat.com/browse/HMS-10705

pkg/handler/utils_test.go
```diff
@@ -247,3 +259,96 @@ func (suite *UtilsSuite) TestFetchSnapshotUUIDsForRepos() {
...
+	assert.Equal(suite.T(), uuid.Nil, resultID)
+}
+
+func (suite *UtilsSuite) TestEnqueueUpdateSnapshotPublishedTask() {
+	c := mockEchoContext(suite.T())
+	mockTC := client.NewMockTaskClient(suite.T())
+	repoUUID := uuid.NewString()
+	snapshotUUID := uuid.NewString()
+	taskID := uuid.New()
+	depID := uuid.New()
```
rverdile: Imo the tests in this file are kinda redundant because this will be tested by the handler tests. However, there may be an opportunity to define some helper functions here for mock enqueuing, which could simplify the handler test code. What do you think?

pkg/handler/snapshots.go
```diff
@@ -228,6 +229,75 @@ func (sh *SnapshotHandler) getRepoConfigurationFile(c echo.Context) error {
...
+		return ce.NewErrorResponse(ce.HttpCodeForDaoError(err), "Error fetching repository config", err.Error())
+	}
+	if repo.OrgID != orgID {
+		return ce.NewErrorResponse(http.StatusForbidden, "Error (un)publishing snapshot", "Cannot (un)publish snapshots for repositories outside your organization.")
+	}
+	if !repo.Partner {
+		return ce.NewErrorResponse(http.StatusBadRequest, "Error (un)publishing snapshot", "Cannot (un)publish snapshots of non-partner repositories.")
+	}
+
+	snapshot, err := sh.DaoRegistry.Snapshot.UpdatePublishedStatus(c.Request().Context(), *params.Published, repoUUID, snapshotUUID)
```
rverdile: What do you think about moving these checks into the dao method for UpdatePublishedSnapshots? Might make sense to gate this at the dao level.

pkg/handler/utils_test.go
dominikvagner: good idea! fixed 👍🏼😄

pkg/handler/snapshots.go
dominikvagner: yeah, might make sense to check it there, so callers of the dao method don't forget to check those 💭 
moved 🚚 🫡

rverdile (dismissed):
nice job!

pkg/handler/snapshots.go
```diff
@@ -228,6 +229,66 @@ func (sh *SnapshotHandler) getRepoConfigurationFile(c echo.Context) error {
...
 }
 
+// PublishSnapshot godoc
+// @summary        Publish or unpublish a snapshot
+// @ID             publishSnapshot
+// @Description    This enables changing the published status of a specific snapshot.
+// @Tags           snapshots
+// @Param          repo_uuid     path string                        true "Repository UUID."
+// @Param          snapshot_uuid path string                        true "Snapshot UUID."
+// @Param          body          body api.ListSnapshotByDateRequest true "Published status."
```
rverdile: the body type here is should be SnapshotPublishedUpdateRequest

dominikvagner: oops, fixed 👍🏼 🫡

rverdile (approved):

PR #1586, opened 2026-07-13
HMS-10971: add npm packages support for lightwell

TenSt:
This PR:
- Bumps `tang` to `v0.0.29` and adds `ContentTypeNpm`
- Adds npm package list/detail APIs via tang (`NpmPackageList`, `NpmPackageGet`, etc.)
- Uses `{scope}/{name}` routes; `-` for unscoped (e.g. `@types`/`is-odd`)
- Reuses `PackageItem.group` for npm scope in list responses
- Adds handler unit tests and regenerates OpenAPI spec

PR #1594, opened 2026-07-21
HMS-10709: add Mark/Unmark partner repositories

Starle21:
## Summary

- new endpoint /admin/repositories/:uuid/partner to mark/unmark repositories as partner added
- new dao method SetPartnerRepo added
- new feature AdminPartnerRepositories added
- new go tests for the partner endpoint in admin_repositories_test added
- new go tests for the SetPartnerRepo in partner_visibility_test added

At this first mvp version, if there still exist any published snapshots, unmarking the repo is disabled in the dao level. Might want to review this later for the more final version.

## Testing steps
- add an upload repository and then manually test using the endpoint that it gets marked as partner
- the new go tests pass


xbhouse:
https://issues.redhat.com/browse/HMS-10708

rverdile:
I think there's a typo in the card number, should be HMS-10709 :)

pkg/handler/admin_repositories.go
```diff
@@ -0,0 +1,59 @@
...
+
+	adminRepositoriesHandler := AdminRepositoriesHandler{
+		DaoRegistry: *daoReg,
+	}
+
+	addRepoRoute(engine, http.MethodPatch, "/admin/repositories/:uuid/partner", adminRepositoriesHandler.toggleAsPartner, rbac.RbacVerbWrite, checkAdminPartnerReposAccessible)
+}
+
+func (arh *AdminRepositoriesHandler) toggleAsPartner(c echo.Context) error {
+	// extract repoconfig uuid and partner repo flag (true or false)
```
rverdile: Nitpick, but I think we can remove the comments in this method. The code you've written is readable :)

Starle21: Sure, thanks, forgot to delete them :)

xbhouse:
https://issues.redhat.com/browse/HMS-10709

pkg/dao/partner_visibility_test.go
dominikvagner, line 1: I think these test should probably be in the `pkg/dao/repository_configs_test.go` file and not here 💭 
they are testing `SetPartnerRepo` from `repository_configs.go` and not `partner_visibility.go`, it makes it easier to find 😅

pkg/handler/admin_repositories.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	adminRepositoriesHandler := AdminRepositoriesHandler{
+		DaoRegistry: *daoReg,
+	}
+
+	addRepoRoute(engine, http.MethodPatch, "/admin/repositories/:uuid/partner", adminRepositoriesHandler.toggleAsPartner, rbac.RbacVerbWrite, checkAdminPartnerReposAccessible)
+}
+
+func (arh *AdminRepositoriesHandler) toggleAsPartner(c echo.Context) error {
```
dominikvagner, line 41: right now empty body will be treated as `partner=false`, it might be nice to add a check here to verify it's not nil _(the Partner field in the request struct would have to be changed to a pointer)_ 💭

dominikvagner (commented):
two small comments, otherwise works as expected! ✨ good job! 👏🏼

pkg/dao/partner_visibility_test.go
Starle21: True, moving the tests there.

pkg/handler/admin_repositories.go
Starle21: Agreed, nil check should be there, thanks for catching this!

pkg/config/config.go
```diff
@@ -431,6 +432,10 @@ func setDefaults(v *viper.Viper) {
 	v.SetDefault("features.kessel.organizations", nil)
 	v.SetDefault("features.kessel.users", nil)
 	v.SetDefault("features.lightwell.enabled", false)
+	v.SetDefault("features.admin_partner_repositories.enabled", false)
```
xbhouse, line 437: i think we'd want to add these new vars to the deployment.yaml too

dominikvagner: true, thanks for catching that 🔍

Starle21: I have added the new envs, should I also set default values? I am not sure about this @xbhouse

dominikvagner: defaults should be false or nil, and you are doing that so all good imo 😄 
then the actual values can be set up in the deploy config, when it's actually deployed 😌

xbhouse: yep what @dominikvagner said :) thanks!

Starle21: I actually meant defaults in the clowder template - set up in the `parameters` section in `env-variables`. So that there is a `true for stage`, so that the feature functions after the PR is merged with main and then automatically redeployed to stage. Or do I have to always rather set it up in app-interface? :)

dominikvagner (commented):
the second commit can be dropped _(or at least renamed || put into a different PR)_, as it doesn't actually generate anything for the new endpoint, and it shouldn't as we aren't including `/admin/**/*` routes in the spec/client 😅 
the actual generated stuff comes from different changes for which it wasn't done 🫣

the rest looks good! ✨

pkg/config/config.go
dominikvagner: _imo:_ it's better/safer to not enable it by default, and explicitly control it via the deploy config 😅 
_(this will need the orgs defined there anyway and those shouldn't be in the default values here)_

xbhouse: i think the defaults should be set from the config defaults as long as the variables are in the parameters section of the template, _but_ the deployment failed so i'm monitoring the latest konflux pipeline ([this one](https://github.com/content-services/content-sources-backend/pull/1594/checks?check_run_id=89982492326)) run to see if the values are pulled correctly. it might be the case that we need to add those variables in app-interface first or set the defaults in the template too

xbhouse: ok the [variables aren't set](https://console-openshift-console.apps.crc-eph.r9lp.p1.openshiftapps.com/k8s/ns/ephemeral-rg3zvs/pods/content-sources-backend-service-7c694686fb-2n5pw/environment) when deployed, so let's go ahead and set the default for `FEATURES_ADMIN_PARTNER_REPOSITORIES_ENABLED` to `false` in the template parameters, and set it to `true` in [app-interface for stage](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/content-sources/deploy.yml?ref_type=heads#L145). usually if we add additional variables in the deployment template, we need to set them in app-interface at the same time :)

Starle21: I see, thank you @xbhouse! I'll do that and open an MR for that.

Starle21:
> the second commit can be dropped _(or at least renamed || put into a different PR)_, as it doesn't actually generate anything for the new endpoint, and it shouldn't as we aren't including `/admin/**/*` routes in the spec/client 😅 the actual generated stuff comes from different changes for which it wasn't done 🫣
> 
> the rest looks good! ✨

Aha, ok, I can create a new PR for that. 😅 Do you have any idea @dominikvagner which PR provided the changes the openapi should have been regenerated?

dominikvagner:
> Aha, ok, I can create a new PR for that. 😅 Do you have any idea @dominikvagner which PR provided the changes the openapi should have been regenerated?

that'd be great! thx 🙇🏼 
most likely [this one](https://github.com/content-services/content-sources-backend/pull/1586) which adds `npm` support, must have been run only with `make openapi-doc` without `make openapi-js`, it's easy to forget 😅

dominikvagner (approved):
nice work! 👏🏼 

ack ✅ 🚀

PR #1595, opened 2026-07-21
HMS-10597: If add-uploads task fails, retrying same doesn't work

swadeley:
## Summary

Add orphan repo version check to upload repo
    
    HMS-10544: Add lookup for orphaned distributions

## Testing steps

`go test ./pkg/tasks/ -run 'TestAddUploadsSuite' -v -count=1`

For local manual test:
make compose-up, make run

Create an upload repository (snapshot enabled).
Upload an RPM (UI, or docs/upload-rpm.py against your local API).
Simulate the orphan: delete the CS snapshot row for that repo, leave Pulp alone. Or kill the worker after ModifyRpmRepositoryContent and before snapshot create.
Re-upload the same RPM.
Expect:
Task succeeds
Log: using orphaned repo version ...
New snapshot exists and the RPM shows in the repo
Without the fix, step 4 succeeds but the repo still has no packages or snapshot

You might need to add:
 `Permission: "content-sources:*:upload",` and ` Permission: "content-sources:repositories:upload",`
to your `pkg/test/mocks/rbac/rbac.go` file.

### DIFF
--- a/pkg/tasks/add_uploads.go
+++ b/pkg/tasks/add_uploads.go
@@ -116,6 +116,7 @@ func (ur *AddUploads) Run() (err error) {
 				log.Debug().Msgf("added content to repository %v, but no new version created, likely already present", ur.payload.RepositoryConfigUUID)
 				return nil
 			}
+			log.Info().Msgf("using orphaned repo version %v for repository %v", *orphanedHref, ur.payload.RepositoryConfigUUID)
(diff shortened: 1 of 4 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-10597

TenSt (approved):
lgtm!

PR #1597, opened 2026-07-22
HMS-10706: partner repos snapshot updates

TenSt:
This PR:
- filter partner snapshot list/fetch/date so foreign viewers only see `published` snaps
- forbid deleting `published` snapshots
- skip Pulp `{repo}/latest` on partner create; re-point it to latest published (or remove)

### DIFF
--- a/pkg/handler/snapshots.go
+++ b/pkg/handler/snapshots.go
@@ -390,11 +390,14 @@ func (sh *SnapshotHandler) isDeleteAllowed(c echo.Context, orgID, repoUUID strin
 	hasErr := false
 	errs := make([]error, len(snapshotUUIDs))
 	for i := range snapshotUUIDs {
-		err = sh.isSnapInRepo(repoSnaps, snapshotUUIDs[i])
+		snap, err := findSnapInRepo(repoSnaps, snapshotUUIDs[i])
 		if err != nil {
 			hasErr = true
 			errs[i] = err
 			continue
+		}
+		if snap.Published {
+			return ce.NewErrorResponse(http.StatusBadRequest, "Error deleting snapshots", "Cannot delete a published snapshot")
 		}
 	}
 	if hasErr {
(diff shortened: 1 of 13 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-10706

pkg/handler/snapshots.go
dominikvagner, line 142: this is endpoint fails for foreign orgs with "snapshot not found", because it's pointing to the actual latest, not published 🐛 
so it would need something like the `fetchSnapshotForLatestDistribution` helper used elsewhere 💭 

that makes me wonder tho, if it wouldn't be better to have just one "FetchLatestSnapshot" that would check if the repo is partner in the DAO, so callers aren't required to fetch the repo config prior to calling it 😅
what do you think?

_(like here, you would have to fetch the repo first and then depending on that call the right fetch latest function, also would help with not having to repeat the helpers that call the right one and it alredy has the repoUUID)_

pkg/jobs/create_latest_distributions.go
```diff
@@ -54,22 +54,30 @@ func CreateLatestDistributions(_ []string) {
...
 			wg := sync.WaitGroup{}
 			for _, repo := range batch {
-				lastSnapshot := repo.LastSnapshot
-				if lastSnapshot == nil {
+				publicationHref := ""
+				if repo.Partner {
+					latestPublished, fetchErr := daoReg.Snapshot.FetchLatestPublishedSnapshotModel(ctx, repo.UUID)
+					if fetchErr != nil {
+						continue
+					}
```
dominikvagner: do we want to skip on all errors here? maybe just skipping on `ErrRecordNotFound` and surfacing the rest would be better 💭

pkg/handler/snapshots.go
TenSt: Great catch!

pkg/jobs/create_latest_distributions.go
TenSt: Good point, I've updated the code to skip only the `ErrRecordNotFound`.

TenSt:
@dominikvagner I've did the fixes you mentioned in the review + found and fixed small bug that could leak the unpublished snapshots for partner repos (there is a test that covers just this case with full explanation in the comment).

dominikvagner (approved):
looks good ✨ thanks 🙇🏼 

ack ✅ 🚀

PR #1602, opened 2026-07-24
Build: pin swag and openapi versions

Starle21:
## Summary
Swag and openapi were pulling latest releases
when `make openapi` command was run. Now both
of them are pinned to a particular version.

- swag-version is referenced from go.mod
- openapi-image-version is referenced from docker-compose.versions.yml, which was created so dependabot can provide updates for it

Both libraries are checked by dependabot for updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,3 +27,11 @@ updates:
       npm:
         patterns:
           - '*'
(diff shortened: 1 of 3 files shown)


TenSt (approved):
lgtm

PR #1616, opened 2026-08-03
Make create_lightwell_repo.sh idempotent

subpop:
## Summary

- Make `create_lightwell_repo.sh` find-or-create Pulp distributions by `base_path` so re-runs reuse existing repos instead of failing on duplicates
- Always re-run Maven populate and the CS import after ensure
- Fix usage comment, switch shebang to `#!/usr/bin/env bash`, and require bash 4+ (Darwin-specific install hint)

## Testing steps

- [ ] Run `./scripts/create_lightwell_repo.sh` against a fresh local Pulp; confirm domains/dists are created, maven populate runs, and CS import succeeds
- [ ] Re-run the same script; confirm existing distributions are reused (skipped create) and populate + import still run, exit 0
- [ ] On bash 3.x, confirm the script fails fast with a clear error (and Darwin install hint on macOS)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/scripts/create_lightwell_repo.sh
+++ b/scripts/create_lightwell_repo.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Creates Maven and Python repositories in Pulp and imports them into the
 # application as lightwell repositories.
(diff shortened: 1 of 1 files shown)


rverdile (approved):
thanks, looks good!

PR #1617, opened 2026-08-04
HMS-10707: templates changes to support partner repos

dominikvagner:
## Summary
This updates the repo validation for templates and the two template tasks along with surrounding helpers, to support consuming partner repositories.

- select published snapshots when foreign organizations use partner repositories
- update template and latest distributions after publication changes
- resolve partner content through the repository owner's Pulp domain
- synchronize template associations, snapshots, Candlepin content, and overrides

## Testing steps
This is a bit tedious to test, but it's doable. I did test it very thoroughly with local client VM. 
_(so you probably don't have to go to that extent)_
- create uploads repos with org X ➡️ mark them as partner ➡️ publish some snapshots.
- use them in templates (latest and fixed date), in both the owner and some viewer orgs.
- use those templates on a local VM: 
  - verify the content is accessible and registration works.
  - make changes to the templates + publish and unpublish snapshots ➡️ and watch that the templates are updated correctly.

### DIFF
--- a/pkg/dao/dao_mock.go
+++ b/pkg/dao/dao_mock.go
@@ -4243,6 +4243,32 @@ func (_c *MockSnapshotDao_FetchLatestSnapshot_Call) RunAndReturn(run func(ctx co
 	return _c
 }
 
+// HasPublishedSnapshot provides a mock function for the type MockSnapshotDao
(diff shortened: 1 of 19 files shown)


xbhouse:
https://issues.redhat.com/browse/HMS-10707

PR #1619, opened 2026-08-04
Build: POC - add Lightwell access tokens with Bearer auth

subpop:
Experimental proof-of-concept feature enabling org-admin token lifecycle and package-API Bearer auth so CI and clients can access Lightwell without interactive sessions, plus an internal validate path for Pulp.


jlsherrill:
thanks @subpop i think we have another way forward that we'll be investigating.

PR #1621, opened 2026-08-04
[RHCLOUD-49129] Fetch default ws bypassing permission check

astrozzc:
## Summary
https://redhat.atlassian.net/browse/RHCLOUD-49129
RBAC require explicit parameter to bypass the permission check when fetching default workspace

### DIFF
--- a/pkg/clients/kessel_client/client.go
+++ b/pkg/clients/kessel_client/client.go
@@ -60,7 +60,7 @@ func (k *kesselClientImpl) GetDefaultWorkspaceID(ctx context.Context, orgID stri
 	if err != nil {
 		return "", statusCode, fmt.Errorf("failed to get rbac url: %w", err)
 	}
-	url := fmt.Sprintf("%s/api/rbac/v2/workspaces/?type=default", rbacUrl)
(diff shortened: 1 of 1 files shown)


pkg/clients/kessel_client/client.go
```diff
@@ -60,7 +60,7 @@ func (k *kesselClientImpl) GetDefaultWorkspaceID(ctx context.Context, orgID stri
 	if err != nil {
 		return "", statusCode, fmt.Errorf("failed to get rbac url: %w", err)
 	}
-	url := fmt.Sprintf("%s/api/rbac/v2/workspaces/?type=default", rbacUrl)
+	url := fmt.Sprintf("%s/api/rbac/v2/workspaces/?type=default&with_ancestry=true", rbacUrl)
```
swadeley, line 64: Can we add brief code comment and the Jira ref before this line?
e.g.: // "with_ancestry=true"  required to fetch org's default workspace (RHCLOUD-49129)

astrozzc: Sure, added

